PR #41, opened 2023-10-03
Feature/turn based system

MathiasGruber:
# Pull Request
Converts battle system to be turn-based rather than round-based / live.

## License


MathiasGruber:
Works locally, and since we're alpha, pushing directly to main 🗡️

PR #136, opened 2024-09-25
Fixed bloodline manual text and updated who can see and handle reports.

Northern-Fayt:
# Pull Request

Allowing mods/admins to handle reports they make, but not handle reports made about them. 
Added ability for mod/admins to unstuck users in battle without ban/unbanning them
image

image

Added PVE Fights and a counter for how many they've done today to public profile
image

Also updated bloodline text in manual to direct players to wake island. 
image

Added heal and go option to BA. Discounts from village aren't applied. They're ordering ramen to go so it's more expensive.
image

Added training arena with configurable stats, gives 10 jutsu exp per completion. Exhausts itself at round 7. No limit daily.
image
image

Allow user to pick which A rank mission they would like to do
image

## License


app/src/server/api/routers/panel.ts
```diff
@@ -0,0 +1,63 @@
...
+export const panelRouter = createTRPCRouter({
+  forceAwake: protectedProcedure
+    .output(baseServerResponse)
+    .input(z.object({ userId: z.string() }))
+    .mutation(async ({ ctx, input }) => {
+      // Query
+      const [user, targetUser] = await Promise.all([
+        fetchUser(ctx.drizzle, ctx.userId),
+        fetchUser(ctx.drizzle, input.userId),
+        ctx.drizzle.insert(reportLog).values({
```
MathiasGruber: Think this should go to a Promise.all() section after the guard together with the update(userData)

app/src/server/api/routers/panel.ts
```diff
@@ -0,0 +1,63 @@
...
+import { eq } from "drizzle-orm";
+import { reportLog, userData } from "@/drizzle/schema";
+import { fetchUser } from "@/routers/profile";
+import { getServerPusher, updateUserOnMap } from "@/libs/pusher";
+import type { UserStatus } from "@/drizzle/constants";
+import { z } from "zod";
+import { inferRouterOutputs } from "@trpc/server";
+import { nanoid } from "nanoid";
+
+export const panelRouter = createTRPCRouter({
```
MathiasGruber: Maybe we should name this one staffRouter for clarify?

app/src/server/api/routers/panel.ts
```diff
@@ -0,0 +1,63 @@
...
+  forceAwake: protectedProcedure
+    .output(baseServerResponse)
+    .input(z.object({ userId: z.string() }))
+    .mutation(async ({ ctx, input }) => {
+      // Query
+      const [user, targetUser] = await Promise.all([
+        fetchUser(ctx.drizzle, ctx.userId),
+        fetchUser(ctx.drizzle, input.userId),
+        ctx.drizzle.insert(reportLog).values({
+          id: nanoid(),
```
MathiasGruber: I'm thinking instead of logging this into reportLog, maybe we should log it into the actionLog table. The reason being we could then add information about the user into the `changes` field - e.g. I'd love to know the userData prior to reset, the battle (or non-existence) of the battle data, etc. So that we can eventually figure out why this happens in the first place - ideally this should never happen, after all.

app/src/layout/Combat.tsx
```diff
@@ -624,13 +640,23 @@ const Combat: React.FC<CombatProps> = (props) => {
...
+                    id="return"
+                    className="basis-1/2 w-full"
+                    onClick={() => startArenaBattle({ aiId: arenaOpponentId })}
+                  >
+                    Go Again
+                  </Button>
+
+                  <Button
+                    id="heal-return"
+                    className="basis-1/2 w-full mt-1"
```
MathiasGruber, line 655: An option for this which would avoid adding new endpoints could be to put in the ramen options on this page, e.g. generalize the menu components from `app/ramenshop/page.tsx` into the `layout` folder so that we could insert them directly on this page. Let's discuss with Neon

```typescript
<div className="grid grid-cols-3 text-center font-bold italic p-3">
          <MenuEntry
            title="Small Bowl"
            entry="small"
            image="/ramen/small_bowl.webp"
            userData={userData}
            onPurchase={() => mutate({ ramen: "small" })}
          />
          <MenuEntry
            title="Medium Bowl"
            entry="medium"
            image="/ramen/medium_bowl.webp"
            userData={userData}
            onPurchase={() => mutate({ ramen: "medium" })}
          />
          <MenuEntry
            title="Large Bowl"
            entry="large"
            image="/ramen/large_bowl.webp"
            userData={userData}
            onPurchase={() => mutate({ ramen: "large" })}
          />
```

app/src/server/api/routers/profile.ts
```diff
@@ -746,6 +748,19 @@ export const profileRouter = createTRPCRouter({
         limit: 5,
       });
     }),
+  getUserDailyPveBattleCount: protectedProcedure
```
MathiasGruber, line 751: Since we are using this on all public profiles, and likely want it visible to everyone, I'm thinking we should add it directly the `getPublicUser` endpoint. I believe it can be added directly to the `with` statement on the `getPublicUser` query.

app/src/utils/time.ts
```diff
@@ -9,6 +9,22 @@ export const getGameTime = () => {
   return `${hours}:${minutes}:${seconds}`;
 };
 
+/**
+ * Get time since last reset which is in YYYY-MM-DDTHH:mm:ss.sssZ format
+ */
+export const getTimeOfLastReset = () => {
+  var date = new Date();
```
MathiasGruber, line 16: I hate dates in javascript, but yeah, 👍😂

MathiasGruber (commented):
Awesome, have a few suggestions for improvements again :)

app/src/server/api/routers/panel.ts
Northern-Fayt: Renamed

app/src/server/api/routers/panel.ts
Northern-Fayt: I changed it to actionLog and moved it after the guards and logged previous state and battleId. If you want to log more specifically it might be easier for you to add it in.

app/src/layout/Combat.tsx
Northern-Fayt: Neon wanted to keep the 500 ryo, although I will utilize this for battle pyramid.

app/src/server/api/routers/profile.ts
Northern-Fayt: There's currently no relationship to battleHistory from userData at least that I can include on With.
Also if I were to add it then all battle history would be pulled making the query quite heavy. Although its a new query, I did make it very light and targeted.

I changed the PvE fights to be publically viewable though.

app/src/utils/time.ts
Northern-Fayt: You should have seen my face when I looked up how to do it xD

PR #176, opened 2024-11-19
Fix Ramen Shop discount, Show only players inside village at hospital

WizzzzardMedia:
# Pull Request
Fix: Right now the ramen discount doesn't get applied. It only shows the discounted price in the UI, but the full, undiscounted price is deducted when you buy the ramen.

Add: Also code always applied the ramen discount of the players home village, it makes more sense if a player visits an allied village that the discount of that village is used instead

Fix: Right now all players in the sector are shown at the hospital for meds to heal them, but only players inside the village can be healed.

Fix: calcIsInVillage missing one location (4, 1)

385248794-4317eb11-eb14-4142-89b5-4f8ab39ffc39
^Should cost 44.5 ryo (50% discount), but 89 is charged

a491e2ca19760d7e1fb34dba6bf8cc3d
^Can't heal user outside the village, but user is still listed...

image
^4,1 still inside village

## License

### DIFF
--- a/app/src/app/hospital/page.tsx
+++ b/app/src/app/hospital/page.tsx
@@ -11,6 +11,7 @@ import Image from "next/image";
 import { hasRequiredRank } from "@/libs/train";
 import { Button } from "@/components/ui/button";
 import { structureBoost } from "@/utils/village";
+import { calcIsInVillage } from "@/libs/travel/controls";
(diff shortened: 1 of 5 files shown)


MathiasGruber:
So sorry for being slow on this, I'm merging, thanks for the contribution! I will add 50 reputation points to your account!

PR #192, opened 2024-11-26
Daily Reset functionality, log resets and check multiple times a day …

rauf543:
…to make sure resets have gone through

# Pull Request

In response to this issue:

https://github.com/MathiasGruber/TheNinjaRPG/issues/191

Haven't had a chance to test it yet fully as couldn't figure out how to edit game-time, but I thought I'd get some feedback on the approach as I don't think I fully understand the code (many of the checks go through twice a day, so something must be limit them which I assume is this:)

export const lockWithGameTimer = async (
  client: DrizzleClient,
  frequency: number,
  unit = "h",
  prefix = "timer",
) => {
  const response = await checkGameTimer(client, frequency, unit, prefix);
  const idx = `${prefix}-${frequency}${unit}`;
  if (!response) {
    await updateGameSetting(client, idx, 0, new Date());
  }
  return response;
};

and the other code in the gamesettings file, but to be honest the logic of it evades me.

Overall my approach might be redundant but I'm open to comments

## License

### DIFF
--- a/app/src/app/api/daily-bank/route.ts
+++ b/app/src/app/api/daily-bank/route.ts
@@ -1,6 +1,6 @@
 import { eq, ne, and, or, sql, lt } from "drizzle-orm";
 import { drizzleDB } from "@/server/db";
-import { userData } from "@/drizzle/schema";
+import { userData, dailyReset } from "@/drizzle/schema";
 import { lockWithGameTimer, handleEndpointError } from "@/libs/gamesettings";
 import { structureBoost } from "@/utils/village";
 import { calcBankInterest } from "@/utils/village";
@@ -86,8 +86,33 @@ export async function GET() {
         ]);
       }),
     );
+
+    // Log the reset into the DailyReset schema
+    await drizzleDB.insert(dailyReset).values({
+      id: crypto.randomUUID(), // Generate a unique ID for the reset entry
+      resetType: "daily-bank",
+      scheduledDate: new Date().toISOString(), // Use current timestamp for scheduled date
+      executedDate: new Date().toISOString(), // Log the execution timestamp
+      status: "completed", // Mark as completed since the reset was successful
+      lastChecked: new Date().toISOString(),
+      retryCount: 0,
+      isManualOverride: false,
+    });
+
     return Response.json(`OK`);
   } catch (cause) {
+    // Log the reset as failed if an error occurs
+    await drizzleDB.insert(dailyReset).values({
+      id: crypto.randomUUID(),
+      resetType: "daily-bank",
+      scheduledDate: new Date().toISOString(),
+      executedDate: null, // No execution timestamp since it failed
+      status: "failed",
+      lastChecked: new Date().toISOString(),
+      retryCount: 0,
+      isManualOverride: false,
+      errorLog: String(cause), // Capture the error for debugging
+    });
     return handleEndpointError(cause);
   }
 }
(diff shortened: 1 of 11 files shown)


MathiasGruber:
Terribly sorry that I only see this now, after I've pushed an alternative fix. I will push comments to this PR to give feedback, it is highly appreciated for people to help out with the code, so I'm truly sorry for this. Please see my commit for the changes:
https://github.com/MathiasGruber/TheNinjaRPG/commit/89a4cefa632d4624ae026f2e5ae8f640cd5d52e1

A bit of explanations;

1. the main issue with the CRON jobs is that apparently we are not guarenteed that they fire - this is why for each daily task, we had two entries in the `vercel.json` file. 

2. it is an issue if a daily job fails for whatever reason - this is however not something we neccesarily need to track explicitly, as we have both Sentry.io and vercel logs which capture these errors.

3. if a given daily job fails mid-way, e.g. the bank-job, it is dangerous as it may have given interest to some players, but not others.

As such the solution I pushed changes:

1. vercel.json to have each daily job run on every 5th minute on hour 0.
2. no changes here, the console.error is enough to get our errors logged and for me to see the reports
3. major change to the bank-update, so it tracks the application of each interest.

I truly appreciate the effort you put into this PR, and I'd like to still award you reputation points for the effort; what is your username?

app/src/app/api/daily-bank/route.ts
```diff
@@ -86,8 +86,33 @@ export async function GET() {
         ]);
       }),
     );
+
+    // Log the reset into the DailyReset schema
+    await drizzleDB.insert(dailyReset).values({
+      id: crypto.randomUUID(), // Generate a unique ID for the reset entry
```
MathiasGruber, line 92: This is theoretically fine, but for consistency we use the nanoid() across the site :)

app/src/app/api/daily-counters/route.ts
```diff
@@ -20,8 +20,30 @@ export async function GET() {
...
+      scheduledDate: new Date().toISOString(), // Use current timestamp for scheduled date
+      executedDate: new Date().toISOString(), // Log the execution timestamp
+      status: "completed", // Mark as completed since the reset was successful
+      lastChecked: new Date().toISOString(),
+      retryCount: 0,
+      isManualOverride: false,
+    });
     return Response.json(`OK`);
   } catch (cause) {
+    await drizzleDB.insert(dailyReset).values({
```
MathiasGruber, line 36: I think all the inserts could probably have been abstracted into a separate function just taking a resetType.. also no need to specify values which already have a default value

app/src/app/api/daily-reset/route.ts
```diff
@@ -0,0 +1,61 @@
+import { drizzleDB } from "@/server/db";
+import { dailyReset } from "@/drizzle/schema";
+import { eq, gte, and } from "drizzle-orm";
+import { lockWithGameTimer, handleEndpointError } from "@/libs/gamesettings";
+
+export async function GET() {
+  try {
+    // Ensure this function doesn't run too frequently
+    const timerCheck = await lockWithGameTimer(drizzleDB, 3, "h", "daily-reset-check");
```
MathiasGruber, line 9: In my commit I replaced this one with a daily timer, which checks if the date has changed.

app/src/validators/dailyReset.ts
```diff
@@ -0,0 +1,34 @@
+import { z } from "zod";
+
+// Define the DailyReset schema
+export const dailyResetSchema = z.object({
```
MathiasGruber, line 4: These can often be inferred directly from the drizzle schema :)

PR #196, opened 2024-11-27
Fix drizzle config and /.env.example folder for deployment

KernelClyde:
# Pull Request

First time doing a pull request. These two errors prevent people from being able to create their own environment.

Add Open AI API to /.env.example. Local TNR website will not load without this. 
Update drizzle.config.ts to use MYSQL_URL instead of DATABASE_URL. make dbpush command will fail. 

## License


MathiasGruber:
Awesome, thank you for the contribution! :)

Issue #217, opened 2024-12-27
Friendly Fire Doesn't Work


theeneon:
**Describe the bug**
The friendly fire settings on Jutsu's in the combat arenas are not working as intended. This is affecting normal combat as well as clan battles. 

**To Reproduce**
1. Add a Jutsu that uses AOE target ground, You can use stun + movement jutsu. The player who casted the jutsu is stunned along with their opponent. 
2. Create a clan battle use an AoE jutsu that is set to ally or enemy, both friendly fire settings(ally, enemy) will work on both allies and enemies alike. i.i Allies will take damage from Jutsu that is set to enemy only while enemies will benefit from Jutsu that is set to ally only. 

**Expected behavior**
1. The player that uses a jutsu should not be affected by the jutsu they casted,
3. Jutsu that are set to ally should not affect the opponent the player is fighting. 
4. Jutsu that are set to enemy should only affect the opponent the player is fighting and not players who are on it's own team.

MathiasGruber:
[OpenHands](https://github.com/All-Hands-AI/OpenHands) started fixing the issue! You can monitor the progress [here](https://github.com/MathiasGruber/TheNinjaRPG/actions/runs/12560344769).

MathiasGruber:
A potential fix has been generated and a draft PR #220 has been created. Please review the changes.

PR #224, opened 2025-01-01
Revert "Revert "Fix issue #217: Friendly Fire Doesn't Work""

MathiasGruber:
Reverts MathiasGruber/TheNinjaRPG#223

This was merged pre-maturely. We need to test this on preview deployment before mergin.


theeneon (commented):
Friendly Fire is fixed. Execept on anything that targets ground or empty ground. You still take damage even when marked as Enemy. This might be isolated to the stun tag in single player combat and most tags in Clan battles

theeneon:
@MathiasGruber Could you assist in getting friendly fire to be turned off in spars?

PR #227, opened 2025-01-02
Fix issue #226: Allow Buff Prevent to prevent Reflect

### LINKED ISSUE #226: Allow Buff Prevent to prevent Reflect
Describe the bug
The buff prevent tag should also prevent users from preventing reflect

To Reproduce
Use action with buff prevent tag in combat. Afterwards target will still be able to use action with reflect prevent tag

Expected behavior
Should not be possible to use reflect prevent tag after being hit with buff prevent tag.

TheNinja-RPG:
This pull request fixes #226.

The issue appears to be successfully resolved based on the AI agent's explanation and test results. The fix implemented:

1. Added a buff prevent check within the reflect function
2. Implemented logic to verify if there's an active buff prevent tag on the target
3. Added timing validation to ensure buff prevent tags applied before reflect attempts will block the reflect action
4. Successfully passed all tests

The implementation follows existing patterns in the codebase for similar buff-type effects, and directly addresses the reported issue where buff prevent tags should block reflect actions. The timing check ensures proper sequencing of effects, which is crucial for combat mechanics.

For a human reviewer, I would summarize:
"This PR implements buff prevent functionality to block reflect actions. When a character is affected by a buff prevent tag, they can no longer use reflect abilities. The implementation includes proper timing checks and follows existing patterns for buff mechanics in the codebase. All tests are passing, confirming the desired behavior is working as expected."


theeneon:
This works and is good to go.

MathiasGruber:
LGTM 👍

PR #303, opened 2025-01-22
Fix issue #300: Buff Prevent and Debuff Prevent bug

### LINKED ISSUE #300: Buff Prevent and Debuff Prevent bug
**Describe the bug**

Buff Prevents and Debuff Prevents are lasting one less round than they should.  If it is supposed to last two rounds, it will only last one round instead.

Battle Log:

#1: With a flicker of their eyes, Phrosfire manipulates the surrounding light, creating multiple afterimages of themselves that confuse and disorient %target. These illusions mirror %target's attacks back at them, leaving them unsure of the true Phrosfire's position. Each reflection seems real until %target is struck by a precise blow from the shadows, capitalizing on their disoriented state.
- Phrosfire cannot be debuffed for the next 2 rounds

Screenshots are included below

**To Reproduce**

1. Go into battle
2. Use any ability that uses a buff prevent or debuff prevent
3. Looks at the effects list and compare it to the log

**Expected behavior**
A buff and debuff should last the stated number of rounds.

**Further Information**
Going through the combat tags, the buffPrevent and debuffPrevent tag were set to change the number of rounds to -= 1 when the effect was new, so it was losing a round when being used.

**Screenshots**
Image

Image

Phrosfire:
Both the buff prevent and debuff prevent were subtracting 1 turn from the total number of turns that the effect should be active if it was a new effect.  I removed the subtraction so that the number of turns should be correct.

# Pull Request

Removed where 1 round would be subtracted from the effect if it was a new effect.

## License

Fix #300 

## Summary by CodeRabbit

- **Bug Fixes**
	- Adjusted buff and debuff duration tracking mechanism to prevent unintended early reduction of effect rounds.

### DIFF
--- a/app/src/libs/combat/tags.ts
+++ b/app/src/libs/combat/tags.ts
@@ -85,9 +85,6 @@ export const buffPrevent = (
   if (mainCheck) {
     const info = getInfo(target, effect, "cannot be buffed");
     effect.power = 100;
-    if (effect.isNew && effect.rounds) {
-      effect.rounds -= 1;
-    }
     return info;
   } else if (effect.isNew) {
     effect.rounds = 0;
@@ -108,9 +105,6 @@ export const debuffPrevent = (
   if (mainCheck) {
     const info = getInfo(target, effect, "cannot be debuffed");
     effect.power = 100;
-    if (effect.isNew && effect.rounds) {
-      effect.rounds -= 1;
-    }
     return info;
   } else if (effect.isNew) {
     effect.rounds = 0;


MathiasGruber:
@theeneon does this operate as expected now? I cannot remember why I did the subtracting, but I expect it was due to some other bug.

theeneon:
@MathiasGruber Yes we're good to go here.

MathiasGruber:
Very cool, I've added 20 reps to your account for this :)

PR #304, opened 2025-01-22
Fix issue #302: Users cannot be cleansed of decreaseHeal tag or healP…

### LINKED ISSUE #302: Users cannot be cleansed of decreaseHeal tag or healPrevent tag.
**Describe the bug**

When a user is effected by decreaseHeal or healPrevent, they are unable to cleanse themselves of the effect.  Attempting to cleanse will not change the effect on them.

healPrevent is listed as a positive effect, meaning the person who used it on you, could remove it from you if they cleared you.

**To Reproduce**

1. Go to battle
2. Be hit by a jutsu tagged with decreaseheal or healprevent
3. Attempt to cleanse yourself
4. Debuff is not removed. 

**Expected behavior**
When a user cleanses themselves, the debuff should be removed.

**Further Information**
healprevent DOES NOT show on the list of current effects in battle, so that is a separate issue

**Screenshots**
1. Jutsu with decreaseheal was used on me.

Image

2. I attempted to cleanse the tag.

Image

3. The tag persisted

Image

Phrosfire:
# Pull Request

The constant DecreaseHealGivenTag was looking for a literal of "decreaseheal" which was correct.  The issue, is in the list for isNegativeUserEffect, instead of "decreaseheal" it was "decreasehealgiven".  I updated this to "decreaseheal" in the list so that it can be properly found when someone is attempting to cleanse "decreaseheal" effects on them.

I also moved "healprevent" from being in the isPositiveUserEffect list to the isNegativeUserEffect list.  With the healprevent being in the wrong list, if a user effected their opponent with it, it was treated as a positive effect.  So if they were cleared by the enemy, it would fall off instead of staying on.  This was also preventing those effected by it from cleansing themselves of the debuff.

Fix #302 

## License

### DIFF
--- a/app/src/libs/combat/types.ts
+++ b/app/src/libs/combat/types.ts
@@ -744,7 +744,6 @@ export const isPositiveUserEffect = (tag: ZodAllTags) => {
       "decreasepoolcost",
       "heal",
       "lifesteal",
-      "healprevent",
       "increasedamagegiven",
       "increaseheal",
       "increasestat",
@@ -776,7 +775,7 @@ export const isNegativeUserEffect = (tag: ZodAllTags) => {
       // "buffprevent",
       "decreasedamagegiven",
       "increasedamagetaken",
-      "decreasehealgiven",
+      "decreaseheal",
       "increasepoolcost",
       "decreasestat",
       "clear",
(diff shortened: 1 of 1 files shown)


MathiasGruber:
Awesome! I've merged and added 20 reputation points to your account!

PR #381, opened 2025-02-17
Fix issue #380: Player being Silenced is still able to Talk in Tavern

### LINKED ISSUE #380: Player being Silenced is still able to Talk in Tavern
**Describe the bug**
When a player is silenced they are still able to speak in the tavern 

**To Reproduce**

1. Silence a player 
2. After Player is Silence they can still speak 

**Expected behavior**
When a player is silenced they should not be able to speak in the tavern

TheNinja-RPG:
This pull request fixes #380.

The issue has been successfully resolved based on the changes made. The core problem was that silenced players could still speak in the tavern because the Conversation component wasn't checking the user's silenced status. The fix directly addresses this by:

1. Adding `!userData.isSilenced` to the conditional check that controls whether the message input field is displayed
2. Placing this check alongside existing conditions (!conversation.isLocked, userData exists, !userData.isBanned)
3. Using the same pattern already proven to work for banned users

The change means the message input field will now be hidden completely when userData.isSilenced is true, making it technically impossible for silenced users to send messages in the tavern. This is exactly what was needed to match the expected behavior described in the issue.

The solution is appropriately minimal and uses existing patterns in the codebase. Since the change prevents the message input from even being rendered for silenced users, it provides a robust fix that can't be circumvented through normal usage of the interface.

### DIFF
--- a/app/src/layout/Conversation.tsx
+++ b/app/src/layout/Conversation.tsx
@@ -242,27 +242,31 @@ const Conversation: React.FC<ConversationProps> = (props) => {
           topRightContent={props.topRightContent}
           onBack={props.onBack}
         >
-          {conversation && !conversation.isLocked && userData && !userData.isBanned && (
-            <div className="relative mb-2">
-              <RichInput
-                id="comment"
-                refreshKey={editorKey}
-                height="120"
-                disabled={isCommenting}
-                placeholder="Write comment..."
-                control={control}
-                error={errors.comment?.message}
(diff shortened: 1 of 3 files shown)


theeneon (changes requested):
@MathiasGruber For whatever reason Silence doesn't work anymore, when a user is silence they are still able to view the Tavern and we have persons who are perma Silence with the issue I told you about getting around it and creating havoc.

PR #394, opened 2025-02-22
New Wallpaper Image

Phrosfire:
# Pull Request

Updating the village wallpaper to a new one.

## License

## Summary by CodeRabbit

- **New Features**
	- Introduced two new wallpaper images for enhanced visual options.
- **Refactor**
	- Updated the layout to consistently display the new wallpaper, disabling the seasonal background rotation and improving display properties.

### DIFF
--- a/app/drizzle/constants.ts
+++ b/app/drizzle/constants.ts
@@ -794,6 +794,8 @@ export const IMG_WALLPAPER_HALLOWEEN =
   "https://utfs.io/f/Hzww9EQvYURJACVJVzoZUC4muiGcQNzjfEndY5y1w20B8hTW";
 export const IMG_WALLPAPER_BRIGHT_VILLAGE =
   "https://i.imgur.com/t1kligo.jpeg";
+export const IMG_WALLPAPER_SHINE_TEST =
+  "https://i.imgur.com/9qlpR52.jpeg";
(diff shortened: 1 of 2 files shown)


MathiasGruber:
Why are we removing the season switcher?

Also, I actually want to move all wallpapers to our bunny.net CDN, which will let them be served faster, and thus optimize our SEO scores :D

Phrosfire:
> Why are we removing the season switcher?
> 
> Also, I actually want to move all wallpapers to our bunny.net CDN, which will let them be served faster, and thus optimize our SEO scores :D

Neon wanted to test out different wallpapers, so I adjusted the code a little bit so he could test them out. The goal also seems to be changing the switcher so that the background is dependent on the user's village.

MathiasGruber:
Perfect!.. I took the code from here and adapted it to be based on an entry on the village table - i.e. easily allowing us to change the background of each village; https://github.com/MathiasGruber/TheNinjaRPG/commit/22bf950a95120af746bb4a484666e1b2ead920e5

So far I've only set the image for shine.. Just pass me any updated wallpapers for the rest of the images, and I'll compress and put them on bunny CDN :D .. awesome work.. especially the css for making this look nice, that is something I always struggle with 😬

Issue #349, opened 2025-01-29
Poison


theeneon:
Create a new Jutsu tag called Poison.

This is a round based tag that does percentage damage to the target based on how much Chakra and Stamina they use.

TheNinja-RPG:
A potential fix has been generated and a draft PR #350 has been created. Please review the changes.

PR #409, opened 2025-02-27
Fix Issue #349: Poison

Phrosfire:
# Pull Request

Adding poison as a new tag

## License

## Summary by CodeRabbit

- **New Features**
	- Enhanced combat visuals with expanded color options for effect feedback.
	- Introduced a poison effect that applies damage based on resource usage.
	- Improved tracking of resource consumption during combat actions for enhanced strategic gameplay.
	- New logic for handling and displaying poison effects in combat interactions.
	- Conditional styling for village name display based on its value.

- **Bug Fixes**
	- Fixed visual feedback issues related to effect colors during combat.


Phrosfire:
@MathiasGruber I have done several commits with this tag and while it "works", it's not in the prettiest way.  It's entirely possible I'm going about this the wrong way.

Poison is technically behaving as it should by dealing damage based on the amount of chakra and stamina spent using actions.  The damage is being handled in action.ts though immediately after the action is used.  I've tried having it handle the calculations in tags.ts instead, but I can't seem to get the chakra and stamina cost to pass properly with each action.  With this setup, I also can't trigger the damage message.

Do you have any suggestions on this?  This tag is a bit different from the rest, and I am stuck.

MathiasGruber:
> @MathiasGruber I have done several commits with this tag and while it "works", it's not in the prettiest way. It's entirely possible I'm going about this the wrong way.
> 
> Poison is technically behaving as it should by dealing damage based on the amount of chakra and stamina spent using actions. The damage is being handled in action.ts though immediately after the action is used. I've tried having it handle the calculations in tags.ts instead, but I can't seem to get the chakra and stamina cost to pass properly with each action. With this setup, I also can't trigger the damage message.
> 
> Do you have any suggestions on this? This tag is a bit different from the rest, and I am stuck.

Haha, yeah, I also had to think extra for this one.. I think I have an advantage in being experienced with the codebase here, but the combat system does need a bigger re-factor to make the flow more easy to understand - this is something I'll get around to eventually. 

Anyways, to stay as consistent with the logic as possible in terms of having tag-logic within the tag functions and use the consequences array, I tried re-structuring a bit as shown in attached diff.. I've not tested it, but I think this would be the best way to go about it. Let me know if I missed something - as mentioned, I didn't test this yet.

[proposed_structure.txt](https://github.com/user-attachments/files/19052597/proposed_structure.txt)

Phrosfire:
@MathiasGruber As long as this change works, this should be good to go.  I made adjustments to see if I can take the effects of increasepoolcost and decreasepoolcost into the damage calculation for poison.  

The initial changes you did worked, but I had to adjust some things so it works with how we wanted it to.  For some reason, Vercel isn't deploying the change for us to test anymore though.

Phrosfire:
Well, the good news is that the tag works.  The bad news is it doesn't work exactly as we want.  It finally deployed.  It's not only using the cost of opponents jutsu.  Its also using the cost of your own jutsu to damage your opponent through poison.

Phrosfire:
This tag is now fully functional and works as expected.   It only took a lot of commits to get it there.

app/src/libs/combat/process.ts
```diff
@@ -98,6 +98,8 @@ export const realizeTag = <T extends BattleEffect>(props: {
   level: number | undefined;
   round?: number;
   barrierAbsorb?: number;
+  cpSpent?: number;
```
MathiasGruber: Looks overall good, but I don't think these two cpSpent and spSpent are being used anymore?

MathiasGruber (changes requested):

app/src/libs/combat/process.ts
Phrosfire: They were no longer being used, I have removed them from here as well as the references in action.ts

MathiasGruber (approved):
Looks good

Issue #414, opened 2025-03-02
Update Jutsu Transfer to add improvements


Phrosfire:
**Is your feature request related to a problem? Please describe.**

Not currently a problem, but this would be an update to improve jutsu level transfers.

**Describe the solution you'd like**

Add the ability to transfer level 25 jutsu.  Max transfer level will be 25 instead of 20.  Pick how many levels to transfer from source jutsu to new jutsu.

PR #415, opened 2025-03-02
Fix Issue #414: Update Jutsu Transfer to add improvements

Phrosfire:
# Pull Request

Change max transfer level to 25 and allow users to select how many levels to transfer.

## License

### DIFF
--- a/app/drizzle/constants.ts
+++ b/app/drizzle/constants.ts
@@ -452,7 +452,7 @@ export const BLOODLINE_ROLL_TYPES = ["NATURAL", "ITEM"] as const;
 // Jutsu level transfer config
 export const JUTSU_TRANSFER_DAYS = 20;
 export const JUTSU_TRANSFER_COST = 20;
-export const JUTSU_TRANSFER_MAX_LEVEL = 20;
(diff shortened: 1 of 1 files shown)


Phrosfire:
In addition to the jutsu transfer updates.  I also gave myself some color, as it was using the default slate for both staff colors.  I also added set up the missions help button to use modal with the ability for event staff to use html in the descriptions.

MathiasGruber:
Isn't this very exploitable? i.e. level 5 jutsus to level 5, transfer all levels to another jutsu, which is then at 25, even though training the 5 first is much easier than training 1 to level 25?

theeneon:
> Isn't this very exploitable? i.e. level 5 jutsus to level 5, transfer all levels to another jutsu, which is then at 25, even though training the 5 first is much easier than training 1 to level 25?

Yea I did think of that, but it is gated behind a rep paywall. Seeing as you only have 5 at max free transfers per month. 
 Could add a minimum level of training for it to be eligible

Phrosfire:
This was updated to have a minimum of level 10 for the source jutsu in order to transfer the levels.

MathiasGruber (approved):
LGTM

Issue #418, opened 2025-03-06
Users are being unsilenced before their silence time end


Phrosfire:
**Describe the bug**

Some users who are silenced are having their silence ending early.  From the code, I can see that the silence is setting banEnd, but it's still ending before that date is met.

**To Reproduce**

1) Silence a user for several days
2) User is unsilenced at daily reset

**Expected behavior**

User stays silenced until the endDate is met.

nanubula:
Are userReports ever cleaned up once the ban/silence ends? I could see if someone has multiple userReports and one silence has already ended, maybe that old report is unsilencing the user [here](https://github.com/MathiasGruber/TheNinjaRPG/blob/476576155c990cfc13ea3f8da7724ab1d3459ff2/app/src/app/api/daily-reports/route.ts#L38).
If userReports aren't cleaned up to keep a history, maybe there should be another banState for reports that are no longer valid and can be ignored when checking expiredReports daily. I could be totally wrong on cause though, just my first thoughts

PR #419, opened 2025-03-06
Fix Issue #418: Users are being unsilenced before their silence time end

Phrosfire:
# Pull Request

Adjusting logic on the daily reports to check if banEnd <= now before it changes isBanned and isSilenced to false.

## License

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced the process for lifting account restrictions, ensuring that bans and silences are accurately removed based on the current time, leading to improved moderation outcomes for users.

### DIFF
--- a/app/src/app/api/daily-reports/route.ts
+++ b/app/src/app/api/daily-reports/route.ts
@@ -37,13 +37,13 @@ export async function GET() {
 
     // Process each expired report
     for (const report of expiredReports) {
-      if (report.status === "BAN_ACTIVATED" && report.reportedUser?.isBanned && report.banEnd <= now) {
+      if (report.status === "BAN_ACTIVATED" && report.reportedUser?.isBanned && (report.banEnd ?? now)  <= now) {
         // Unban user if ban has expired
         await drizzleDB
           .update(userData)
           .set({ isBanned: false })
           .where(eq(userData.userId, report.reportedUser.userId));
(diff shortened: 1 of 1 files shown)


Phrosfire:
The changes I made has it also confirm that the banEnd date <= now.  This should hopefully prevent users from being unbanned/unsilenced ahead of when they are scheduled to be.

app/src/app/api/daily-reports/route.ts
```diff
@@ -36,16 +37,13 @@ export async function GET() {
 
     // Process each expired report
     for (const report of expiredReports) {
-      if (report.status === "BAN_ACTIVATED" && report.reportedUser?.isBanned) {
+      if (report.status === "BAN_ACTIVATED" && report.reportedUser?.isBanned && (report.banEnd ?? now)  <= now) {
```
MathiasGruber, line 40: Isn't this just a repeat from the logic above `lt(userReport.banEnd, new Date())`?.. or am I just dumb since people are actually being unsilenced before their time is up, I'm just wondering what is going on here.

Phrosfire: You're 100% right, it's doing the same thing.  I saw lt(userReport.banEnd, new Date()) and it didn't click that its the same thing.  Back to the drawing board I go.  Mods keep having to apply the ban and silences when it lasts multiple days.

MathiasGruber: Could be the logic on `getBan` endpoint in `@/routers/reports.ts` but I cannot figure out why it would be wrong.

PR #427, opened 2025-03-14
:sparkles: Rewrite of copy & mirror tag solution

MathiasGruber:
# Pull Request
Re-write of the Copy & Mirror tag implementations. 

Previous PRs:
https://github.com/MathiasGruber/TheNinjaRPG/pull/421
https://github.com/MathiasGruber/TheNinjaRPG/pull/329

Issues: 
https://github.com/MathiasGruber/TheNinjaRPG/issues/328
https://github.com/MathiasGruber/TheNinjaRPG/issues/336

## License

## Summary by CodeRabbit

- **New Features**
  - Introduced two new combat effects: **Copy** and **Mirror**.
    - **Copy** allows players to replicate beneficial effects from opponents.
    - **Mirror** enables players to reflect negative effects onto targets.
  - Updated effect ordering ensures these new mechanics integrate seamlessly during battle.

- **Improvements**
  - Enhanced effect processing logic to accommodate new effect types.
  - Simplified visibility logic for submit buttons across various forms, ensuring they are always displayed regardless of changes.


MathiasGruber:
Did a few small adjustments to previous solutions

theeneon (dismissed):
While the text triggers incombat the abilities are not copied or mirrored

MathiasGruber:
> While the text triggers incombat the abilities are not copied or mirrored

I see why this is not working. The solution depends on how we want rounds to work. Say e.g. we set rounds=5 on copy - do we want it to copy positive effects the following rounds (i.e. also new positive effects after the tag was used), or do we want it to copy current positive effects and have them active for the next 5 rounds.

The latter is a lot simpler to implement I think.

theeneon:
> > While the text triggers incombat the abilities are not copied or mirrored
> 
> I see why this is not working. The solution depends on how we want rounds to work. Say e.g. we set rounds=5 on copy - do we want it to copy positive effects the following rounds (i.e. also new positive effects after the tag was used), or do we want it to copy current positive effects and have them active for the next 5 rounds.
> 
> The latter is a lot simpler to implement I think.

Its the latter that we would want to be implemented actually.

Phrosfire:
> > > While the text triggers incombat the abilities are not copied or mirrored
> > 
> > 
> > I see why this is not working. The solution depends on how we want rounds to work. Say e.g. we set rounds=5 on copy - do we want it to copy positive effects the following rounds (i.e. also new positive effects after the tag was used), or do we want it to copy current positive effects and have them active for the next 5 rounds.
> > The latter is a lot simpler to implement I think.
> 
> Its the latter that we would want to be implemented actually.

I switched copiedEffect.rounds = 1 to be copiedEffect.rounds = effect.rounds so that it would use the number of rounds from the tag itself, but it didn't change the outcome.  

I also added a return at the end of the else statement so we could know for sure that the else (where the effectcopy is happening) is triggering.  The return message is coming through, which means the foreach loop is being reached as well.  It just doesn't seem to be actually pushing the positive effects to the user.

MathiasGruber:
I'll work to fix it.. the reason is that we loop through the effects with forEach, but that does not account for new tags being added into the usersEffects

MathiasGruber:
@theeneon took a bit of a larger refactor, but I think it works as intended now

theeneon:
@MathiasGruber approve preemptively so that we can merge it when Vercel is done.

theeneon:
@MathiasGruber So tag works but is bugged, The effects are applying on every action which leads to a stacking issue Look below:

Mirror Test https://pastebin.com/iar3jJT1

As you can see consistent damage before the tag is applied after the tag is applied on every action the damage increases, in my side bar i had this:
image

Instead of 30% debuff I had 90% debuff, when the tag was first used it was at 60% which means it's taking the 30% for 2 rounds and applying it all one round which leads to the 60%. I end my turn and it stacks 30% again which leads to the 90%. 

What should happen here is that the 30% is applied to me for the duration of the mirror tag but the effects shouldn't stack or be applied every action. 

Copy https://pastebin.com/DWeQYrKs

Now in this tag as you can see here, every time i moved the healing was applied which strengthens whats happening with mirror, that the effects that are copied are applying on each action. The status bar also shows this:

Movement 1: 
image

Movement 2:
image

Now this one is even more fun, Madcrep moved and I moved which also doubled the buff:
image

All of that ends with a beautiful one hit Madcrep takes 99772.93 damage 

So i believe once you're able to prevent the effects from applying on every action and only once per round like all other effects we're golden.

MathiasGruber:
> @MathiasGruber So tag works but is bugged, The effects are applying on every action which leads to a stacking issue Look below:
> 
> Mirror Test https://pastebin.com/iar3jJT1
> 
> As you can see consistent damage before the tag is applied after the tag is applied on every action the damage increases, in my side bar i had this: image
> 
> Instead of 30% debuff I had 90% debuff, when the tag was first used it was at 60% which means it's taking the 30% for 2 rounds and applying it all one round which leads to the 60%. I end my turn and it stacks 30% again which leads to the 90%.
> 
> What should happen here is that the 30% is applied to me for the duration of the mirror tag but the effects shouldn't stack or be applied every action.
> 
> Copy https://pastebin.com/DWeQYrKs
> 
> Now in this tag as you can see here, every time i moved the healing was applied which strengthens whats happening with mirror, that the effects that are copied are applying on each action. The status bar also shows this:
> 
> Movement 1: image
> 
> Movement 2: image
> 
> Now this one is even more fun, Madcrep moved and I moved which also doubled the buff: image
> 
> All of that ends with a beautiful one hit Madcrep takes 99772.93 damage
> 
> So i believe once you're able to prevent the effects from applying on every action and only once per round like all other effects we're golden.

Ah, I see, I think I've fixed it not so that they are not re-applied if already there. Works?

PR #434, opened 2025-03-15
Fix Issue #432: Drain/poison now puts you into the negatives for pools

### LINKED ISSUE #432: Drain/poison now puts you into the negatives for pools
**Describe the bug**

The drain tag seems to no longer have logic preventing your pools from going negative.  Poison as the same issue. The lowest pools should go is 0.  When the pool is negative, users are not able to end their turn.  If the user is poisoned and goes below 0 health from it, the battle gets out of sync and they get stuck.

**To Reproduce**

1) Go into battle
2) Get drained/poisoned below 0 health/chakra/stamina.

**Expected behavior**

The amount of current pools stops at 0

Phrosfire:
# Pull Request

Adjusts logic so that pools cannot go below 0. Fix #432 

## License

By making this pull request, I confirm that I have the right to waive copyright and related rights to my contribution and agree that all copyright and related rights in my contributions are waived, and I acknowledge that the Studie-Tech ApS organization has the copyright to use and modify my contribution for perpetuity.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined how character energy is adjusted during drain effects, ensuring stamina, chakra, and health values remain within valid limits.

### DIFF
--- a/app/src/libs/combat/process.ts
+++ b/app/src/libs/combat/process.ts
@@ -584,8 +584,7 @@ export const applyEffects = (
           });
         }
         if (c.poison && c.poison > 0) {
-          target.curHealth -= c.poison;
-          target.curHealth = Math.min(target.maxHealth, target.curHealth);
+          target.curHealth = Math.max(0, Math.min(target.maxHealth, target.curHealth - c.poison));
           actionEffects.push({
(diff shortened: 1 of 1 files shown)


MathiasGruber (approved):
LGTM

PR #436, opened 2025-03-17
Fix XP Variance in Combat Rewards

nanubula:
# Pull Request

Combat that grants XP gives seemingly random amounts and eventually 0 XP for long combats of any kind (20+ rounds, sometimes 30+).

Here's a graph to show how length of a match affects the XP gains with the old method, where common X axis values would often be between 1000 and 3000 for participants wearing armor and using buff jutsu (explanation below). Input values `pre distribution xp` are reasonable examples of the XP that is currently determined before distributing stats, with the Y axis being final resulting XP after distributing stats.
image

The reason for the commonly large X values (1-3k) is because for each buff on armor equipped and jutsu buff per stat counts in the `usedStats` or `usedGenerals`, and counts per action. This can easily be ~50 stats+gens per action, with multiple actions per round.

The resulting variance is due to the usage of math.floor which makes this a stepwise function, which when considering the distribution makes the final formula `Math.floor((exp/totalStatsUsed) * 100) / 100 * totalStatsUsed` and due to the large `totalStatsUsed`, final experience will always eventually approach 0.
Proposed change still uses `Math.floor` but in a loop of 12 instead of thousands, and still has slight variance but only about 0.1-0.2 instead of 10-30.

Also added a vscode launch.json debug config for easy debugging/attaching to server side code if bun is setup locally.
image

Is Arena XP supposed to be impacted by PVP streak? It looks like it currently is, wasn't sure if that's intended.

## License

## Summary by CodeRabbit

- **Chores**
  - Introduced a new configuration to streamline debugging for server-side operations.
- **Refactor**
  - Enhanced the combat system’s logic for tracking usage statistics and distributing experience points, resulting in more accurate gameplay outcomes.
- **New Features**
  - Implemented a more structured representation for tracking used stats and generals, improving efficiency and accuracy in gameplay.
  - Added a new helper function for distributing experience points based on usage.


app/src/libs/combat/util.ts
```diff
@@ -686,38 +686,53 @@ export const calcBattleResult = (battle: CompleteBattle, userId: string) => {
...
         let assignedExp = 0;
-        const gain = Math.floor((experience / total) * 100) / 100;
         const stats_cap = USER_CAPS[user.rank].STATS_CAP;
         const gens_cap = USER_CAPS[user.rank].GENS_CAP;
-        user.usedStats.forEach((stat) => {
-          const value = user[stat] + gain > stats_cap ? stats_cap - user[stat] : gain;
-          result[stat] += value;
-          assignedExp += value;
+
+        const distributeStats = (
```
MathiasGruber: I've only recently learned of the dangers and downsides of closures in javascript - I think we should take this out as a pure function instead :)

MathiasGruber (changes requested):
Looks good and makes a lot of sense. Getting rid of the closure and @theeneon  approving, then we merge :) .. if @theeneon agrees, I'd say we also invite you as a contributor to the repo, which will make using the codebase & running all tests on PRs easier 💪

theeneon:
@nanubula So for a quick summary you're stating that when combat goes +20 rounds the exp that it gives varies correct?

nanubula:
> @nanubula So for a quick summary you're stating that when combat goes +20 rounds the exp that it gives varies correct?

If you are wearing armor with a number of stats listed on it and using buff jutsus in combat, it can vary by a few xp after just 1 round, but that variance increases as the rounds go on. After 20-30 rounds it will drop to 0 xp (this happens to me often in combat lol). If you're not wearing any armor and not using buff jutsu, it would probably take hundreds of rounds to approach 0 xp.

theeneon:
> > @nanubula So for a quick summary you're stating that when combat goes +20 rounds the exp that it gives varies correct?
> 
> If you are wearing armor with a number of stats listed on it and using buff jutsus in combat, it can vary by a few xp after just 1 round, but that variance increases as the rounds go on. After 20-30 rounds it will drop to 0 xp (this happens to me often in combat lol). If you're not wearing any armor and not using buff jutsu, it would probably take hundreds of rounds to approach 0 xp.

Okay I understand, this might be a larger issue at play though outside of this.

nanubula:
> > > @nanubula So for a quick summary you're stating that when combat goes +20 rounds the exp that it gives varies correct?
> > 
> > 
> > If you are wearing armor with a number of stats listed on it and using buff jutsus in combat, it can vary by a few xp after just 1 round, but that variance increases as the rounds go on. After 20-30 rounds it will drop to 0 xp (this happens to me often in combat lol). If you're not wearing any armor and not using buff jutsu, it would probably take hundreds of rounds to approach 0 xp.
> 
> Okay I understand, this might be a larger issue at play though outside of this.

If you mean the logic for determining stat usage isn't accurate, maybe. Could filter out armor/BL effects at least if you don't want that to impact stat usage (and xp distribution). Then in BA for defense you would only get Tai stats unless you used defensive buff jutsu that has other stats listed.

MathiasGruber:
@theeneon I've approved the dev deployment so it should be possible to test now

nanubula:
Let me know if there's anything else needed on this one, thanks!

PR #442, opened 2025-03-19
Add a delete option for Achievements

theeneon:
Add a delete option for Achievements that are in a players logbook 

## Summary by CodeRabbit

- **New Features**
	- The quest log now shows only active challenges by filtering out completed quests, offering a more focused view of pending tasks.

### DIFF
--- a/app/src/layout/Logbook.tsx
+++ b/app/src/layout/Logbook.tsx
@@ -77,6 +77,7 @@ const LogbookAchievements: React.FC = () => {
     <div className="">
       {userData?.userQuests
         .filter((uq) => ["tier", "achievement"].includes(uq.quest.questType))
+        .filter((uq) => uq.completed === 0)
         ?.map((uq, i) => {
           const tracker = userData?.questData?.find((q) => q.id === uq.questId);
           return (
--- a/app/src/server/api/routers/quests.ts
+++ b/app/src/server/api/routers/quests.ts
@@ -350,7 +350,10 @@ export const questsRouter = createTRPCRouter({
         if (questData.questRank !== "A") {
           return errorResponse(`Only A rank missions/crimes are allowed`);
         }
-        if (!user.isOutlaw && !canAccessStructure(user, "/missionhall", sectorVillage)) {
+        if (
+          !user.isOutlaw &&
+          !canAccessStructure(user, "/missionhall", sectorVillage)
+        ) {
           return errorResponse("Must be in your allied village to start quest");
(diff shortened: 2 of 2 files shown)


theeneon:
@MathiasGruber  When trying to delete the achievement this error comes up 
image

MathiasGruber:
Fixed up the solution; however, there is a fundamental issue with doing this, as when you delete an achievement, it is automatically re-entered for the user. The reason for this is that achievements need to be on the user in order to be tracked. Are we looking more for a frontend thing here to toggle active/available vs. completed achievements?

theeneon:
> Fixed up the solution; however, there is a fundamental issue with doing this, as when you delete an achievement, it is automatically re-entered for the user. The reason for this is that achievements need to be on the user in order to be tracked. Are we looking more for a frontend thing here to toggle active/available vs. completed achievements?

Yes as completed achievements for whatever reason don't leave a players logbook, it just stays there all ticked:

image

MathiasGruber:
> > Fixed up the solution; however, there is a fundamental issue with doing this, as when you delete an achievement, it is automatically re-entered for the user. The reason for this is that achievements need to be on the user in order to be tracked. Are we looking more for a frontend thing here to toggle active/available vs. completed achievements?
> 
> Yes as completed achievements for whatever reason don't leave a players logbook, it just stays there all ticked:
> 
> image

Should I just make it so that finished achievements do not show in the logbook? I guess if each achievement gives a badge, then the user can always see which achievements they completed based off their badges.

theeneon:
> > > Fixed up the solution; however, there is a fundamental issue with doing this, as when you delete an achievement, it is automatically re-entered for the user. The reason for this is that achievements need to be on the user in order to be tracked. Are we looking more for a frontend thing here to toggle active/available vs. completed achievements?
> > 
> > 
> > Yes as completed achievements for whatever reason don't leave a players logbook, it just stays there all ticked:
> > image
> 
> Should I just make it so that finished achievements do not show in the logbook? I guess if each achievement gives a badge, then the user can always see which achievements they completed based off their badges.

That would be great then as that was the entire issue.

MathiasGruber:
Should work now in dev deployment

MathiasGruber (approved):
LGTM

theeneon:
@MathiasGruber Not sure you're aware but 
image
 We're getting this on Live any time we make post in the tavern or anywhere. We also got message saying we have reached our limit check your account.

PR #448, opened 2025-03-22
Home Final Attempt

theeneon:
In the Overview section of the Home page Players will be able to upgrade their homes.

Each upgrade offers Item storage and an increase in Regeneration, players can also downgrade their homes

Players can take items from their inventory and story it in their homes item storage

Each upgrade will cost Ryo, the upgrades are as follows:

One Bed Room Apartment this offers +20 Regen, +5 Item Storage Cost 3,000,000 Ryo

Studio Apartment +50 Regen, +10 Item Storage Cost 10,000,000 Ryo

Two Bed Room House + 70 Regen +15 Item Storage Cost 20,000,000 Ryo

Town House + 100 Regen + 20 Item Storage Cost 35,000,000 Ryo

Small Mansion +130 Regen + 25 Item Storage Cost 45,000,000 Ryo

Small Estate + 150 Regen + 30 Item Storage Cost 60,000,000 Ryo

Large Estate 200 Regen + 40 Item Storage Cost 100,000,000 Ryo


theeneon:
@MathiasGruber Could you look at this for me and fix it.

theeneon:
> I redesigned the database schema to have a storedAtHome boolean on UserItem, which is much more efficient than adding a json field on userData (which is retrieved on each pageload/api call)..
> 
> Also updated the client component a bit to be more in-line with other components/pages. For me looks good now, but maybe good to test before merge.

A question I wanted to ask, in the future we wanted to add the ability to decorate your homes for rep, using the 3D implementation, do you think this is possible going forward?

theeneon:
@Phrosfire Tested to ensure it all works, can merge.

PR #464, opened 2025-04-12
Fix Issue #465: GCD should be split by tag

Phrosfire:
# Pull Request

With the addition of the gcd on actions, using a single tag in the gcd list pull all other tags listed on cooldown as well.  This change splits it to handle only the same tag.

In addition to this fix, I also adjust pierce to go through shields, reflect doesn't affect shields, Pierce cannot be recoiled, and pierce can be reflected.  Drain can now affect health as well. The profile shows how long you have been in a village for.  Battle actions now display an actual number showing how much longer they are on cooldown for.

Also contains the solution for the heal tag changes.  Static heals have a multiplier of 10 which allows it to have better heal scaling.  Basic heal is now a static heal instead of percentages.

Fix issue #465 
Fix issue #467 

## License

## Summary by CodeRabbit

- **New Features**
  - Introduced a configurable limit for equipping piercing techniques, with dynamic feedback when limits are exceeded.
  - Expanded support for additional effect types in visual components.
  - Added display of membership duration in the village for user profiles.
  - Enhanced cooldown display to show remaining time for actions.
- **Improvements**
  - Refined combat mechanics with enhanced shared cooldown handling and updated damage calculations for specific effects, including special handling for pierce damage.
  - Adjusted drain effects to impact additional pools, including health.
  - Enhanced type definitions for effects, allowing for broader tagging capabilities.
  - Increased combat healing percentages based on healer ranks.
- **Bug Fixes**
  - Improved readability of combat history color coding without changing visuals.

### DIFF
--- a/app/drizzle/constants.ts
+++ b/app/drizzle/constants.ts
@@ -323,6 +323,7 @@ export const TrainingSpeeds = ["15min", "1hr", "4hrs", "8hrs"] as const;
 export type TrainingSpeed = (typeof TrainingSpeeds)[number];
 
 export const JUTSU_MAX_RESIDUAL_EQUIPPED = 4;
+export const JUTSU_MAX_PIERCE_EQUIPPED = 9999;
 
 export const UserAssociations = ["MARRIAGE", "DIVORCED"] as const;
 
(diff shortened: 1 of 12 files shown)


Phrosfire:
@MathiasGruber When you see this, can you also check your DM from me please in discord?  It involves a bug I didn't want to touch on too much in the public.

theeneon:
> Refactored the solution a little bit; mainly simplified logic in performBattleAction a little bit, and also refactored the shield reduction logic to be before we do consequence collapse - i.e. if we have two damage conseuqnces, they are first dealt with wrt. shields before the collapse, so that we can separate out the pierce logic for each. Once this has been tested @theeneon we can merge

Please proceed.

theeneon:
@MathiasGruber I think you forgot to approve this.

MathiasGruber (approved):
Forgot to do this before, we 🛳️ then

PR #471, opened 2025-04-16
Allow jutsu reskins and link jutsu to parent

Phrosfire:
# Pull Request

This allows jutsu to be reskinned and also link a jutsu to a parent.  Updating a parent jutsu will update all children jutsu.  When a jutsu is reskinned, it copies all data about the parent jutsu to the reskin copy including level and experience.  The jutsu is automatically inserted into the list of learned jutsu.  I set the training filter so it'll only show the "Reskin" type if the user has the jutsu already.

When a new jutsu is created, whether by staff or a player creating a reskin, it will display who created it.  In the below screenshot it is a reskin, so "createdBy displays who reskinned it.

Limit on the number of reskins starts at 3(defined in constants), which can be increased by consuming a token (consumed count is in UserData).  It checks the number of reskins the user has done and if its less than the allowed and the user has enough rep, the reskin succeeds.  The screen will also display how many reskins are remaining.  I just can't upload the pictures until vercel finishes the push.

All that's left is having the edit screen properly list all parent jutsu, not just a chunk of them.  Speaking with content, we think it would be a good idea to have the report api used in the tavern also check to make sure that the descriptions being used are safe.

Current images of the reskin view.  This may be adjusted as I continue.

image

image

image

## License

## Summary by CodeRabbit

* **New Features**
  * Full reskin system: create/update/remove jutsu reskins (limits, costs, free for authorized roles); staff-managed bloodline reskins; select bloodline reskins in profile; reskinned jutsu and bloodlines applied across UI and combat.
  * Staff tools: reskin directories and editors with infinite scroll, filtering, edit flows, and moderation.
  * Consumables and items to increase reskin slots; permission helpers for reskin roles.

* **Style**
  * Icon-only action buttons with hover tooltips; palette icon denotes reskinned items; bloodline popovers and "Created By" on item cards.

* **Chores**
  * Added Makefile force-push-staging target.

### DIFF
--- a/app/drizzle/schema.ts
+++ b/app/drizzle/schema.ts
@@ -420,11 +420,53 @@ export const bloodline = mysqlTable(
 export type Bloodline = InferSelectModel<typeof bloodline>;
 export type BloodlineRank = Bloodline["rank"];
 
+// Bloodline reskins are curated by staff and can be assigned to users
+export const bloodlineReskin = mysqlTable(
+  "BloodlineReskin",
+  {
+    id: varchar("id", { length: 191 }).primaryKey().notNull(),
+    bloodlineId: varchar("bloodlineId", { length: 191 }).notNull(),
+    name: varchar("name", { length: 191 }).notNull(),
+    description: text("description").notNull(),
+    image: varchar("image", { length: 191 }).notNull(),
+    createdBy: varchar("createdBy", { length: 191 }).notNull(),
+    createdAt: datetime("createdAt", { mode: "date", fsp: 3 })
+      .default(sql`(CURRENT_TIMESTAMP(3))`)
+      .notNull(),
+    updatedAt: datetime("updatedAt", { mode: "date", fsp: 3 })
+      .default(sql`(CURRENT_TIMESTAMP(3))`)
+      .notNull(),
+  },
+  (table) => {
+    return {
+      bloodlineIdIdx: index("BloodlineReskin_bloodlineId_idx").on(table.bloodlineId),
+      createdByIdx: index("BloodlineReskin_createdBy_idx").on(table.createdBy),
+      bloodlineNameKey: uniqueIndex("BloodlineReskin_bloodlineId_name_key").on(
+        table.bloodlineId,
+        table.name,
+      ),
+    };
+  },
+);
+export type BloodlineReskin = InferSelectModel<typeof bloodlineReskin>;
+
 export const bloodlineRelations = relations(bloodline, ({ one, many }) => ({
   users: many(userData),
   village: one(village, {
     fields: [bloodline.villageId],
     references: [village.id],
+  }),
+  reskins: many(bloodlineReskin),
+}));
+
+export const bloodlineReskinRelations = relations(bloodlineReskin, ({ one }) => ({
+  bloodline: one(bloodline, {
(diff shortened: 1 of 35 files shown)


Phrosfire:
@MathiasGruber I'm running into an issue where the Parent Jutsu dropdown isn't loading all of the jutsu in to pick from.  It's only able to pull a chunk of the jutsu.  When you have time, could you provide me with some assistance or push me in the right direction please?  I'd also like to have the api used for messages check the fields during creation to make sure there is nothing harmful during the reskin creation.

I appreciate any assistance you can provide!

app/src/app/manual/jutsu/edit/[jutsuid]/page.tsx
```diff
@@ -63,6 +66,50 @@ const SingleEditJutsu: React.FC<SingleEditJutsuProps> = (props) => {
...
+  const [lastElement, setLastElement] = useState<HTMLDivElement | null>(null);
+
+  // Query jutsus based on search
+  const {
+    data: searchResults,
+    fetchNextPage,
+    hasNextPage,
+  } = api.jutsu.getAll.useInfiniteQuery(
+    { 
+      limit: 20,
```
Comment (MathiasGruber) [inline app/src/app/manual/jutsu/edit/[jutsuid]/page.tsx]: I think we should just increase this to e.g. 50, and ignore all pagination here - i.e. when you're trying to re-skin, I'd imagine you know which jutsu name you want to reskin. Maybe a small info box that it only shows top50 results

app/drizzle/schema.ts
```diff
@@ -994,13 +995,15 @@ export const jutsu = mysqlTable(
     villageId: varchar("villageId", { length: 191 }),
     method: mysqlEnum("method", consts.AttackMethods).default("SINGLE").notNull(),
     hidden: boolean("hidden").default(false).notNull(),
+    parentJutsuId: varchar("parentJutsuId", { length: 191 }).default("").notNull(),
```
MathiasGruber: This might be a little annoying, and I'm sorry about that, but I think I'd prefer we introduce a "JutsuReskin" or "JutsuSkin" table, instead of storing reskins in the jutsu table. Reasons:

**Single responsibility & clarity**
The Jutsu table remains focused on the immutable, game‑mechanic aspects of a jutsu (energy cost, effects, cooldown, etc.). All presentation‑only or cosmetic data lives in JutsuReskin, so each table has one clear purpose.

**Easier indexing & smaller row size**
Narrow tables = better I/O. By keeping the core table slim, indexes stay smaller and faster, and caching is more effective.

**Independent evolution**
As we extend “reskin” features—adding alternate icons, shaders, animations, lore text we can ALTER JutsuReskin without touching the hot‑path Jutsu schema. This reduces migration risk and deployment blast radius.

**Clearer ownership & permissions**
i.e. the JutsuReskin table will have a userId column & a jutsu column.. userJutsu table will get a nullable reference to the reskin currently active.

**Simplified queries**
Queries that only need the game‑logic fields don’t have to JOIN or filter out extra columns. And when you do need the skin, you explicitly JOIN JutsuReskin, making your intent clearer. I fear we may have multiple instances otherwise where reskinned jutsus are "leaked" into unintended functionality.

MathiasGruber (changes requested):

MathiasGruber (approved):
LGTM now

PR #537, opened 2025-06-25
Bug fixes

Phrosfire:
# Pull Request

- Fix applied to GCD so it always applies the 3 turn cooldown when necessary.

## License

### DIFF
--- a/app/src/libs/combat/actions.ts
+++ b/app/src/libs/combat/actions.ts
@@ -697,6 +697,23 @@ export const performBattleAction = (props: {
 
   // Update the action state, so as keep state for technique cooldowns
   if (action.cooldown && action.cooldown > 0) {
+    // Restore original cooldowns when GCD expires (run this BEFORE applying new cooldowns)
+    user.jutsus.forEach((jutsu) => {
+      if (jutsu.jutsu.cooldown === 3 && jutsu.lastUsedRound && (jutsu.jutsu as any).originalCooldown !== undefined) {
+        if (battle.round - jutsu.lastUsedRound >= 3) {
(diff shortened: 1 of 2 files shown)


MathiasGruber:
Will review this evening or tomorrow 👍 looks good on first skim through

app/src/libs/combat/util.ts
```diff
@@ -582,13 +582,15 @@ export const calcBattleResult = (battle: CompleteBattle, userId: string) => {
...
+      let experience = didWin
+        ? (battleType === "COMBAT" ? 300 : eloDiff) * expBoost
+        : 0;
       const streakBonus = 1 + user.pvpStreak * 0.05; // 5% per streak
       if (["COMBAT", "TOURNAMENT"].includes(battleType)) {
         experience *= 1.5;
         if (battleType === "COMBAT") {
           experience *= streakBonus;
-          experience = Math.min(experience, 100);
+          experience = Math.min(experience, 500);
```
MathiasGruber, line 593: This is dangerous, as it's inviting cheating, because it's skipping the previously used `maxGain` we're using in the elo calculation, which is scaled with the `battle.rewardScaling` (how many times players have attacked each other continously. This means, in this new implementation, users can continously attack each other, gaining a minimum of 500exp on every battle. I reckon we should refactor it so that we always scale experience according to how many battles the user has been in with the given user. I've added this change.

MathiasGruber (approved):
I refactored a bit and cleaned a few things (mostly cleaning the retry endpoint + logbook changes which was broken on certain layouts). And then the experience scaling. To me looks good now.

PR #562, opened 2025-07-11
Addition bug fixes and improvements

Phrosfire:
# Pull Request

- Properly apply gcd on cooldowns with less than 3 turns remaining on active cooldown
- User can no longer target barriers with any justu that have the move effect.
- Barriers were causing users to take less residual damage even when barriers are gone.  Adjusted it so barriers no longer effect resid.  **There may be a better way to do this, but this was the only solution I found that worked.**
- Prevent residuals from dealing instant damage to barriers.  **They still have an issue of triggering damage with every single action starting the next turn.**
- Hide creator name and hunter names from API besides staff for the bounty board.  It will still displayer hunter ID
- Hide completed bounties from the bounty board
- Allow users to stop tracking a bounty.

I was able to mostly fix up the bugs here.  The only ones not fully working as I had hoped both involve residuals and barriers.  One of the bugs causes residuals to continue to deal less damage after the barrier disappears.  The other causes every action taken (Now after the initial cast turn) after a residual is used to cause damage.

R1: Enemy places Barrier
R1: Use residual on barrier
R2: Moves 3 times, barrier takes residual damage 3 times.

I'm not sure if you want to approve and release it in its current state or not.  As of right now, barriers in this PR are in a better state than they are on live.

## License

## Summary by CodeRabbit

* **New Features**
  * Added the ability for users to stop tracking bounties directly from the bounty board with a new "Stop Tracking" button.
  * Introduced a notification toast when tracking is stopped.
* **Bug Fixes**
  * Improved handling of barrier targeting and cooldown logic in combat, refining which actions can interact with barriers and how cooldowns are applied.
  * Enhanced damage application rules to distinguish between instant and residual damage, ensuring correct barrier absorption.
* **Chores**
  * Updated bounty board data visibility based on user permissions and improved safety in handling optional fields.

### DIFF
--- a/app/src/layout/BountyBoard.tsx
+++ b/app/src/layout/BountyBoard.tsx
@@ -29,14 +29,6 @@ import {
 import type { z } from "zod";
 import type { UserWithRelations } from "@/server/api/routers/profile";
 
-// Type for hunter user data
-type HunterUser = {
-  username: string;
-  level: number;
-  rank: "STUDENT" | "GENIN" | "CHUNIN" | "JONIN" | "ELITE JONIN" | "ELDER" | "NONE";
-  isOutlaw: boolean;
-};
-
 type CreateBountyFormData = z.infer<typeof createBountySchema>;
 type UserSearchFormData = z.infer<ReturnType<typeof getSearchValidator>>;
 
@@ -275,14 +267,14 @@ export default function BountyBoard({ userData }: BountyBoardProps) {
           ) : undefined,
           huntingInfo: b.huntingUsers ? (
             <div className="space-y-1">
-              {(b.huntingUsers as HunterUser[]).map((hunter, idx) => (
+              {b.huntingUsers?.map((hunter, idx) => (
                 <div key={idx} className="text-sm">
-                  <p className="font-medium">{hunter.username}</p>
+                  <p className="font-medium">{hunter?.username ?? "Unknown User"}</p>
                   <p className="text-xs text-gray-600">
-                    Lvl. {hunter.level}{" "}
+                    Lvl. {hunter?.level ?? "Unknown"}{" "}
                     {showUserRank({
-                      rank: hunter.rank,
-                      isOutlaw: hunter.isOutlaw,
+                      rank: hunter?.rank ?? "NONE",
+                      isOutlaw: hunter?.isOutlaw ?? false,
                     })}
                   </p>
                 </div>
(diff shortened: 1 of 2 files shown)


app/src/layout/BountyBoard.tsx
```diff
@@ -29,6 +29,14 @@ import {
 import type { z } from "zod";
 import type { UserWithRelations } from "@/server/api/routers/profile";
 
+// Type for hunter user data
+type HunterUser = {
```
MathiasGruber: Why do we need an explicit type & cast here?

app/src/libs/combat/process.ts
```diff
@@ -585,11 +585,17 @@ export const applySingleEffect = (
...
     curUser
   ) {
-    const result = damageBarrier(newGroundEffects, curUser, effect, config);
-    if (result) {
-      longitude = result.barrier.longitude;
-      latitude = result.barrier.latitude;
-      actionEffects.push(result.info);
+    // For barrier damage, only apply if target is the actor (not if effect is new)
+    // This prevents residual damage from applying to barriers on every action
+    const isTarget = effect.targetId === actorId;
```
MathiasGruber, line 590: Is this the part that doesn't fully work? since barriers are not actors?

app/src/server/api/routers/bounty.ts
```diff
@@ -28,38 +29,69 @@ export const bountyRouter = createTRPCRouter({
...
-            amountRyo: true,
-            createdAt: true,
-            status: true,
-            collectedAt: true,
-            claimedAt: true,
-            targetUserId: true,
-            claimedByUserId: true,
-            creatorUserId: true,
+      // Fetch user info first to determine permissions
+      const currentUser = await fetchUser(ctx.drizzle, userId);
```
MathiasGruber: Why change it to include 2 roundabouts to the database? Fetching both in one roundabout, then filtering out unneeded afterwards should be faster?.. each sequential call to the DB adds processing time.

app/src/layout/BountyBoard.tsx
MathiasGruber: I've fixed this

app/src/server/api/routers/bounty.ts
MathiasGruber: Fixed this

MathiasGruber (approved):
LGTM now 👍

PR #591, opened 2025-07-19
Qol Bug Update

Phrosfire:
# Pull Request

- Allow Reflect to stack and cap at 60%
- Cap Lifesteal at 60%
- Fixed Absorb
- Allow Staff to clear a tracked bounty
- Prevent Skill Tree passives from being able to be cleared
- Anbu Kills should Reward: 500 Token, 300 Prestige 5 Anbu Point (Anbu points require difference to be within 10 levels)
- Normal PVP should offer 300 Token 150 Prestige
- In elemental Roll, allow player to select primary or secondary element to roll. Ensure that player can't get same element twice until they cycle through all elements
- Med Nin will get a medical mission of a lower tier if none are available for them
- Capped Medical missions at 9 and split them off from regular missions.
- Changed hour prestige cost of locking kage challenges.
- Max challenge lock daily set to 12 hours.  Resets with daily reset.
- Change 24-hour kage lockout to 5 days
- Reapply kage lockout to changing village boards
- Kage cannot bypass anbu disband cost by kicking the last anbu member.  If only one member remains, they must use the disband button.
- Death penalties occur for kage regardless of level difference.

Note:

Thanks for telling me how to manually run the hourly timers!  That will make troubleshooting things way easier when they are effected by the cronjobs.

This is a way bigger PR than I generally try to do.  The changes for kage were a later additional after the initial PR was built.

## License

## Summary by CodeRabbit

* **New Features**
  * Added support for medical missions with daily limits, rank-based availability, and UI integration in the mission hall.
  * Introduced a "Remove All Trackers" action for staff on open bounties to mass-remove bounty trackers.
  * Added daily locked challenge time tracking and limits for Kage users with UI display and auto-unlock logic.
  * Added separate reroll options for primary and secondary elements with individual confirmation dialogs.

* **Enhancements**
  * PvP rewards updated to grant tokens, prestige, and ANBU points, with higher rewards for ANBU members.
  * ANBU squad points increment based on PvP results.
  * Quest start logic improved with medical rank filtering and daily limits for errands and medical missions.
  * Enhanced reroll element tracking with reset logic to avoid repeating previously rolled elements.
  * Updated prestige loss for Kage challenges to use a minimum plus percentage-based penalty instead of fixed hourly loss.
  * Kage challenge toggling cooldown now enforced via action log with daily locked time limit preventing excessive locking.
  * Reduced reroll cost and adjusted various cooldown and delay durations for Kage and ANBU actions.
  * Lowered legendary item drop chance for certain gathering ranks.
  * Added "afterburn" tag to shared cooldown tags.

* **Bug Fixes**
  * Reflect, lifesteal, and absorb effects capped at 60% of damage before shields.
  * Status effects from skills are preserved when clearing effects.
  * Prevented kicking the last member of an ANBU squad; squad must be disbanded instead.

* **Refactor**
  * Improved mission and quest logic including daily counter updates and normalized handling of rolled elements.
  * Bounty action buttons refactored to support multiple simultaneous actions including removing all trackers.
  * Updated toggle logic for Kage challenge openness with cooldown and daily locked time enforcement.
  * Combat result calculations refined to incorporate new PvP reward constants and ANBU point awards.
  * Consolidated reroll element logic with separate buttons and refined server-side tracking.
  * Removed unused imports from multiple pages for cleaner code.

* **Style**
  * Removed unused imports from multiple pages for cleaner code.

### DIFF
--- a/app/src/server/api/routers/blackmarket.ts
+++ b/app/src/server/api/routers/blackmarket.ts
@@ -18,10 +18,11 @@ import { MAX_EXTRA_JUTSU_SLOTS } from "@/drizzle/constants";
 import { COST_REROLL_ELEMENT } from "@/drizzle/constants";
 import { RYO_FOR_REP_MAX_LISTINGS } from "@/drizzle/constants";
 import { RYO_FOR_REP_MIN_REPS } from "@/drizzle/constants";
-import { UserRanks, BasicElementName, ElementNames } from "@/drizzle/constants";
+import { BasicElementName, ElementNames } from "@/drizzle/constants";
 import { getRandomElement } from "@/utils/array";
 import { genders } from "@/validators/register";
 import { baseServerResponse, errorResponse } from "../trpc";
+import { canRollPrimaryElement, canRollSecondaryElement } from "@/utils/permissions";
 import { filterValidElementsTypeguard } from "@/libs/train";
 import type { DrizzleClient } from "@/server/db";
 import type { ElementName } from "@/drizzle/constants";
@@ -417,14 +418,16 @@ export const blackMarketRouter = createTRPCRouter({
         fetchUser(ctx.drizzle, ctx.userId),
         getRolledElements(ctx.drizzle, ctx.userId),
       ]);
-      const rankId = UserRanks.findIndex((r) => r === user.rank);
 
       // Guard
       if (user.reputationPoints < COST_REROLL_ELEMENT) {
         return errorResponse("Not enough reputation points");
       }
-      if (rankId < 1) {
-        return errorResponse("You must be at least a Genin to reroll elements");
+      if (input.elementType === "primary" && !canRollPrimaryElement(user)) {
+        return errorResponse("Must be at least a genin to reroll primary elements");
+      }
+      if (input.elementType === "secondary" && !canRollSecondaryElement(user)) {
+        return errorResponse("Must be at least a chunin to reroll secondary elements");
       }
 
       // All the promises to be executed at the end
--- a/app/src/utils/permissions.ts
+++ b/app/src/utils/permissions.ts
@@ -570,5 +570,13 @@ export const canEditCannedResponses = (userRole: UserRole) => {
 };
 
 export const canAwardExperience = (user: UserData) => {
-  return ["CODING-ADMIN","CODER"].includes(user.role);
+  return ["CODING-ADMIN", "CODER"].includes(user.role);
+};
+
+export const canRollPrimaryElement = (user: UserData) => {
+  return !["STUDENT", "NONE"].includes(user.rank);
+};
+
+export const canRollSecondaryElement = (user: UserData) => {
+  return !["STUDENT", "GENIN", "NONE"].includes(user.rank);
 };


app/drizzle/migrations/0257_add_rolled_elements.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE `UserData` ADD `rolledElements` json NOT NULL DEFAULT ('[]');
```
MathiasGruber: This is not a good idea to add to the UserData table, which is one of the most queried tables. A json field is roughly the same as a LONGBLOB / LONGTEXT field in terms of storage. I think we need a separate table to record rolled elements instead

app/drizzle/migrations/0257_add_rolled_elements.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE `UserData` ADD `rolledElements` json NOT NULL DEFAULT ('[]'); 
+ALTER TABLE `UserData` ADD `dailyMedicalMissions` smallint unsigned DEFAULT 0 NOT NULL; 
+ALTER TABLE `UserData` ADD `dailyLockedTimeSeconds` int NOT NULL DEFAULT 0;
```
MathiasGruber: I think here it'd also make sense to instead log all kage toggles into e.g. the `actionLog` table instead, and then use that for counting. This prevents adding a new column to the UserData table, which is used by only a very small percentage of users.

app/src/app/api/hourly-kage-prestige/route.ts
```diff
@@ -27,12 +29,74 @@ export async function GET() {
...
+        );
+
+        await drizzleDB
+          .update(userData)
+          .set({
+            villagePrestige: sql`${userData.villagePrestige} - ${prestigeLoss}`,
+          })
+          .where(eq(userData.userId, kage.userId));
+      }
     }
```
MathiasGruber, line 89: You can test the cron jobs by simply entering the `localhost:3000/api/hourly-kage-prestige` in the browser, that'll execute the endpoint.

app/src/libs/combat/util.ts
```diff
@@ -700,7 +706,18 @@ export const calcBattleResult = (
             deltaPrestige -= isAlly || sameVillage ? FRIENDLY_PRESTIGE_COST : 0;
           }
 
-          // Village tokens for killing enemies
+          // Base prestige for PvP kill
+          deltaPrestige += user.anbuId ? PVP_KILL_PRESTIGE_REWARD_ANBU : PVP_KILL_PRESTIGE_REWARD;
```
MathiasGruber: Should this not be only for killing enemies? Otherwise we reward killing fellow villages with prestige as well?

app/src/libs/combat/util.ts
```diff
@@ -700,7 +706,18 @@ export const calcBattleResult = (
...
+
+          // Base village tokens for PvP kill
+          deltaTokens += user.anbuId ? PVP_KILL_TOKEN_REWARD_ANBU : PVP_KILL_TOKEN_REWARD;
```
MathiasGruber: Should also only be for killing enemies? I'm guessing we can include the ANBU-logic into the statement at line 721 for only applying this to enemies?

app/src/libs/combat/util.ts
```diff
@@ -700,7 +706,18 @@ export const calcBattleResult = (
...
+
+          // ANBU points for PvP kill (only if target is not more than 10 levels under)
+          if (user.anbuId && (user.level - target.level) <= 10) {
```
MathiasGruber: Should also only be for enemies? So we don't award it for killing fellow villages

app/src/server/api/routers/bounty.ts
```diff
@@ -457,6 +457,60 @@ export const bountyRouter = createTRPCRouter({
...
+            columns: {
+              userId: true,
+              username: true,
+            },
+          },
+        },
+      });
+      
+      // Execute deletion and logging in a transaction
+      await ctx.drizzle.transaction(async (tx) => {
```
MathiasGruber: We should not use transactions on planetscale to avoid deadlocks

app/src/server/api/routers/quests.ts
```diff
@@ -275,14 +277,81 @@ export const questsRouter = createTRPCRouter({
...
     .mutation(async ({ ctx, input }) => {
-      // Fetch user
-      const [updatedUser, sectorVillage, results] = await Promise.all([
-        fetchUpdatedUser({
-          client: ctx.drizzle,
-          userId: ctx.userId,
-        }),
-        fetchSectorVillage(ctx.drizzle, input.userSector),
-        ctx.drizzle
+      // Fetch user first to get medical rank
```
MathiasGruber: 1. Do not make queries in a loop - all endpoints need to be as optimized as possible, especially the ones called often, to keep server time & thereby costs down.
2. I think in this endpoint, we can query everything at once disregarding rank, and then sort to get the highest rank, getting the index from from `LetterRanks` in constants file.
3. I think we can simplify this endpoint to apply the logic of getting lower ranked mission if none is available to also non-medical.

MathiasGruber (changes requested):

app/drizzle/migrations/0257_add_rolled_elements.sql
MathiasGruber: Possibly we can use the actionLog table for logging this.

app/src/app/townhall/page.tsx
```diff
@@ -545,6 +562,26 @@ const KageChallenge: React.FC<{
...
+                  {usedHours}h {usedMinutes}m / {KAGE_CHALLENGE_MAX_DAILY_LOCKED_HOURS}h
+                </span>
+              </p>
+              {dailyLockedTimeSeconds >= maxDailySeconds ? (
+                <p className="text-red-500 font-bold">
+                  Daily limit reached! Challenges will be automatically unlocked at the start of the next day.
+                </p>
+              ) : (
+                <p>
+                  <span className="font-bold">Remaining Lock Time: </span>
```
MathiasGruber, line 576: We have a Countdown component which we can use here `app/src/layout/Countdown`.

app/src/app/townhall/page.tsx
```diff
@@ -440,6 +448,15 @@ const KageChallenge: React.FC<{
   );
   const isAtWar = activeVillageWars && activeVillageWars.length > 0;
 
+  // Calculate daily locked time information
+  const dailyLockedTimeSeconds = dailyLockedTimeData?.dailyLockedTimeSeconds ?? 0;
+  const maxDailySeconds = KAGE_CHALLENGE_MAX_DAILY_LOCKED_HOURS * 60 * 60;
+  const remainingSeconds = Math.max(0, maxDailySeconds - dailyLockedTimeSeconds);
+  const remainingHours = Math.floor(remainingSeconds / 3600);
+  const remainingMinutes = Math.floor((remainingSeconds % 3600) / 60);
```
MathiasGruber: We have a convenience function for this `getDaysHoursMinutesSeconds` in `app/src/utils/time.ts`

app/src/server/api/routers/bounty.ts
```diff
@@ -457,6 +457,58 @@ export const bountyRouter = createTRPCRouter({
...
+        fetchUser(ctx.drizzle, ctx.userId),
+      ]);
+      
+      // Guards
+      if (!curBounty) return errorResponse("Bounty not found");
+      if (!canSeeHiddenBountyInfo(user.role)) return errorResponse("Staff access required");
+      if (curBounty.status !== "OPEN") return errorResponse("Bounty not open");
+      
+      // Get all hunters tracking this bounty
+      const hunters = await ctx.drizzle.query.bountySignup.findMany({
```
MathiasGruber: Can be fetched together with curBounty and user?

app/src/server/api/routers/bounty.ts
```diff
@@ -457,6 +457,58 @@ export const bountyRouter = createTRPCRouter({
...
+              userId: true,
+              username: true,
+            },
+          },
+        },
+      });
+      
+      // Execute deletion and logging (no transaction to avoid PlanetScale deadlocks)
+      // Remove all hunter signups for this bounty
+      await ctx.drizzle
```
MathiasGruber: These two mutation can be run in parallel as well

app/src/server/api/routers/blackmarket.ts
```diff
@@ -527,3 +622,82 @@ export const fetchActiveUserOffers = async (client: DrizzleClient, userId: strin
     where: and(eq(ryoTrade.creatorUserId, userId), isNull(ryoTrade.purchaserUserId)),
   });
 };
+
+// Helper function to get rolled elements from actionLog
+async function getRolledElements(client: DrizzleClient, userId: string, elementType?: "primary" | "secondary"): Promise<{ primary: string[], secondary: string[] }> {
```
MathiasGruber: We try never to use explicit return types for maintainability. For consistency, use arrow-functions, i.e. `const getRolledElements = ...`

app/src/server/api/routers/blackmarket.ts
```diff
@@ -527,3 +622,82 @@ export const fetchActiveUserOffers = async (client: DrizzleClient, userId: strin
...
+  // Helper function to get rolled elements for a specific type
+  async function getRolledElementsForType(type: "primary" | "secondary"): Promise<string[]> {
```
MathiasGruber: This entire method effectively runs up to 4 queries if both primary and secondary element are specified. I believe even in that case we can run only a single query, and then split out the informations subsequently.

app/src/server/api/routers/blackmarket.ts
```diff
@@ -527,3 +622,82 @@ export const fetchActiveUserOffers = async (client: DrizzleClient, userId: strin
...
+    return { primary, secondary: [] };
+  }
+
+  if (elementType === "secondary") {
+    const secondary = await getRolledElementsForType("secondary");
+    return { primary: [], secondary };
+  }
+
+  // If no elementType specified, pull both (for backward compatibility)
+  const [primary, secondary] = await Promise.all([
```
MathiasGruber: Backwards compatibility with what? isn't this a new function?

app/src/server/api/routers/blackmarket.ts
```diff
@@ -405,49 +405,144 @@ export const blackMarketRouter = createTRPCRouter({
...
+      
       // Get the updated elements
       const rankId = UserRanks.findIndex((r) => r === user.rank);
-      if (rankId >= 1) {
-        const available = BasicElementName.filter((e) => e !== user.primaryElement);
-        user.primaryElement = getRandomElement(available) ?? null;
+      const changes: string[] = [];
+      
+      // Get rolled elements from actionLog
+      const rolledElementsData = await getRolledElements(ctx.drizzle, ctx.userId, input.elementType);
```
MathiasGruber: Can be run in parallel with the user query

app/src/server/api/routers/blackmarket.ts
```diff
@@ -405,49 +405,144 @@ export const blackMarketRouter = createTRPCRouter({
...
+      
+      // Ensure current elements are in actionLog if not already tracked
+      if (user.primaryElement && !rolledElementsData.primary.includes(user.primaryElement)) {
```
MathiasGruber: These two can be run in parallel

app/src/server/api/routers/blackmarket.ts
```diff
@@ -405,49 +405,144 @@ export const blackMarketRouter = createTRPCRouter({
...
+      // Helper function to filter valid rolled elements
+      const filterValidRolledElements = (elements: string[]): typeof ElementNames[number][] => {
+        return elements.filter(
+          (e): e is typeof ElementNames[number] => ElementNames.includes(e as typeof ElementNames[number])
         );
-        user.secondaryElement = getRandomElement(available) ?? null;
+      };
+      
+      // Extract common reroll logic
+      const rerollElementType = async (
```
MathiasGruber: Why does this convenience function need to be defined as a closure, when the others are outside the endpoint?

app/src/server/api/routers/kage.ts
```diff
@@ -537,21 +548,61 @@ export const kageRouter = createTRPCRouter({
...
-      }
 
-      // Update
-      await ctx.drizzle
-        .update(village)
-        .set({
-          openForChallenges: !userVillage.openForChallenges,
-          openForChallengesAt: new Date(),
+      // Check cooldown from last toggle (using actionLog)
+      const lastToggle = await ctx.drizzle
```
MathiasGruber: This can be fetched together with user, requests, userVillage. Cost-wise it's better to run the queries as much in parallel as possible (as long as they're not crazy heavy), since it improved endpoint speed due to fewer round-abouts to DB, and also lowers the CPU time of the lambda function.

app/src/server/api/routers/kage.ts
```diff
@@ -537,21 +548,61 @@ export const kageRouter = createTRPCRouter({
...
+          );
+        }
+      }
+
+      // Check if trying to close challenges and daily limit has been reached
+      if (!userVillage.openForChallenges) {
+        // Currently closed, trying to open - this is always allowed
+      } else {
+        // Currently open, trying to close - check daily limit
+        const dailyLockedTimeSeconds = await calculateDailyLockedTime(ctx.drizzle, ctx.userId);
```
MathiasGruber: This can also be run with the other queries in parallel

app/src/server/api/routers/quests.ts
```diff
@@ -275,14 +277,26 @@ export const questsRouter = createTRPCRouter({
...
+      const updatedUser = await fetchUpdatedUser({
```
MathiasGruber: I believe something like this would be more consistent with the way we've generally structuring endpoints, and ensure fewer roundabouts to the database:

```ts
      const [updatedUser, sectorVillage, results] = await Promise.all([
        fetchUpdatedUser({
          client: ctx.drizzle,
          userId: ctx.userId,
        }),
        fetchSectorVillage(ctx.drizzle, input.userSector),
        input.type === "medical"
          ? ctx.drizzle
              .select({
                ...getTableColumns(quest),
                previousAttempts: questHistory.previousAttempts,
                completed: questHistory.completed,
              })
              .from(quest)
              .leftJoin(
                questHistory,
                and(
                  eq(quest.id, questHistory.questId),
                  eq(questHistory.userId, ctx.userId),
                ),
              )
          : ctx.drizzle
              .select({
                ...getTableColumns(quest),
                previousAttempts: questHistory.previousAttempts,
                completed: questHistory.completed,
              })
              .from(quest)
              .leftJoin(
                questHistory,
                and(
                  eq(quest.id, questHistory.questId),
                  eq(questHistory.userId, ctx.userId),
                ),
              )
              .where(
                and(
                  eq(quest.questType, input.type),
                  eq(quest.questRank, input.rank),
                  lte(quest.requiredLevel, input.userLevel),
                  gte(quest.maxLevel, input.userLevel),
                  or(
                    isNull(quest.startsAt),
                    gte(quest.startsAt, new Date().toISOString()),
                  ),
                  or(isNull(quest.endsAt), lte(quest.endsAt, new Date().toISOString())),
                  or(
                    isNull(quest.requiredVillage),
                    eq(
                      quest.requiredVillage,
                      input.userVillageId ?? VILLAGE_SYNDICATE_ID,
                    ),
                  ),
                ),
              ),
      ]);
```

i.e. all queries (conditionally) run as the first thing in the endpoint

MathiasGruber (changes requested):

Phrosfire:
> Still a few comments here :) .. hope it's ok I comment instead of fixing, I'm still on vacation with the kids

That's perfectly fine.  I started going through some of them and I'll finish up when I get home from work

Phrosfire:
I also just added in a change that allows structures in war to take normal damage when lower level kages get killed by a higher level.  There were a few villages with low level kage and they weren't taking kage punishment when dying due to the level of the other player.

MathiasGruber:
> I also just added in a change that allows structures in war to take normal damage when lower level kages get killed by a higher level. There were a few villages with low level kage and they weren't taking kage punishment when dying due to the level of the other player.

I'll check that out as well.. I refactored a bit on element rolling & medical filtering to streamline and optimize the code.. let me know if I messed up some intended logic.

PR #604, opened 2025-07-27
fix: resolve mission assignment for wrong villages issue

### LINKED ISSUE #602: Users not receiving errands or missions
**Describe the bug**

Users are clicking on errands or D - B rank missions and not receiving one, but the counter goes up.  I looked and it seems to be giving missions and missions for random villages instead of their own.  A rank missions work fine since they are chosen specifically.  

If you get a mission or errand from another village, the counter will go up and you will have no mission to complete.  If its one for your village that you receive, you can complete it as normal.

**To Reproduce**

1. Go to the mission page
2. Click on Errands on missions
3. Get a mission or errand of that rank for potentially the wrong village

**Expected behavior**

1. Go to the mission page
2. Click on Errands on missions
3. Get a mission or errand of that rank for your village.

**Further Information**

The 9 mission limit before only getting 40% rewards for regular missions are also affecting med missions.

**Screenshots**

MathiasGruber:
Fixed two issues in the mission/errand system:

1. Village filtering bug: Users were receiving missions/errands for random villages instead of their own because the system allowed quests with NULL requiredVillage to be assigned to all users. Now restricts quest assignment to only match the user's specific village.

2. Mission limit bug: The reward reduction for missions after the 9th daily mission was using > 9 instead of >= 9, causing reduced rewards to only apply starting from mission 11 instead of mission 10.

Fixes #602

Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted the mission reward multiplier to apply when completing 9 or more daily missions, instead of only when exceeding 9.

* **Refactor**
  * Updated quest selection to require a matching village, no longer including quests without a specified village requirement. Only quests with a village requirement matching the user's village or default syndicate will be available.

### DIFF
--- a/app/src/libs/quest.ts
+++ b/app/src/libs/quest.ts
@@ -196,7 +196,7 @@ export const getReward = (
       userQuest.quest.questType,
     );
     const factor =
-      missionLike && user.dailyMissions > 9 ? ADDITIONAL_MISSION_REWARD_MULTIPLIER : 1;
+      missionLike && user.dailyMissions >= 9 ? ADDITIONAL_MISSION_REWARD_MULTIPLIER : 1;
     rawRewards.reward_money = Math.floor(rawRewards.reward_money * factor);
     rawRewards.reward_clanpoints = Math.floor(rawRewards.reward_clanpoints * factor);
     rawRewards.reward_anbupoints = Math.floor(rawRewards.reward_anbupoints * factor);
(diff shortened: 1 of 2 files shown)


MathiasGruber:
This doesn't make any sense to me how this would resolve the issue in question.. if the requiredVillage is not set, it should be fetched, not ignored.

PR #632, opened 2025-08-09
Fix issue #558: Item Filter under Info > Items is inconsistent (Low Priority)

### LINKED ISSUE #558: Item Filter under Info > Items is inconsistent (Low Priority)
Items sometimes show duplicates when using the filter to see only event items

Sometimes the list populates with things from other categories that should be filtered out (toggling event and shop together) i.e. non event items appearing under the event category

List getting stuck if different filters are selected then untoggled etc

It basically creates an inconsistent experience, I was going through Items trying to keep track of what assets are being used and which aren't so that the database could be cleared up when I noticed this.

**To Reproduce**

1. Go to Info > Items
2. Play with the filter, turn off Shop, turn on Event items, turn off Event items, turn on Shop, then turn on Event items then turn off Shop
3. Refresh, turn on 'Event' filter, turn off 'Shop' filter, scroll down and it will populate with duplicates

**Expected behavior**
Filters to respect categories, no duplicates. Other sections under info like Quest and AI did not at all have this issue.

Image

TheNinja-RPG:
This pull request fixes #558.

- Removed placeholderData from useInfiniteQuery on the Items page so a filter change no longer reuses the previous query’s pages. This prevents old (Shop) results from sticking around when switching to Event and eliminates mixed categories and duplicate entries caused by merging placeholder pages with newly fetched pages.
- Changed onlyInShop default from true to undefined in the filtering hook, aligning with a tri-state “All” default. This avoids unintentionally combining “Shop-only” with “Event” when users toggle filters, reducing conflicting filters and the “stuck” list behavior.
- Added a deterministic tie-breaker in the DB ordering (cost, repsCost, id). This stabilizes offset-based pagination, preventing duplicates/missing rows when scrolling under consistent filters.
- Added tests to ensure onlyInShop and eventItems conditions are only applied when explicitly set, guarding against “sticky” filters that would leak items from other categories.

Together, these changes directly address the reported symptoms: duplicates during infinite scroll, non-event items appearing under the Event filter due to stale data, and inconsistent behavior when toggling filters.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the loading behavior when fetching items, resulting in a more accurate display during data transitions.

* **Refactor**
  * Enhanced item sorting to ensure consistent ordering when items have the same cost and reps cost. Items are now further sorted by their ID.

### DIFF
--- a/app/src/app/manual/item/page.tsx
+++ b/app/src/app/manual/item/page.tsx
@@ -40,7 +40,6 @@ export default function ManualItems() {
     },
     {
       getNextPageParam: (lastPage) => lastPage.nextCursor,
-      placeholderData: (previousData) => previousData,
     },
   );
   const allItems = items?.pages.map((page) => page.data).flat();
--- a/app/src/server/api/routers/item.ts
+++ b/app/src/server/api/routers/item.ts
@@ -283,7 +283,11 @@ export const itemRouter = createTRPCRouter({
         offset: skip,
         limit: input.limit,
         where: and(...baseFilters),
-        orderBy: (table, { asc }) => [asc(table.cost), asc(table.repsCost)],
+        orderBy: (table, { asc }) => [
+          asc(table.cost),
+          asc(table.repsCost),
+          asc(table.id),
+        ],
       });
       const nextCursor = results.length < input.limit ? null : currentCursor + 1;
       return {


theeneon:
LGTM

MathiasGruber (approved):
Did a little big of refactoring of AI solution, but otherwise seems to work from my testing as well

PR #634, opened 2025-08-09
Fix issue #633: Unable to Update Quest Data

### LINKED ISSUE #633: Unable to Update Quest Data
**Describe the bug**
Trying to change the background and the reward rank on the Road to Genin starter quest and it's not updating. 
When making edits or changes to a quest that has over 10+ objectives, the objectives  or quest options no longer update. 

When reward_rank is changed from None to Genin it states: 

Image 

And it goes back to none. 

This is also happening on the objectives as well, I can't update the backgrounds or characters. It seems that any field that contains a drop down or a selection isn't being saved.

Also Unable to search for AIs or Items, the field doesn't populate and you have to manually scrolldown to find the AI or item that you're looking for. 

**To Reproduce**

1. Go to Road To Genin Starter Quest
2. Try to Change Rank from none to Genin, 
3. Refresh it's still set at none

**Expected behavior*
Field should be saved

TheNinja-RPG:
This pull request fixes #633.

- Persisting changes: The form now explicitly marks select fields as dirty when their value changes. For multi-selects, the onChange handler was wrapped to call field.onChange and form.setValue(id, vals, { shouldDirty: true }); For single-selects, selecting an option now uses form.setValue with { shouldDirty: true }. If the save logic relies on dirtyFields, this directly addresses “dropdown/selection fields not being saved,” including for quests with many objectives.
- Search usability: In the MultiSelect component, each CommandItem now has value={option.label}, which is what cmdk uses to filter options when typing. This resolves the “unable to search for AIs or Items” problem where the list wouldn’t filter/populate while typing.
- Stability with many fields: Adding a stable key to the spacer div helps React reconcile lists more predictably in large dynamic forms, reducing the risk of mismatched state when many objectives are present.
- Minor UX: “No framework found.” was corrected to a generic “No options found.”

Together, these changes ensure dropdown/multi-select changes are captured and saved, and that searching within multi-selects works as expected.


theeneon:
Solution failed. Still unable to search for AI or Items to add them when editing a quest or the AI.

MathiasGruber:
Should be fixed with my updates here.

PR #635, opened 2025-08-09
Fix issue #628: War: Kage Damage Bug

### LINKED ISSUE #628: War: Kage Damage Bug
**Describe the bug**

During war, a kage can attack and kill any low-level villager of the warring village and it'll deal full war damage.  So if a level 100 kage kills a level 30 chunin, its applying the full kage kill damage.  If a low level attacks and kills a high level kage, only 1 damage is dealt.

**To Reproduce**

1) Start a war
2) Attack a level 30 as a kage while being level 100
3) Full kage war damage applies

**Expected behavior**

A high level kage kills a low-level player, only 1 damage should be applied.

**Further Information**

The only exception to higher levels not dealing full damage to a lower level, is when the lower level is kage.  Kage should always take full damage regardless of the level difference.

TheNinja-RPG:
This pull request fixes #628.

The fix modifies calcBattleResult to cap war damage (both shrine and townhall) to +/-1 only when there is a significant level gap between the attacker and the defeated targets (abs(level diff) > STREAK_LEVEL_DIFF) and no kage was killed. It introduces a wasKageKilled check and applies capping via a capToOne flag; if any defeated target is the kage, capToOne is false and full war damage is preserved. This directly addresses:
- High-level kage killing a low-level non-kage: damage is now capped to 1.
- Low-level killing a high-level kage: no capping, kage takes full damage.

Additionally, the previous logic always capped shrine damage even when a kage was killed; the new code corrects that by applying the same exception (no cap) to both shrine and townhall changes when a kage is killed.

New tests were added to validate both scenarios: one ensures damage is capped to 1 for a high-level attacker vs. low-level non-kage, and another ensures no cap when the defeated target is a kage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved calculation of maximum target level for level difference adjustments, now considering only non-AI target users.
  * Revised shrine and townhall damage adjustments to more accurately reflect battle outcomes, especially regarding kage losses and level differences.
  * Item drop rolls now correctly consider only AI targets.

* **Chores**
  * Removed a debug console log related to item drop checking.

### DIFF
--- a/app/src/libs/combat/util.ts
+++ b/app/src/libs/combat/util.ts
@@ -919,31 +919,32 @@ export const calcBattleResult = (
       });
 
       // Adjust shrine & townhall datamage based on level different
-      const maxTargetLevel = Math.max(...targets.map((t) => t.level), 0);
+      const maxTargetLevel = Math.max(...targetUsers.map((t) => t.level), 0);
       if (Math.abs(user.level - maxTargetLevel) > STREAK_LEVEL_DIFF) {
         // Check if any kage was killed in this battle
-        const wasKageKilled = targets.some(
-          (target) => target.village?.kageId === target.userId,
-        );
-
-        if (shrineChangeHp !== 0) shrineChangeHp /= Math.abs(shrineChangeHp);
-        // Only reduce townhallChangeHP if no kage was killed
-        if (townhallChangeHP !== 0 && !wasKageKilled) {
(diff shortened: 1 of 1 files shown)


MathiasGruber (approved):
Had to scrap AI solution, but tested and fixed logic myself

Issue #658, opened 2025-08-13
Updates to War System


theeneon:
So after discussions with content and players we want to add some quality of life updates to the War System to make it more interactive and give more incentives to engage in wars:

- [x] When a village who's at war loses a sector they lose 300 HP on their town hall. 
- [x] Recapturing an enemy sector restores 100 points on your townhall. 

Better Rewards:

- [x] The following structures will get +3 Levels temporarily for a week
Training Ground, Ramen Shop, Mission Hall, Home,Battle Arena

- [x] 500000 Token

Better Punishment:

- [x] Losing villages and Factions loses 3 levels on all structures 

Decay:
- [x] Increase daily decay to 3% of total Tokens  both Raids and Wars. So both Outlaws and Factions should have decay while raiding and at War.  If War reaches 4 days increase decay to 6% if War Reaches 10 days increase decay to 10% 

Now am not sure this is possible but I hope it is:

War Missions and Objectives

- [x] Allow Factions to form an alliance with Villages. 
- [x] Allow Factions to Request Aid from Villages
- [x] Allow Factions to create alliance with other Factions
- [x] If a village is at war, block a new war from being created. Instead prompt the faction or village to request alliance with the village that started the war. Another option here is to create a Village Health instead of using townhall as the main metric, so whenever a new ar is started a new village health is created for each instance of war. (Maybe we can discuss this) the problem we're having here is that factinos and villages are teaming up on one village where it's 3v1 and the defender basically has to deplete all the attackers' townhalls while one kill on the defender replinishes all the attackers townhall. 
- [x] War Objectives These are random objectives that pop up every couple hours while at war that will do bonus damage and bonus healing to townhalls

TheNinja-RPG:
[OpenHands](https://github.com/MathiasGruber/OpenHands) started fixing the issue! You can monitor the progress [here](https://github.com/studie-tech/TheNinjaRPG/actions/runs/16964405882).

theeneon:
@MathiasGruber saw that you were working on the War system, let me add these here:

- [x] Village at War loses a sector; they lose 300HP. Capturing a sector that you lost gives 150HP
- [x] Losing a war reduces all structures by 3 levels for 7 days after the one week, Kages can start upgrading the village structures again. 
- [x] Losing Village can’t have war declared on them for 10 Days. 
- [x] The winning participants can’t declare war for 3 days. 
[...]

PR #687, opened 2025-08-21
Fix issue #658: Updates to War System

MathiasGruber:
# Pull Request

Work in Progress

## License


theeneon:
@MathiasGruber can you check why Vercel isn't deploying anymore?

theeneon:
- [ ] Allow Factions to form an alliance with Villages. 
- [ ] Allow Factions to Request Aid from Villages
- [ ] Allow Factions to create alliance with other Factions
- [ ] If a village is at war, block a new war from being created. Instead prompt the faction or village to request alliance with the village that started the war. Another option here is to create a Village Health instead of using townhall as the main metric, so whenever a new ar is started a new village health is created for each instance of war. (Maybe we can discuss this) the problem we're having here is that factinos and villages are teaming up on one village where it's 3v1 and the defender basically has to deplete all the attackers' townhalls while one kill on the defender replinishes all the attackers townhall. 
- [ ] War Objectives These are random objectives that pop up every couple hours while at war that will do bonus damage and bonus healing to townhalls

MathiasGruber:
Abandoning solution, starting from scratch

PR #690, opened 2025-08-23
Time Dilation/Compression & Redirection

Phrosfire:
# Pull Request

**New Tags**

- Time Compression - Increases AP usage by 10.  If an elements are assigned, it only effects those elements as well as basic actions.
- Time Dilation - Decreases AP usage by 10.  If an elements are assigned, it only effects those elements as well as basic actions.
- Redirection - Has a direction setting for push/pull.  Power dictates that max distance moved. Target will not overlap the user or get pushed beyond the bounds of the arena.
- Allow up to 30 skill points instead of 20.  20 gained from levels and 10 gained from quests.
- Allow skill point rewards on quests
- Allow requiring a specific bloodline to take a quest

The tags have been tested and are working.  Both tags are doing a lot, so if you end up seeing something you think could be cleaner, please let me know and I can take care of it.

## License

## Summary by CodeRabbit

- New Features
  - Added new combat effects: Time Compression, Time Dilation, and Redirection, impacting AP costs and allowing push/pull movement; shown in Statuses with labels/icons.
  - Quest rewards now support Skill Points; reward toasts and displays include them.
  - Added Bloodline-based quest gating and a Bloodline filter in quest search; quests can require a specific bloodline.
- Changes
  - AP calculations now account for time-based effects per action/element.
  - Level-up Skill Points capped at 20 from leveling; overall cap raised to 30.
- Chores
  - Database updated to support bloodline requirements for quests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -173,5 +173,5 @@ git-branch-uncommit: # Undo the last N commits (keeping changes staged), usage: 
 	@echo "${YELLOW}Uncommitting branch commits${RESET}"
 	git diff --binary --ignore-space-at-eol origin/main...HEAD > branch.patch
 	git reset --hard origin/main
-	git apply --whitespace=nowarn branch.patch
+	git apply --reject --whitespace=fix branch.patch
 	rm -f branch.patch
--- a/app/drizzle/constants.ts
+++ b/app/drizzle/constants.ts
@@ -1648,7 +1648,7 @@ export const BOUNTY_MAX_HUNTERS = 3;
 export const BOUNTY_MIN_AMOUNT = 1000000;
 
 // Skill system constants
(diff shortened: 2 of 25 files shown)


MathiasGruber:
@theeneon we probably need to do some thorough testing of these tags?

theeneon:
> @theeneon we probably need to do some thorough testing of these tags?

Phros was able to test them extensively they don't seem to have any issues but we will be testing them in game before ever releasing them into the wild. The first applications here would likely be on AI fights before exposing this to users.

Phrosfire:
> @theeneon we probably need to do some thorough testing of these tags?

I tested these against AI (self for time compression) and they were working as intended. Testing in various pvp scenarios though would be good though before we make it available to players to actually use.

PR #700, opened 2025-09-02
Bug Fixes/Afterburn Change/New Event Multipliers

Phrosfire:
# Pull Request

**Bugs**

- Ground effects no longer last one action longer than it should.
- Fixed Kage battles by adding a new battletype and having it check statuses to start the battle
- Display the map during travel
- When checking inventory counts, materials are no longer counted with the normal inventory

**Tag Change**

- Don't have pierce damage trigger afterburn

**Event Modifiers**

- Added additional modifiers that can be enabled from the event page for Jutsu Experience, Battle Experience, and Missions.

**Note**
I was asked by Neon to add the permission for the content staff to use the Event page to enabled exp events.  I let him know I'd add it and the decision would be yours on if its allowed or not during the PR review.  It sounds like they want rotating boosts for players.

## License

## Summary by CodeRabbit

- New Features
  - Added Kage PvP queue status with automatic redirect to Town Hall and updated challenge flow.
  - Introduced global EXP boosts for battle, mission, and jutsu; applied to combat, quests, and jutsu gains.
  - Event page now lets authorized users schedule/set these multipliers.
  - Profile shows active boost notifications with links.

- Bug Fixes
  - Kage challenge expiry now reliably returns users to AWAKE.
  - More accurate damage resolution for afterburn vs. pierce effects.
  - Ground effects/barriers expire immediately.
  - No health/stamina/chakra loss in sparring modes.

- UI
  - Wider popover for grouped notifications.

- Chores
  - CONTENT role can modify event gains.


MathiasGruber (approved):
Refactored a bit here. Primarily:

- Use `make makemigrations` to create proper migrations with drizzle journal etc
- Moved settings into extraState in battle - battle actions need to be highly optimized, so we wish to have as few round-abouts to the DB during battle actions.

PR #702, opened 2025-09-04
Fix issue #701: Reduce Travel Time and Map related questions

### LINKED ISSUE #701: Reduce Travel Time and Map related questions
One of the pain points users have expressed is that travel time on the map is just a bit too long, is it possible to cap Travel time to 10 seconds at max? 

Players have also asked for instant travel to certain locations, specifically Wake Island.

TheNinja-RPG:
This pull request fixes #701.

- Added a new constant WAKE_ISLAND_SECTOR = 222.
- Updated calcGlobalTravelTime to:
  - Return 0 seconds when the destination sector equals WAKE_ISLAND_SECTOR (instant travel to Wake Island).
  - Cap computed global travel times to a maximum of 10 seconds via Math.min(secs, 10).
  - Reduce the fallback time from 300 seconds to 10 seconds to honor the cap even when map data is missing.
- Added tests that verify both the 10-second cap and the instant travel to Wake Island.

These changes directly address the requests: global map travel times will not exceed 10 seconds, and traveling to Wake Island is instantaneous.

### DIFF
--- a/app/src/libs/travel/constants.ts
+++ b/app/src/libs/travel/constants.ts
@@ -8,6 +8,8 @@ export const ALLIANCEHALL_LAT = 7;
 export const HOSPITAL_LONG = 12;
 export const HOSPITAL_LAT = 8;
 
+export const WAKE_ISLAND_SECTOR = 222;
+
 export const groundAssets = [
   {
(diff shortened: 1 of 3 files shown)


MathiasGruber:
Implementing in another commit instead.

Issue #714, opened 2025-09-07
Uncap Anbu MIssions & Pick Medical Missions


Phrosfire:
At the moment, anbu quests can be completed a set number of times based on what's been set on the quest.  We would instead like anbu quests to just have an unlimited number of completions.

For medical quests, we would like users to be able to select which medical quest based on what they qualify for.  It should behave like A rank mission select.

TheNinja-RPG:
[OpenHands](https://github.com/MathiasGruber/OpenHands) started fixing the issue! You can monitor the progress [here](https://github.com/studie-tech/TheNinjaRPG/actions/runs/17535248664).

Phrosfire:
Fix failed

TheNinja-RPG:
[OpenHands](https://github.com/MathiasGruber/OpenHands) started fixing the issue! You can monitor the progress [here](https://github.com/studie-tech/TheNinjaRPG/actions/runs/17551754417).

PR #715, opened 2025-09-08
Fix issue #714: Uncap Anbu MIssions & Pick Medical Missions

TheNinja-RPG:
**QoL Updates**
- Anbu missions uncapped
- medical missions now let you select a mission you qualify for (Same as A rank missions)
- Can set items to require a specific bloodline to be equipped.

**Bug Fixes**
- Med missions no longer greyed out when there are missions available of a lower rank
- Items with a silver cost no longer available to be selected in ranked

**Other Adjustments**
- If you have an item on for a bloodline you don't have active, the item will not process for battle.

## Summary by CodeRabbit

- New Features
  - Items can require a specific bloodline; this is shown on item pages and selectable in item editing.
  - Mission selection for Medical and A‑rank uses a new picker with clear dialogs and actions.

- Improvements
  - Medical missions now aggregate across ranks with “showing up to X rank” messaging.
  - Daily mission limits are enforced and communicated in dialogs; A‑rank flows show reduced‑rewards notices when applicable.
  - Ranked loadouts now only list zero‑cost weapons/consumables.
  - Bloodline requirements are enforced when buying, equipping, loading out, and in combat effects.

### DIFF
--- a/app/drizzle/schema.ts
+++ b/app/drizzle/schema.ts
@@ -1161,6 +1161,7 @@ export const item = mysqlTable(
     canBeHunted: boolean("canBeHunted").default(false).notNull(),
     canBeGathered: boolean("canBeGathered").default(false).notNull(),
     canBeTraded: boolean("canBeTraded").default(false).notNull(),
+    bloodlineId: varchar("bloodlineId", { length: 191 }),
   },
   (table) => {
     return {
@@ -1175,6 +1176,7 @@ export const item = mysqlTable(
       costIdx: index("Item_cost_idx").on(table.cost),
       repsCostIdx: index("Item_repsCost_idx").on(table.repsCost),
       requiredLevelIdx: index("Item_requiredLevel_idx").on(table.requiredLevel),
+      bloodlineIdIdx: index("Item_bloodlineId_idx").on(table.bloodlineId),
     };
   },
 );
@@ -1183,9 +1185,13 @@ export type ItemType = Item["itemType"];
 export type ItemSlotType = Item["slot"];
 export type ItemRarity = Item["rarity"];
 
-export const itemRelations = relations(item, ({ many }) => ({
+export const itemRelations = relations(item, ({ one, many }) => ({
   craftingRequirements: many(craftingRequirement, {
     relationName: "craftItems",
+  }),
+  requiredBloodline: one(bloodline, {
+    fields: [item.bloodlineId],
+    references: [bloodline.id],
   }),
 }));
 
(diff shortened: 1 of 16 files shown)


app/drizzle/migrations/0290_add_item_bloodline_requirement.sql
MathiasGruber, line 1: Please use the `make makemigrations` make-command, so that the journal & meta files are updated properly as well :)

app/drizzle/schema.ts
```diff
@@ -1100,6 +1101,7 @@ export const item = mysqlTable(
       costIdx: index("Item_cost_idx").on(table.cost),
       repsCostIdx: index("Item_repsCost_idx").on(table.repsCost),
       requiredLevelIdx: index("Item_requiredLevel_idx").on(table.requiredLevel),
+      requiredBloodlineIdx: index("Item_requiredBloodline_idx").on(table.requiredBloodlineId),
```
MathiasGruber: To align with jutsu-table, I think we should just call it `bloodlineId`

app/src/layout/MissionHall.tsx
```diff
@@ -135,20 +135,107 @@ export default function MissionHall({ userData }: MissionHallProps) {
...
-                    q.questRank === setting.rank &&
-                    (q.medicalRank === "NONE" || q.medicalRank === userMedicalRank),
-                ) ?? false)
+              ? filtered.length > 0  // Show colored if ANY medical missions are available
               : true;
             const grayScale = count === 0 || capped || !rankCheck || !medicalCheck;
 
+
+            if (isMedical) {
+              return (
```
MathiasGruber, line 156: Since we have this same general logic for A-rank, I think we should generalize it out, so that we can re-use it also for other missions more easily in the future if needed, and maintain easier

app/tests/libs/quests-anbu.test.ts
```diff
@@ -0,0 +1,76 @@
+import { describe, it, expect } from "vitest";
+import { isAvailableUserQuests } from "@/libs/quest";
+import { QuestTypesWithMaxAttempts } from "@/drizzle/constants";
+
+describe("ANBU quests unlimited completions", () => {
```
MathiasGruber: No need for this test I think

MathiasGruber (changes requested):
Looks good, but a few polishes needed :)

Phrosfire:
@MathiasGruber I'm downstate for work at the moment. I'll fix this up tomorrow when I'm back home

Phrosfire (approved):
Good to go

PR #737, opened 2025-10-02
Fix immunity & Robbing

Phrosfire:
# Pull Request

I forgot to test wake island with the immunity changes and players were able to be attacked and robbed in wake island.  This update fixes that.

Also fixed the bug that prevented syndicate not in a faction from robbing other non-faction syndicate members.  The same bug prevents non-faction syndi members from robbing non-clan village members.  It was doing a clanId comparison, but was triggered by null = null.

## License


app/drizzle/constants.ts
```diff
@@ -1,6 +1,9 @@
 export const OCCUPATIONS = ["GATHERING", "HUNTER", "CRAFTING"] as const;
 export type OccupationType = (typeof OCCUPATIONS)[number];
 
+export const CURRENCY_TYPES = ["MONEY", "REPUTATION"] as const;
```
MathiasGruber: Why not seichi silver as well?

app/drizzle/constants.ts
```diff
@@ -992,6 +995,7 @@ export const JUTSU_XP_TO_LEVEL = 1000;
 export const JUTSU_LEVEL_CAP = 20;
 export const JUTSU_TRAIN_LEVEL_CAP = 25;
 export const MAX_DAILY_TRAININGS = 64;
+export const MAX_JUTSU_TRAIN_TIME_MS = 60 * 60 * 1000; // 1 hour in milliseconds
```
MathiasGruber, line 1009: Doesn't this make it too easy / fast to max jutsus?

app/src/server/api/routers/combat.ts
```diff
@@ -1312,6 +1356,12 @@ export const initiateBattle = async (
 
   // Check if the villageData is in a pvp enabled zone
   const sectorData = villages.find((v) => v.sector === sector);
+  
+  // Special check for Wake Island - always block if sector is 222
+  if (sector === 222 && battleType === "COMBAT") {
```
MathiasGruber: Use `MAP_WAKE_ISLAND_SECTOR` from constants

app/src/server/api/routers/combat.ts
```diff
@@ -1013,20 +1013,64 @@ export const combatRouter = createTRPCRouter({
...
-      // Optimized: fetch only userId, username, avatar from usersState JSON
-      const results = await ctx.drizzle.execute(
-        sql`
-          SELECT 
-            id, battleType, createdAt, round, updatedAt,
-            JSON_EXTRACT(usersState, '$[*].userId') as userIds,
-            JSON_EXTRACT(usersState, '$[*].username') as usernames,
-            JSON_EXTRACT(usersState, '$[*].avatar') as avatars
-          FROM Battle
-          WHERE battleType = ${input.battleType}
```
MathiasGruber, line 1033: I don't like having a double-query for performance reasons. I believe we can simply update the old query with a new `(battleType NOT IN ('SPARRING', 'RANKED_SPARRING') OR spectable = true` condition, or something similar.

app/drizzle/constants.ts
Phrosfire: We don't intend for silver to be a tradeable currency, it should stay tied specifically to the user.

app/drizzle/constants.ts
Phrosfire: This will cap out the training time at 1 hour which is certainly faster.  The cost of training a jutsu is remailing the same though.  I was asked by Neon to implement the cap of an hour and others in content were on-board so long as the cost remained the same

app/drizzle/constants.ts
Phrosfire: I just added Silver to currency types and made a second one for tradeable currencies.  I used the tradeable currencies one for the auction house.

Phrosfire:
@MathiasGruber this PR seems to be stuck on the coderrabbit request, but it was fixed.

PR #760, opened 2025-11-11
Set stack quantity in auction.  Display stack quantitity

Phrosfire:
# Pull Request

When creating an auction, users can now pick the quantity they want to trade if its a stacked item.  The quantity is now displayed in the auction house to users as well.  The ability to split the stack quantity was added to the item view as well

## License

## Summary by CodeRabbit

* **New Features**
  * Split stackable items via a new Split Stack dialog and two-step workflow.
  * Choose a specific quantity when creating an auction; UI shows quantity input, validation, current-stack hints, and quantity in item displays.
  * Bid/Buyout confirmation flows with pending state and post-confirmation clearing.

* **Server**
  * Auctions may use a partial stack: server accepts optional quantity, validates it, performs stack split when needed, and records quantity in logs.

### DIFF
--- a/app/src/app/auctionhouse/page.tsx
+++ b/app/src/app/auctionhouse/page.tsx
@@ -332,6 +332,11 @@ const AuctionDetailsDialog: React.FC<AuctionDetailsDialogProps> = ({
             bank: userData.bank - data.amountToDeduct,
           });
         }
+        // Clear state after successful bid
+        setBidAmount("");
+        setPendingBidAmount(null);
+        setShowBidConfirmation(false);
+        setShowBuyoutConfirmation(false);
       }
     },
   });
@@ -380,17 +385,12 @@ const AuctionDetailsDialog: React.FC<AuctionDetailsDialogProps> = ({
   const confirmBid = () => {
     if (pendingBidAmount !== null) {
       handlePlaceBid(auctionId, pendingBidAmount);
-      setBidAmount("");
-      setPendingBidAmount(null);
-      setShowBidConfirmation(false);
     }
   };
 
   const confirmBuyout = () => {
     if (pendingBidAmount !== null && listing?.buyoutPrice) {
       handlePlaceBid(auctionId, listing.buyoutPrice);
-      setPendingBidAmount(null);
-      setShowBuyoutConfirmation(false);
     }
   };
 


app/src/server/api/routers/item.ts
```diff
@@ -390,6 +390,78 @@ export const itemRouter = createTRPCRouter({
...
+      }
+
+      if (input.quantityToKeep < 1) {
+        return { success: false, message: "Quantity to keep must be at least 1" };
+      }
+
+      const quantityToSplit = currentUserItem.quantity - input.quantityToKeep;
+
+      // Update current stack to keep the specified quantity
+      await ctx.drizzle
```
MathiasGruber: I believe these two queries can be run in parallel with Promise.all. Also, I think we should abstract the splitting functionality out to a convenience method in the @item.ts file, and then re-use it in the auction.ts file as well; that'll make it easier to maintain, and given it touches upon splitting, there could be some exploit related to the code at some point - I think think so, it looks solid, but if there is, then it's nice to only have to review and fix the exploit in one place.

MathiasGruber (changes requested):

app/src/server/api/routers/item.ts
Phrosfire: I'll make the change when I get back home. Had to travel downstate for work and I won't return until tomorrow or Thursday.   Moving to draft for now

app/src/server/api/routers/item.ts
```diff
@@ -1388,3 +1456,73 @@ export const itemDatabaseFilter = (
...
 };
+
+/**
+ * Split an item stack into two stacks
+ * @param client - The database client
+ * @param userItemId - The ID of the user item to split
+ * @param quantityToKeep - The quantity to keep in the original stack
+ * @returns The ID of the newly created stack, or null if splitting failed
+ */
+export const splitItemStack = async (
```
MathiasGruber, line 1420: Now we have the logic duplicated even more; I believe this function should return a fixed return type including potential failure message, so that we can return that in case if fails any of the checks. Right now it's running the checks both in the endpoint and in this function, i.e. a lot of code duplication, so if we want to change the logic at some point, we need to change the checks 3 places

Phrosfire: I believe I've got it cleaned up now.  I also added a confirmation for placing bids or buying out an item from the auction house.

MathiasGruber (approved):

PR #762, opened 2025-11-13
No longer apply durability loss in spars

Phrosfire:
# Pull Request

- Prevents durability from being lost when sparring.  Ranked spars don't need this since it uses their ranked loadouts and doesn't effect their gear.
- Allows users to modify their ranked loadout when there isn't an active season.  This allows users to edit their loadout for ranked spars in the off-season.

## License

## Summary by CodeRabbit

* **New Features**
  * Weapon and armor durability are no longer consumed during sparring battles.

* **Improvements**
  * Ranked loadout selector now renders without requiring an active season.

### DIFF
--- a/app/src/libs/combat/process.ts
+++ b/app/src/libs/combat/process.ts
@@ -46,6 +46,7 @@ import {
   BATTLE_TAG_STACKING,
   ID_ANIMATION_SMOKE,
   DURABILITY_USABILITY_THR,
+  NO_DURABILITY_LOSS_COMBATS,
 } from "@/drizzle/constants";
 import type { BattleUserState, ReturnedUserState } from "./types";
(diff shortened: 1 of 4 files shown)


app/src/libs/combat/process.ts
```diff
@@ -413,17 +413,19 @@ export const applyEffects = (
             color: "red",
             types: c.types,
           });
-          // Reduce armor durability by 1 when hit
-          const t = newUsersState.find((u) => u.userId === target.userId);
```
MathiasGruber, line 417: Better introduce a `NO_DURABILITY_LOSS_COMBATS = ['SPARRING']` in @constants file, so that if we need to add other battle types, we automatically get it updated all the right places

MathiasGruber (changes requested):

PR #777, opened 2025-11-26
New bug fixes

Phrosfire:
# Pull Request

- Don't allow users to sleep in wartorn sector
- When users lose money from wartorn, don't let it put them below 0 money
- Don't apply scaled levels from combat to user_level tasks
- Don't apply scaled levels from combat to minimum/maximum level needed for quests (Was causing quests to reset if going above max level)
- Properly assign dailies based on user level

## License

## Summary by CodeRabbit

* **Bug Fixes**
  * Money loss in war-torn/mixed combat no longer drops balances below zero.
  * Daily quests now respect players' preserved pre-combat level for availability and honor daily completion/reset rules.

* **New Features**
  * Sleeping is blocked while in dangerous battlegrounds.
  * Staff link on Users page now opens the in-app staff manual.
  * Players' original pre-combat level is preserved for quest and validation checks.

### DIFF
--- a/app/src/libs/quest.ts
+++ b/app/src/libs/quest.ts
@@ -617,7 +617,12 @@ export const getNewTrackers = (
 
           // General updates we want to apply every time
           if (task === "user_level") {
-            status.value = user.level;
+            // Use originalLevel if available (for combat-scaled users), otherwise use current level
+            const userLevel =
+              "originalLevel" in user && typeof user.originalLevel === "number"
+                ? user.originalLevel
+                : user.level ?? 1;
+            status.value = userLevel;
           } else if (task === "days_in_village") {
             const days = Math.floor(secondsPassed(user.joinedVillageAt) / 60 / 60 / 24);
             status.value = days;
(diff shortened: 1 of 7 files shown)


app/src/libs/quest.ts
```diff
@@ -1146,6 +1156,12 @@ export const isAvailableUserQuests = (
       );
     });
 
+  // Daily quests should be available even if previously completed (they reset daily)
+  const completedCheck =
+    questAndUserQuestInfo.questType === "daily" ||
+    !questAndUserQuestInfo.completed ||
+    questAndUserQuestInfo.completed === 0;
```
MathiasGruber, line 1164: This conflicts with `eventCompletedCheck`, no? i.e. some quests we allow multiple completes of, but this would prevent any non-daily multiple completes

app/src/libs/quest.ts
```diff
@@ -1172,14 +1189,18 @@ export const isAvailableUserQuests = (
     message += `Quest requires medical rank ${capitalizeFirstLetter(questMedRank ?? "NONE")}\n`;
   if (!huntingRankCheck)
     message += `Quest requires hunting rank ${capitalizeFirstLetter(questHuntRank ?? "NONE")}\n`;
+  if (!completedCheck) message += "Quest has already been completed\n";
```
MathiasGruber: Again, need to incorporate with the logic for allowing multiple completes

MathiasGruber (changes requested):

Phrosfire:
I made adjustments to account for the quests that allow multiple completions.  I adjusted one of the events to have multiple allowed completions and I was able to properly complete it the set number of times.

While making the adjustment, I noticed that I ran into an issue with not getting dailies again on dev.  I set up some logging and found that it wasn't properly limiting you to dailies within your level and it was causing missions to not be assigned if you were outside the required levels.  I've tested it and confirmed that it now only assigns missions within your level range.

MathiasGruber:
Great work, thanks! :D

PR #781, opened 2025-11-26
 ✨staff: vote status, admin notification, delete action

### LINKED ISSUE #780: Improve staff application feature
Application system could use a bit of improvement: 
 

- [ ]  CODING_ADMIN should get a "delete" delete entries. Needs a new permission in permissions.ts, used on both frontend and backend 

- [ ] All ADMINS should have a notification in their menu for when there is active applications which they didn't vote for yet. This needs to be implemented in a minimally "invasive" manner so that we don't run a heavy query for all users doing this check on each pageload.

- [ ] On the list of applications, /manual/staff/applications/, show current voting status

rajanarahul93:
# Pull Request

## What was implemented

This PR improves the staff applications system with the following changes:

1. **CODING_ADMIN delete permission**
   - Added a new `canDeleteStaffApplication` permission.
   - Implemented a secure backend `delete` mutation for staff applications.
   - UI now shows a Delete button only for CODING_ADMIN, with confirmation and disabled state while processing.

2. **Voting status on applications list**
   - Backend now enriches the applications list with a `myVote` field for the current user.
   - UI displays a new **Your Vote** column showing:
     - Approved  
     - Rejected  
     - Not voted

3. **Admin notification for pending votes**
   - Added a lightweight `pendingVoteCount` query.
   - Uses SQL `LEFT JOIN` + `IS NULL` to only count applications:
     - Still `PENDING`
     - Not yet voted on by the current admin
     - Belonging to the admin’s approval group.
   - Hooked into `UserContext` to show a notification in the admin menu.

4. **Frontend safety & UX**
   - Delete button is disabled during mutation to avoid double submits.
   - Application list is invalidated precisely after deletion using the current filter parameters.

---

## Why this was implemented

The staff application system lacked three core features:

- CODING_ADMIN couldn’t delete applications.
- Staff had no visibility into whether they had voted or not.
- Admins had no clear way to know when applications were waiting for their vote.

This PR closes those gaps while keeping the database queries minimal and non-invasive, in line with the issue’s request.

---

## Breaking changes

No breaking changes.  
All changes are backward-compatible and scoped only to the staff application feature.

---

## License

fixes #780 

## Summary by CodeRabbit

* **New Features**
  * Applications table shows "Your Vote" (Approved / Rejected / Not voted).
  * Authorized users (coding-admin) see a Delete action for pending applications with confirmation, disabled state while processing, and error handling.
  * Deleting an application refreshes the list immediately while preserving current filters.
  * Admins get a navbar notification with the count of pending staff application approvals.

### DIFF
--- a/app/src/server/api/routers/profile.ts
+++ b/app/src/server/api/routers/profile.ts
@@ -22,6 +22,8 @@ import {
   poll,
   quest,
   questHistory,
+  staffApplication,
+  staffApplicationApproval,
   userAttribute,
   userBlackList,
   userData,


app/src/app/manual/staff/applications/page.tsx
```diff
@@ -65,6 +76,37 @@ export default function ApplicationsPage() {
...
+        {me && canDeleteStaffApplication(me.role) && (
+          <button
+            disabled={deleteMutation.isPending}
+            className={`text-red-600 hover:underline ${
+              deleteMutation.isPending ? "opacity-50 cursor-not-allowed" : ""
+            }`}
+            onClick={async (e) => {
+              e.preventDefault();
+              if (deleteMutation.isPending) return;
+              const ok = window.confirm("Delete this application? This cannot be undone.");
```
MathiasGruber: We have a @Confirm component, which we'd usually use for confirm dialogs. Also there is a @Button component which we usually use for actions in table. See other places in the codebase for reference :)

app/src/server/api/routers/applications.ts
```diff
@@ -134,18 +135,80 @@ export const applicationsRouter = createTRPCRouter({
...
       const nextCursor = results.length < limit ? null : currentCursor + 1;
-      return { data: results, nextCursor };
+      return { data: enriched, nextCursor };
     }),
 
+  delete: protectedProcedure
+    .input(z.object({ id: z.string() }))
+    .output(baseServerResponse)
+    .mutation(async ({ ctx, input }) => {
+      const user = await fetchUser(ctx.drizzle, ctx.userId);
```
MathiasGruber: For consistency, we try to wrap the different parts in comments like // query, // guards, // mutations etc, for better readability :)

app/src/server/api/routers/applications.ts
```diff
@@ -134,18 +135,80 @@ export const applicationsRouter = createTRPCRouter({
...
+      if (!user) return errorResponse("Not allowed");
+      if (!canDeleteStaffApplication(user.role)) return errorResponse("Not allowed");
+
+      const app = await ctx.drizzle.query.staffApplication.findFirst({
```
MathiasGruber: user & app can be fetched in parallel at the store of the mutation for optimization. We want each endpoint to be as efficient as possible, as that reduce our server cost.

app/src/utils/UserContext.tsx
```diff
@@ -108,6 +108,53 @@ export function UserContextProvider(props: { children: React.ReactNode }) {
     });
   };
 
+  // Pending staff-application votes (admin) — lightweight query used to surface a
```
MathiasGruber: This is very unlike how we do it all other places for notifications. Think for now we should stick with the current approach, so that it becomes easier to refactor if we decide to clean up (I don't like the current approach particularly much either, but at least we can stay consistent)

app/src/server/api/routers/applications.ts
```diff
@@ -134,18 +135,80 @@ export const applicationsRouter = createTRPCRouter({
...
+      });
+      if (!app) return errorResponse("Application not found");
+      if (app.state !== "PENDING")
+        return errorResponse("Only pending applications can be deleted");
+
+      await ctx.drizzle.delete(staffApplication).where(eq(staffApplication.id, input.id));
+      return { success: true, message: "Application deleted" };
+    }),
+
+  pendingVoteCount: protectedProcedure.query(async ({ ctx }) => {
```
MathiasGruber: I would not make a separate mutation for this. Instead I'd include it where we do the rest, i.e. in the profile route on line 531, we could introduce something like [pseudocode]

```typescript
let reportCounts: { count: number }[] = [];
      let ticketCounts: { count: number }[] = [];
      let applicationCounts: { count: number }[] = [];
      if(user.role !== "USER"){
        [reportCounts, ticketCounts, applicationCounts] = await Promise.all([
          canModerateRoles.includes(user.role) ? query1 : null,
          canModerateRoles.includes(user.role) ? query2 : null,
          user.role.includes("ADMIN") ? query3 : null,

        ]);
      }
```

app/src/server/api/routers/applications.ts
```diff
@@ -163,51 +163,28 @@ export const applicationsRouter = createTRPCRouter({
...
-        and(
-          eq(staffApplication.state, "PENDING"),
-          isNull(staffApplicationApproval.applicationId),
-        ),
-      );
-
-    return { count: rows.length };
-  }),
+  // NOTE: `pendingVoteCount` was intentionally removed here; its count is now
+  // included in `profile.getUser` to avoid a separate endpoint and keep
```
MathiasGruber: No need for these comments. Careful with these AI introduced messages :)

app/src/utils/UserContext.tsx
```diff
@@ -108,52 +108,9 @@ export function UserContextProvider(props: { children: React.ReactNode }) {
     });
   };
 
-  // Pending staff-application votes (admin) — lightweight query used to surface a
```
MathiasGruber: Again, no need for these comments introduced by the AI.

app/src/server/api/routers/profile.ts
```diff
@@ -565,6 +565,38 @@ export const profileRouter = createTRPCRouter({
             notificationCount: userTickets,
           });
         }
+
+        // Application counts for admins: pending applications that the current
+        // admin hasn't personally voted on yet (only if the role participates in approvals)
+        let applicationCounts = [{ count: 0 }];
+        if (user.role !== "USER" && StaffApprovalGroups.includes(user.role as any)) {
```
MathiasGruber: See earlier comments - we'd still like to run this query in parallel with the user reports fetching, in order to make this endpoint as efficient and fast as possible. Every round-about to the database takes maybe 100ms, so sequentially stacking them up results in bad UX.

MathiasGruber (approved):
Awesome, good work! 🫶

MathiasGruber (approved):
Awesome, good work! 🫶

PR #826, opened 2025-12-16
Set Jutsu/Items to PVP, PVE, or BOTH

Phrosfire:
# Pull Request

Content and Event requested the ability to set if a jutsu or item is for pvp, pve, or both.  This will allow them to create special jutsu that only work for one or the other, so they don't affect each other.

**Changes**
- Adds a battle type drop down on the item/jutsu edit page.
- Items/Jutsu with battle type set to "BOTH" will work in any battle.
- Items/Jutsu with battle type set to "PVE" or "BOTH" will be pulled into battle if its "QUEST" or "RANDOM ENCOUNTER".
- Items/Jutsu with battle type set to "PVP" or "BOTH" will be pulled into battle if its not "QUEST" or "RANDOM ENCOUNTER".

**UI Adjustment:**
- Selecting an item or jutsu will display the battle type.  
- Added Battle Type to the item, shop, and jutsu filter.
- Added Action Cost to the item, shop, and jutsu filter.

**Can you also please check your DM's from me?  I sent you a few questions.**

image

## License

## Summary by CodeRabbit

* **New Features**
  * Added new battle mode categories and expanded battle-type options for items and jutsus.
  * Combat now only offers items/jutsus compatible with the current battle mode.
  * Added "Battle Type" and "Action Cost (%)" filters across item, jutsu lists and shop (shop hook exposes setters).
  * Edit forms and item/jutsu details show Battle Type (displays "PVP & PVE" for BOTH).

### DIFF
--- a/app/drizzle/constants.ts
+++ b/app/drizzle/constants.ts
@@ -442,18 +442,18 @@ export const PvpBattleTypes: BattleType[] = [
   "TOURNAMENT",
   "RANKED_SPARRING",
   "KAGE_PVP",
+  "KAGE_AI",
   "RANKED_PVP",
+  "SHRINE_WAR",
 ];
 
 export const PveBattleTypes: BattleType[] = [
   "ARENA",
   "QUEST",
   "RANDOM_ENCOUNTER",
-  "KAGE_AI",
   "TRAINING",
   "VILLAGE_PROTECTOR",
   "CLAN_CHALLENGE",
-  "SHRINE_WAR",
 ];
 
 export const QuestBattleTypes: BattleType[] = [


app/src/libs/combat/actions.ts
```diff
@@ -118,6 +118,20 @@ export const availableUserActions = (
     ...(user?.jutsus
       ? user.jutsus
           .filter((userjutsu) => {
+            // Filter by battleUsageType
+            if (battle) {
+              const isPvEBattle =
+                battle.battleType === "QUEST" ||
+                battle.battleType === "RANDOM_ENCOUNTER";
```
MathiasGruber: What about all the other PVE battle types? probably easier to get the PVP battle types here?

app/src/libs/combat/actions.ts
```diff
@@ -152,6 +166,20 @@ export const availableUserActions = (
           .filter((ui) => {
             if (ui.quantity <= 0) return false;
             if (ui.item.preventBattleUsage) return false;
+            // Filter by battleUsageType
+            if (battle) {
+              const isPvEBattle =
+                battle.battleType === "QUEST" ||
```
MathiasGruber: Same here, we have a lot of PVE types

app/src/server/api/routers/combat.ts
```diff
@@ -2162,13 +2162,23 @@ export const processUsersForBattle = async (
     }
 
     // Set jutsus updatedAt to now (we use it for determining usage cooldowns)
+    const isPvEBattle = battleType === "QUEST" || battleType === "RANDOM_ENCOUNTER";
```
MathiasGruber: We should probably define the list of PVE/PVP battle types all the way in the constants file, so we can use it all these places we check.

app/src/libs/combat/actions.ts
Phrosfire: The only PVE types that need it are QUEST and RANDOM_ENCOUNTER.    The battle types such as Arena and Training are used for leveling jutsu, so we shouldn't take that away from people who want to level PVP jutsu for it.  The VILLAGE_PROTECTOR PVE fight comes from infiltrating a village, so we should be allowing PVP jutsu to be used since PVP is the primary reason to enter another village.

app/src/server/api/routers/combat.ts
Phrosfire: I create a list of pve battle types and then an additional list for just quest battle types which would be quest and random_encounter.  It pulls in the quest battle type list for using pve jutsu since those two are the only ones that need to use it.  If we create more pve battle types that need it such as dungeons or world bosses, it can go in there.

Issue #626, opened 2025-08-06
Support Ticket: Scheduled Shrine Boosts


MathiasGruber:
**Original Support Ticket**: https://www.theninja-rpg.com/support/nkH6TdoR1p

**Category**: FEATURE_REQUEST
**Priority**: MEDIUM
**Status**: OPEN
**Created by**: Malsz
**Created at**: 2025-08-06T15:10:15.748Z

**Description**:
Thank you for implementing the shrine system.

However, a schedule feature would be a really nice addition to have as it would allow more players to enjoy the benefits without being bound to a the kage’s timezone and/or the kage to not have to set an alarm at 3AM to activate the boost.

Appreciate your time reading this!

**Comments**:

**Malsz** - 2025-08-06T15:10:15.749Z:
Thank you for implementing the shrine system.
<br />
<br />
However, a schedule feature would be a really nice addition to have as it would allow more players to enjoy the benefits without being bound to a the kage’s timezone and/or the kage to not have to set an alarm at 3AM to activate the boost.
<br />
<br />
Appreciate your time reading this!

TheNinja-RPG:
[OpenHands](https://github.com/All-Hands-AI/OpenHands) started fixing the issue! You can monitor the progress [here](https://github.com/studie-tech/TheNinjaRPG/actions/runs/16851466154).

milanig-creator:
I am taking a look at this one

PR #833, opened 2025-12-20
Shrine Schedule: Addresses Support Ticket: Scheduled Shrine Boosts (#626)

milanig-creator:
# Pull Request

## Summary

This PR implements **scheduled Shrine Boosts**, addressing Support Ticket **#626**.

Village leaders can now schedule shrine boosts to activate and expire automatically at defined times, removing the need for manual activation and ensuring consistency across restarts and deployments.

---

## What Was Implemented

- Added a new `ShrineBoostSchedule` table to persist scheduled boosts
- Implemented server-side scheduling logic to:
  - Activate boosts when their start time is reached
  - Expire boosts cleanly when their duration ends
- Integrated the scheduler with the existing Shrine system
- Updated Shrine UI to display and manage scheduled boosts
- Added cron configuration to ensure scheduled boosts are processed reliably in production

---

## Why This Change

Previously, shrine boosts required manual activation and could not be reliably scheduled in advance.  
This caused operational friction for village leadership and inconsistencies during downtime or missed activations.

This change provides:
- Predictable boost timing
- Reduced manual intervention
- Better long-term scalability for shrine mechanics

---

## Testing

- Tested locally and in dev
- Walked through scheduling, activation, and expiration flows
- Confirmed boosts activate and expire as expected
- Verified UI reflects scheduled and active states correctly

---

## Breaking Changes

- None expected

---

## Notes for Reviewers

- Scheduler logic is defensive and idempotent
- UI changes are null-safe if no schedules exist
- Open to follow-up refinements if preferred patterns differ

## License

## Summary by CodeRabbit

* **New Features**
  * Shrine boost scheduling: calendar/time picker, per-boost scheduling UI, and APIs to list/schedule/cancel boosts.
  * New Calendar component for date selection.

* **Improvements**
  * Minute-by-minute maintenance tick for more responsive boost activation and cleanup.
  * Shrine Hall UI: activate now or schedule boosts, manage scheduled boosts, improved active-boost display.
  * Stronger concurrency and token safeguards across shrine actions.

* **Chores**
  * New time utilities and a minute-based timer utility.
  * Increased maintenance cron frequency.

---

> [!NOTE]
> Introduces persistence for scheduled shrine boosts and refines related constants/guidelines.
> 
> - Creates `ShrineBoostSchedule` table (with indexes on `villageId`, `boostType`, `startAt`, `endAt`) to store scheduled boosts
> - Adds `MAX_BOOSTS_PER_SHRINE = 50` in `app/drizzle/constants.ts`
> - Updates `CLAUDE.md` DB patterns to emphasize minimizing database roundtrips (prefer parallel queries with `Promise.all()`)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 609ce63fadaa35c4e7bf92a586f7890d7fb74552. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/server/api/routers/shrine.ts
+++ b/app/src/server/api/routers/shrine.ts
@@ -1,7 +1,8 @@
 import { z } from "zod";
 import { nanoid } from "nanoid";
 import { createTRPCRouter, protectedProcedure, publicProcedure } from "../trpc";
-import { baseServerResponse, errorResponse } from "../trpc";
+import { baseServerResponse, errorResponse, serverError } from "../trpc";
+import { canSeeSecretData } from "@/utils/permissions";
 import { eq, and, gte, sql, asc, gt, isNull, desc, inArray } from "drizzle-orm";
 import {
   SHRINE_UPGRADE_COST,
@@ -86,16 +87,28 @@ export const shrineRouter = createTRPCRouter({
     .query(async ({ ctx, input }) => {
       const now = new Date();
 
-      const schedules = await ctx.drizzle
-        .select()
-        .from(shrineBoostSchedule)
-        .where(
-          and(
-            eq(shrineBoostSchedule.villageId, input.villageId),
-            gt(shrineBoostSchedule.startAt, now), // hide boosts that already started
-          ),
-        )
-        .orderBy(asc(shrineBoostSchedule.startAt));
+      // Run queries in parallel
+      const [user, schedules] = await Promise.all([
+        fetchUser(ctx.drizzle, ctx.userId),
+        ctx.drizzle
+          .select()
+          .from(shrineBoostSchedule)
+          .where(
+            and(
+              eq(shrineBoostSchedule.villageId, input.villageId),
+              gt(shrineBoostSchedule.startAt, now), // hide boosts that already started
+            ),
+          )
+          .orderBy(asc(shrineBoostSchedule.startAt)),
+      ]);
+
+      // Authorization: only allow access if user belongs to the village or has admin/moderator access
+      if (user.villageId !== input.villageId && !canSeeSecretData(user.role)) {
+        throw serverError(
+          "FORBIDDEN",
+          "You can only view scheduled boosts for your own village",
+        );
+      }
 
       return schedules;
     }),
@@ -367,22 +380,27 @@ export const shrineRouter = createTRPCRouter({
         return errorResponse("Not enough village tokens to schedule boost");
       }
 
-      try {
-        await ctx.drizzle.insert(shrineBoostSchedule).values({
-          id: nanoid(),
-          villageId: input.villageId,
-          boostType: input.boostType,
-          startAt,
-          endAt,
-          createdByUserId: ctx.userId,
-        });
-      } catch (_err) {
-        // best-effort refund (guarded)
+      // Atomic guarded insert to prevent concurrent overlap
+      const scheduleId = nanoid();
+      const insertRes = await ctx.drizzle.execute(sql`
+        INSERT INTO ShrineBoostSchedule (id, villageId, boostType, startAt, endAt, createdByUserId, createdAt, updatedAt)
+        SELECT ${scheduleId}, ${input.villageId}, ${input.boostType}, ${startAt}, ${endAt}, ${ctx.userId}, NOW(3), NOW(3)
+        WHERE NOT EXISTS (
+          SELECT 1 FROM ShrineBoostSchedule s
+          WHERE s.villageId = ${input.villageId}
+            AND s.boostType = ${input.boostType}
+            AND s.startAt < ${endAt}
+            AND s.endAt > ${startAt}
+        )
+      `);
+
+      if (insertRes.rowsAffected === 0) {
+        // Overlap detected by another concurrent request - refund tokens
         await ctx.drizzle
           .update(village)
           .set({ tokens: sql`${village.tokens} + ${SHRINE_BOOST_COST}` })
           .where(eq(village.id, user.villageId));
-        return errorResponse("Failed to create schedule. Tokens were refunded.");
+        return errorResponse("Boost time overlaps with an existing scheduled boost");
       }
 
       return {
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -116,6 +116,7 @@ The combat system is the most complex feature, with dedicated files:
 - Schema is centralized in `@/drizzle/schema.ts`
 - We use the react compiler, and therefore must use useWatch hook, not watch, for react-hook-form.
 - **No Legacy Fields**: When refactoring database schema, fully remove legacy/deprecated fields rather than keeping them for backward compatibility. Do not leave legacy fields in the schema - migrate all code to use new field names immediately.
+- **Minimize DB Roundtrips**: For queries, prefer reducing the number of database roundtrips over reducing the amount of data fetched. Running queries in parallel with `Promise.all()` is faster than running them sequentially, even if it means fetching slightly more data upfront.
 
 ## tRPC Patterns
 


app/src/layout/ShrineHall.tsx
```diff
@@ -393,21 +424,169 @@ const BoostsTab = ({ user, isActive }: TabProps) => {
...
+                              return;
+                            }
+                            
+                            // Create date in UTC (treating user input as UTC time)
+                            // This ensures no timezone conversion happens
+                            const utcDate = new Date(Date.UTC(year, month - 1, day, hours, minutes));
+                            scheduleBoost({
+                              boostType,
+                              villageId: user.villageId!,
+                              startAt: utcDate.toISOString(),
```
MathiasGruber: No need to run this validation, this is happening within tRPC on the scheduleBoost endpoint which specified a zod-validator.

app/src/layout/ShrineHall.tsx
```diff
@@ -393,21 +424,169 @@ const BoostsTab = ({ user, isActive }: TabProps) => {
...
+                            // Set default to 1 hour from now
+                            const defaultDate = new Date();
+                            defaultDate.setHours(defaultDate.getHours() + 1);
+                            setScheduleDateTime(
+                              defaultDate.toISOString().slice(0, 16),
+                            );
+                          }
+                        }}
+                      >
+                        {isScheduling ? "Schedule" : "Schedule..."}
```
MathiasGruber: I think it'll be more aesthetic to simply have a calender icon here instead of text

app/src/layout/ShrineHall.tsx
```diff
@@ -393,21 +424,169 @@ const BoostsTab = ({ user, isActive }: TabProps) => {
...
+                      </Button>
+                    </div>
+                    {isScheduling && (
+                      <div className="space-y-3 pl-2 border-l-2 border-muted">
```
MathiasGruber: I think this can be styled a bit better. Usually we use borders with rounded corners, and maybe the bg-popover or other color than the background

app/src/layout/ShrineHall.tsx
```diff
@@ -393,21 +424,169 @@ const BoostsTab = ({ user, isActive }: TabProps) => {
...
+                          <Label htmlFor={`schedule-${boostType}`} className="text-sm">
+                            Start Date & Time
+                          </Label>
+                          <Input
+                            id={`schedule-${boostType}`}
+                            type="datetime-local"
+                            value={scheduleDateTime}
+                            min={new Date().toISOString().slice(0, 16)}
+                            max={
+                              new Date(Date.now() + 7 * 24 * 60 * 60 * 1000)
```
MathiasGruber: We have a secondsFromNow convenience method & DAY_S constant that can be used

app/src/layout/ShrineHall.tsx
```diff
@@ -393,21 +424,169 @@ const BoostsTab = ({ user, isActive }: TabProps) => {
...
+                        {/* New Schedule Input */}
+                        <div className="space-y-2">
+                          <Label htmlFor={`schedule-${boostType}`} className="text-sm">
+                            Start Date & Time
+                          </Label>
+                          <Input
+                            id={`schedule-${boostType}`}
+                            type="datetime-local"
+                            value={scheduleDateTime}
+                            min={new Date().toISOString().slice(0, 16)}
```
MathiasGruber: I think the input here can simply be the date, maybe with toLocalString.. I don't thinkwe need to splice etc, but I could be wrong.

app/src/layout/ShrineHall.tsx
MathiasGruber: Also, I think we should have an explicit submit button next to the close button. It is not intuitive that this button has to be pressed twice to me

app/src/server/api/routers/shrine.ts
```diff
@@ -180,6 +192,181 @@ export const shrineRouter = createTRPCRouter({
...
+  scheduleBoost: protectedProcedure
+    .input(
+      z.object({
+        boostType: z.enum(SHRINE_BOOST_TYPES),
+        villageId: z.string(),
+        startAt: z.string().datetime().optional(), // ISO datetime string
+      }),
+    )
+    .output(baseServerResponse)
+    .mutation(async ({ ctx, input }) => {
```
MathiasGruber, line 272: Add a MAX_BOOSTS_PER_SHRINE to the constants file, and check it here. Malicious user could create thousands of schedules on a given shrine, which would not be good for the system

app/src/server/api/routers/shrine.ts
```diff
@@ -180,6 +192,181 @@ export const shrineRouter = createTRPCRouter({
...
+          endAt: endAt,
+          createdByUserId: ctx.userId,
+        }),
+        ctx.drizzle
+          .update(village)
+          .set({
+            tokens: sql`${village.tokens} - ${SHRINE_BOOST_COST}`,
+          })
+          .where(eq(village.id, user.villageId)),
+      ]);
```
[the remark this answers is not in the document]
MathiasGruber: Agreed with coderabbit

app/vercel.json
```diff
@@ -79,6 +79,10 @@
     {
       "path": "/api/auction-expire",
       "schedule": "*/5 * * * *"
+    },
+    {
+      "path": "/api/shrine-boost-scheduler",
+      "schedule": "* * * * *"
```
MathiasGruber: Vercel by default only gives 20 cron jobs, so we cannot add more without aggregating previous ones. I believe this new one should be built into the previous shrine-maintainance job though. And then we can change that one to run every minute

app/src/app/api/shrine-boost-scheduler/route.ts
```diff
@@ -0,0 +1,105 @@
...
+  activeAiIds: string[];
+};
+function defaultShrineSettings(): ShrineSettings {
+  return {
+    unlockedAiIds: [],
+    activeBoosts: {},
+    activeAiIds: [],
+  };
+}
+export async function runShrineBoostTick(now: Date = new Date()) {
```
MathiasGruber: Be very careful with these - they need to run as efficient as possible, with as few sequential DB calls as possible. Doing a DB call within a loop will make it take too long. We also need to lock it down so people can't DDOS this endpoint by calling it a million times per second. I'd refactor to something like this:

```typescript
import { drizzleDB } from "@/server/db";
import { shrineBoostSchedule, village } from "@/drizzle/schema";
import { and, lte, gt, eq, inArray } from "drizzle-orm";
import {
  handleEndpointError,
  lockWithMinuteTimer,
  updateGameSetting,
} from "@/libs/gamesettings";
import { cookies } from "next/headers";

const ENDPOINT_NAME = "shrine-boost-scheduler";

type ShrineSettings = {
  unlockedAiIds: string[];
  activeBoosts: Record<string, string>; // boostType -> ISO endAt
  activeAiIds: string[];
};

const defaultShrineSettings = (): ShrineSettings => ({
  unlockedAiIds: [],
  activeBoosts: {},
  activeAiIds: [],
});

export async function GET() {
  // Disable cache for this server action
  await cookies();

  // Check timer - run once per minute
  const timerCheck = await lockWithMinuteTimer(drizzleDB, ENDPOINT_NAME);
  if (!timerCheck.isNewMinute && timerCheck.response) return timerCheck.response;

  try {
    const now = new Date();

    // Fetch active and expired schedules in parallel
    const [activeSchedules, expiredSchedules] = await Promise.all([
      drizzleDB
        .select()
        .from(shrineBoostSchedule)
        .where(
          and(
            lte(shrineBoostSchedule.startAt, now),
            gt(shrineBoostSchedule.endAt, now),
          ),
        ),
      drizzleDB
        .select()
        .from(shrineBoostSchedule)
        .where(
          and(
            lte(shrineBoostSchedule.endAt, now),
            lte(shrineBoostSchedule.startAt, now),
          ),
        ),
    ]);

    // Process active and expired schedules in parallel
    const [activeApplied, expiredProcessed] = await Promise.all([
      processActiveSchedules(activeSchedules),
      processExpiredSchedules(expiredSchedules),
    ]);

    return Response.json({
      success: true,
      message: `Applied ${activeApplied} active boosts, processed ${expiredProcessed} expired schedules`,
    });
  } catch (cause) {
    // Rollback
    await updateGameSetting(drizzleDB, ENDPOINT_NAME, 0, timerCheck.prevTime);
    return handleEndpointError(cause);
  }
}

/**
 * Apply active boost schedules to villages
 */
async function processActiveSchedules(
  schedules: (typeof shrineBoostSchedule.$inferSelect)[],
) {
  if (schedules.length === 0) return 0;

  // Batch fetch all relevant villages
  const villageIds = [...new Set(schedules.map((s) => s.villageId))];
  const villages = await drizzleDB.query.village.findMany({
    where: inArray(village.id, villageIds),
    columns: { id: true, shrineSettings: true },
  });

  const villageMap = new Map(villages.map((v) => [v.id, v]));

  // Prepare updates - filter to only those that need updating
  const updates = schedules
    .map((schedule) => {
      const v = villageMap.get(schedule.villageId);
      if (!v) return null;

      const settings: ShrineSettings =
        (v.shrineSettings as ShrineSettings | null) ?? defaultShrineSettings();
      const currentEndTime = settings.activeBoosts?.[schedule.boostType];
      const newEndTime = schedule.endAt.toISOString();

      // Skip if already set to the same end time (idempotent check)
      if (currentEndTime === newEndTime) return null;

      return {
        villageId: schedule.villageId,
        shrineSettings: {
          ...settings,
          activeBoosts: {
            ...(settings.activeBoosts ?? {}),
            [schedule.boostType]: newEndTime,
          },
        },
      };
    })
    .filter((u) => u !== null);

  // Execute all updates in parallel
  if (updates.length > 0) {
    await Promise.all(
      updates.map((update) =>
        drizzleDB
          .update(village)
          .set({ shrineSettings: update.shrineSettings })
          .where(eq(village.id, update.villageId)),
      ),
    );
  }

  return updates.length;
}

/**
 * Process expired schedules: remove boost from village settings and delete schedule
 */
async function processExpiredSchedules(
  schedules: (typeof shrineBoostSchedule.$inferSelect)[],
) {
  if (schedules.length === 0) return 0;

  // Batch fetch all relevant villages
  const villageIds = [...new Set(schedules.map((s) => s.villageId))];
  const villages = await drizzleDB.query.village.findMany({
    where: inArray(village.id, villageIds),
    columns: { id: true, shrineSettings: true },
  });

  const villageMap = new Map(villages.map((v) => [v.id, v]));

  // Prepare village updates and collect schedule IDs to delete
  const villageUpdates: { villageId: string; shrineSettings: ShrineSettings }[] = [];
  const scheduleIdsToDelete: string[] = [];
  let processedCount = 0;

  for (const schedule of schedules) {
    scheduleIdsToDelete.push(schedule.id);

    const v = villageMap.get(schedule.villageId);
    if (!v?.shrineSettings) continue;

    const settings = v.shrineSettings as ShrineSettings;
    const activeBoosts = settings.activeBoosts ?? {};

    // Skip if boost not present
    if (!(schedule.boostType in activeBoosts)) continue;

    // Remove the expired boost
    const { [schedule.boostType]: _, ...remainingBoosts } = activeBoosts;
    void _; // Suppress unused variable warning

    villageUpdates.push({
      villageId: schedule.villageId,
      shrineSettings: { ...settings, activeBoosts: remainingBoosts },
    });
    processedCount++;
  }

  // Execute deletions and updates in parallel
  await Promise.all([
    // Delete all expired schedules in one batch
    scheduleIdsToDelete.length > 0
      ? drizzleDB
          .delete(shrineBoostSchedule)
          .where(inArray(shrineBoostSchedule.id, scheduleIdsToDelete))
      : Promise.resolve(),
    // Update villages in parallel
    ...villageUpdates.map((update) =>
      drizzleDB
        .update(village)
        .set({ shrineSettings: update.shrineSettings })
        .where(eq(village.id, update.villageId)),
    ),
  ]);

  return processedCount;
}
```

Which of course then requires a new function in @gamesettings.ts:

```typescript
export const lockWithMinuteTimer = async (client: DrizzleClient, name: string) => {
  const timer = await getGameSetting(client, name);
  const prevTime = timer.time;
  const now = new Date();
  const isNewMinute =
    now.getUTCMinutes() !== prevTime.getUTCMinutes() ||
    now.getUTCHours() !== prevTime.getUTCHours() ||
    now.getUTCDate() !== prevTime.getUTCDate();
  let response: string | null = null;
  if (!isNewMinute) {
    response = "Wait until the next minute to run this again";
  } else {
    await updateGameSetting(client, name, 0, new Date());
  }
  return { isNewMinute, prevTime, response: Response.json(response, { status: 200 }) };
};
```

app/drizzle/migrations/meta/_journal.json
```diff
@@ -2189,22 +2189,8 @@
...
-    },
-    {
-      "idx": 314,
-      "version": "5",
-      "when": 1766302186625,
-      "tag": "0314_wise_chat",
+      "when": 1766192750157,
+      "tag": "0312_sparkling_glorian",
       "breakpoints": false
     }
```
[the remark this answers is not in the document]
milanig-creator: This was addressed in latest commit.

app/src/app/api/shrine-boost-scheduler/route.ts
milanig-creator: Excellent point here, I should have addressed this in my latest commit

app/vercel.json
milanig-creator: Consolidated CRON jobs as requested

app/src/server/api/routers/shrine.ts
milanig-creator: Max shrine boosts set to 50. My thinking here is if we are extending this to a max of 7 days scheduling, there could be legitimate scheduling of 5 boosts a day.

app/src/layout/ShrineHall.tsx
milanig-creator: Addressed

app/src/layout/ShrineHall.tsx
milanig-creator: Added submit button to the calendar UI

app/src/layout/ShrineHall.tsx
milanig-creator: Addressed, let me know what you think on the styling here

app/src/layout/ShrineHall.tsx
milanig-creator: Good call — this logic was removed during the scheduler consolidation.
Date bounds are now enforced centrally in the scheduler/server logic rather than the UI, so the inline date math in ShrineHall.tsx is no longer needed here.

app/src/layout/ShrineHall.tsx
milanig-creator: Yes — this was addressed as part of the scheduler consolidation.
Validation and bounds checking now live entirely in the scheduleBoost tRPC layer, and the UI should only pass the selected start time without enforcing additional rules.

app/src/layout/ShrineHall.tsx
```diff
@@ -393,21 +424,169 @@ const BoostsTab = ({ user, isActive }: TabProps) => {
...
+                                  return (
+                                    <div
+                                      key={schedule.id}
+                                      className="flex items-center justify-between p-2 bg-muted rounded text-sm"
+                                    >
+                                      <div className="flex-1">
+                                        <div className="font-medium">
+                                          {startDate.toLocaleString()} -{" "}
+                                          {endDate.toLocaleString()}
+                                        </div>
```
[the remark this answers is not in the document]
milanig-creator: Added a disclaimer that time is in the local users time

app/src/app/api/shrine-boost-scheduler/route.ts
```diff
@@ -0,0 +1,304 @@
+// app/src/app/api/shrine-boost-scheduler/route.ts
+import { drizzleDB } from "@/server/db";
+import { shrineBoostSchedule, village } from "@/drizzle/schema";
+import { and, lte, gt, eq, inArray } from "drizzle-orm";
+import { cookies, headers } from "next/headers";
+import { lockWithMinuteTimer } from "@/libs/gamesettings";
```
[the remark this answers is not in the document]
milanig-creator: I see the issue,will address when I'm back home

app/src/app/api/shrine-boost-scheduler/route.ts
```diff
@@ -0,0 +1,305 @@
+// app/src/app/api/shrine-boost-scheduler/route.ts
+import { drizzleDB } from "@/server/db";
+import { shrineBoostSchedule, village } from "@/drizzle/schema";
+import { and, lte, gt, eq, inArray } from "drizzle-orm";
+import { cookies, headers } from "next/headers";
+import { lockWithHourlyTimer } from "@/libs/gamesettings";
```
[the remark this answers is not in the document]
milanig-creator: The shrine boost scheduler now uses a minute-level lock to match the * * * * * cron schedule.

Added lockWithMinuteTimer to libs/gamesettings.ts following the existing timer helpers

Updated /api/shrine-boost-scheduler to use lockWithMinuteTimer and isNewMinute instead of the hourly lock

Fixed the indentation issue in the timer guard

app/src/app/api/shrine-boost-scheduler/route.ts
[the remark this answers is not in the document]
milanig-creator: Confirmed.

lockWithMinuteTimer has been added to libs/gamesettings.ts in commit c1cfe60, and the scheduler now imports and uses it correctly.

Build and typecheck verified locally.

milanig-creator:
The recent commits include follow-up changes to restore the Shrine scheduling UI and address a few regressions that occurred during a prior build-fix iteration (including calendar sizing and cancellation behavior). I’m holding off on squashing until review to keep discussion context intact.

app/src/app/api/shrine-boost-scheduler/route.ts
```diff
@@ -0,0 +1,298 @@
...
+  if (auth?.startsWith("Bearer ")) {
+    const token = auth.slice("Bearer ".length).trim();
+    if (token === CRON_SECRET) return true;
+  }
+
+  const xSecret = h.get("x-cron-secret");
+  if (xSecret && xSecret === CRON_SECRET) return true;
+
+  return false;
+}
```
Phrosfire: We don't need to use a cron secret for authorization.  This along with the other references here can be removed and it'll work perfectly fine :)

Phrosfire:
The front end works great for scheduling boosts as well as cancelling them.  I did notice though that when a boost becomes active, it's not being removed from the scheduler.  To avoid clutter and prevent people from cancelling already active/completed boosts for a refund (Won't actually stop an already running boost), we should hide scheduled boosts with a start time earlier than the current time as well as put a safeguard on the delete function so that you can't delete already past/current scheduled boosts.

app/src/app/api/shrine-boost-scheduler/route.ts
milanig-creator: Thank you, I removed this

milanig-creator:
> The front end works great for scheduling boosts as well as cancelling them. I did notice though that when a boost becomes active, it's not being removed from the scheduler. To avoid clutter and prevent people from cancelling already active/completed boosts for a refund (Won't actually stop an already running boost), we should hide scheduled boosts with a start time earlier than the current time as well as put a safeguard on the delete function so that you can't delete already past/current scheduled boosts.

I have addressed this and tested in my latest commit. The past schedules are removed from view and added a two-layer safeguard on cancel:

an early check that rejects cancellation if startAt <= now

and a DB-level guard so the delete only succeeds when startAt > now

app/drizzle/migrations/0320_cold_mariko_yashida.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE `ShrineBoostSchedule` (
+	`id` varchar(191) NOT NULL,
```
MathiasGruber, line 2: Remove all migrations, rebase on top of main, and create a new set of migrations :) 

Perhaps the easiest way to do this, if it works, is to run `make git-branch-uncommit`, which will uncommit everything in your branch, remove all migration files, then create a new `make makemigrations`, and commit all in a clean commit :)

app/src/app/api/shrine-boost-scheduler/route.ts
```diff
@@ -0,0 +1,274 @@
...
+  return removedCount;
+}
+
+export async function runShrineBoostTick(now: Date = new Date()) {
+  const activeSchedules = await drizzleDB
+    .select()
+    .from(shrineBoostSchedule)
+    .where(and(lte(shrineBoostSchedule.startAt, now), gt(shrineBoostSchedule.endAt, now)));
+
+  const expiredSchedules = await drizzleDB
```
MathiasGruber: What's the reason for these running sequentially here, and then being passed to the functions? They could be run, in parallel, within, to minimize complexity. Alternatively, they are kept here, but run in parallel, together with the village query, which is run in both applyActive... and processExpired.. functions.

app/vercel.json
MathiasGruber: Am I missing something? still seems to be in the shrine-boost-scheduler route?

app/src/layout/ShrineHall.tsx
```diff
@@ -1,6 +1,7 @@
+// app/src/layout/ShrineHall.tsx
```
MathiasGruber: Why this comment?

MathiasGruber: Why this comment?

MathiasGruber (changes requested):
I feel like the whole time-handling is potentially over-engineered. This is a review by opus on the matter, which reflects my own thinking reading through this:

image

app/drizzle/migrations/0320_cold_mariko_yashida.sql
milanig-creator: This is now consolidated into a single commit. 

I believe I addressed what you asked for regarding the migrations, though this part I don't entirely understand. 

I am only seeing 0320_cold_mariko_yashida.sql. To my knowledge, I had initially deleted all migrations from main if I understand my mistake correctly. Im pretty sure this is not what you were asking for, and I reverted this.

app/vercel.json
milanig-creator: This was my fault-- not fully understanding, I'm sorry! I removed the dedicated cron entry from vercel.json and moved the scheduling trigger into the existing shrine maintenance cron.

Its still saying I have this in the file and I cannot find it at all at a second glance. Will look at this in a second

MathiasGruber:
@milanig-creator  seems to be a lot of conflicts. I'm having a look, and will force-push a clean version soon 👍

MathiasGruber:
@milanig-creator @theeneon  @Phrosfire  I've force pushed a single commit, which resolves all the merge conflicts. For reference, what I had to do to clean up here was:

`git rebase main`

Followed by resolving many merge conflicts, after each resolve use `git rebase --continue`

Finally, I ran `make git-branch-uncommit` to uncommit all the branch commits, and then made a single new commit.

I'm happy with the code at this point - let me know if I missed anything, and it works as intended.

theeneon:
@MathiasGruber From what I can see everything looks to work well. You can merge if they are any issues we can report it.

PR #862, opened 2026-01-10
feat: add MPVP shrine battle system

### LINKED ISSUE #659: MPVP Shrine Wars
This feature suggestion came about through discussing with players of wanting ways to team up and do things with their friends.  I know that we want to keep MPVP stuff for post release, but there is a hunger from the players for stuff they can do together and I think seeing as we have Clan wars which is 3v3 we can use parts of this system to use with other things in the game. 

Allow Shrines to be challenged by up to 3 players at the same time. 

Using the Clan War system that we already have, can it parts of it be repurposed so that when players attack a Shrine, they have to queue in. Once a minimum of 2 players are in the queue they can enter the Shrine to fight the AI that is present there. 

If Possible:

Allow villagers of the Shrine that is present to be the defenders so they can also queue in as a 3 man team to defend the shrine from the team of 3 man attackers.

MathiasGruber:
## Summary

This PR implements the MPVP Shrine Wars feature which allows players to team up and challenge shrines together.

**Changes:**
- Add `battleType`, `attackerEntityId`, `defenderEntityId`, and `sector` fields to `mpvpBattleQueue`
- Add `side` field to `mpvpBattleUser` for explicit attacker/defender designation
- Add MPVP battle type and shrine battle configuration constants
- Update clan battle router to use new entity fields
- Add shrine battle endpoints: `getShrineBattles`, `challengeShrine`, `joinShrineBattle`, `leaveShrineBattle`, `initiateShrineBattle`

**Features:**
- Up to 3 attackers can team up to attack a shrine
- Defenders from the shrine's village can queue to defend (up to 3)
- Falls back to AI defenders if no player defenders join
- 60 second lobby time before battle can start
- Minimum 2 attackers required to initiate battle

Closes #659

## Summary by CodeRabbit

* **New Features**
  * MPvP Shrine Team Battles UI with lobby and full create/join/leave/start flows; clan battles updated to use attacker/defender roles.

* **Behavioral Improvements**
  * Slot-based joining with unique-slot enforcement; lobby countdown (60s), min attackers (2), max 3 players per side; AI defender fallback and real-time notifications.

* **Documentation**
  * Added "No Legacy Fields" guideline.

### DIFF
--- a/app/src/app/shrine/page.tsx
+++ b/app/src/app/shrine/page.tsx
@@ -11,7 +11,11 @@ import Loader from "@/layout/Loader";
 import { Swords, Users, Shield } from "lucide-react";
 import Image from "@/layout/Image";
 import StatusBar from "@/layout/StatusBar";
-import { WAR_SHRINE_IMAGE, VILLAGE_SYNDICATE_ID } from "@/drizzle/constants";
+import {
+  WAR_SHRINE_IMAGE,
+  VILLAGE_SYNDICATE_ID,
+  MAP_RESERVED_SECTORS,
+} from "@/drizzle/constants";
 import RamenShop from "@/layout/RamenShop";
 import ShrineBattleLobby from "@/layout/ShrineBattleLobby";
 import { Tabs, TabsContent, TabsList, TabsTrigger } from "@/components/ui/tabs";
@@ -64,14 +68,29 @@ export default function Shrine() {
     (war) => war.sector === userData.sector,
   );
 
-  // Determine if this sector is owned by another village (for MPVP attacks)
+  // Determine if this sector is owned (for MPVP attacks/defends)
   const sectorOwnerVillageId = sectorData.sectorData?.villageId ?? null;
+  const userIsOwner = sectorOwnerVillageId === userData.villageId;
+  const isSyndicate = sectorOwnerVillageId === VILLAGE_SYNDICATE_ID;
+  const isAtWar = sectorData.warData?.some(
+    (war) =>
+      (war.attackerVillageId === userData.villageId ||
+        war.defenderVillageId === userData.villageId ||
+        war.warAllies?.some((a) => a.villageId === userData.villageId)) &&
+      (war.attackerVillageId === sectorOwnerVillageId ||
+        war.defenderVillageId === sectorOwnerVillageId),
+  );
   const canShowMpvpOption =
-    sectorOwnerVillageId && sectorOwnerVillageId !== userData.villageId;
+    !!sectorOwnerVillageId && (isAtWar || (isSyndicate && !userIsOwner));
 
   // Check if user is queued FOR THIS SECTOR - only then show the lobby
   const isUserQueuedForThisSector =
     userData.status === "QUEUED" && userQueuedBattle?.sector === userData.sector;
+
+  // Check if this sector is protected
+  const isReserved = MAP_RESERVED_SECTORS.includes(userData.sector);
+  const isHome = !!sectorData.village;
+  const isProtected = isReserved || isHome;
 
   if (!activeWars || activeWars.length === 0) {
     return (
@@ -94,7 +113,18 @@ export default function Shrine() {
                 </TabsTrigger>
               </TabsList>
               <TabsContent value="team" className="mt-4">
-                {canShowMpvpOption ? (
+                {userIsOwner ? (
+                  <div className="mb-4 text-sm text-muted-foreground">
+                    Your village owns this sector. You can join as a defender if an
+                    attack party is formed against this shrine.
+                  </div>
+                ) : isProtected ? (
+                  <div className="mb-4 text-sm text-amber-600 font-semibold">
+                    {isReserved
+                      ? "This sector is reserved and cannot be attacked."
+                      : "This is a village home sector and its shrine cannot be attacked."}
+                  </div>
+                ) : canShowMpvpOption ? (
                   <div className="mb-4 text-sm text-muted-foreground">
                     Form a team to attack this shrine together! Up to 3 attackers can
                     join the assault, and defenders from the owning village can queue to
@@ -102,8 +132,8 @@ export default function Shrine() {
                   </div>
                 ) : isUserQueuedForThisSector ? (
                   <div className="mb-4 text-sm text-muted-foreground">
-                    You are queued for a shrine battle. You can view your queue status or
-                    leave the queue below.
+                    You are queued for a shrine battle. You can view your queue status
+                    or leave the queue below.
                   </div>
                 ) : null}
                 <ShrineBattleLobby
@@ -111,47 +141,66 @@ export default function Shrine() {
                   userId={userData.userId}
                   userVillageId={userData.villageId}
                   defenderVillageId={sectorOwnerVillageId}
+                  isProtected={isProtected}
                 />
               </TabsContent>
               <TabsContent value="info" className="mt-4">
                 <div className="flex flex-col items-center">
                   {sectorData.sectorData ? (
-                    sectorOwnerVillageId === userData.villageId ? (
+                    userIsOwner ? (
                       <p>
-                        This sector is owned by your village. You can defend it when other
-                        villages attack.
+                        This sector is owned by your village. You can defend it when
+                        other villages attack.
                       </p>
                     ) : (
                       <p>
                         This sector is claimed by{" "}
-                        <strong>{sectorData.sectorData?.village.name}</strong>. The
-                        leader of your village or faction can attack the shrine to try
-                        and defeat it and claim the sector.
+                        <strong>{sectorData.sectorData?.village.name}</strong>.{" "}
+                        {isProtected
+                          ? "However, because it is a protected sector, it cannot be attacked."
+                          : "The leader of your village or faction can attack the shrine to try and defeat it and claim the sector."}
                       </p>
                     )
                   ) : (
                     <p>
-                      This sector is unclaimed. The leader of your village or faction can
-                      attack the shrine to try and defeat it and claim the sector.
+                      This sector is unclaimed.{" "}
+                      {!isProtected &&
+                        "The leader of your village or faction can attack the shrine to try and defeat it and claim the sector."}
                     </p>
                   )}
                 </div>
               </TabsContent>
             </Tabs>
           ) : (
-            <div className="flex flex-col items-center">
+            <div className="flex flex-col items-center text-center">
               {sectorData.sectorData ? (
-                <p>
-                  {" "}
-                  This sector is claimed by{" "}
-                  <strong>{sectorData.sectorData.village.name}</strong>. The leader of
-                  your village or faction can attack the shrine to try and defeat it and
-                  claim the sector.
-                </p>
+                userIsOwner ? (
+                  <p>
+                    This sector is owned by your village. You can defend it when other
+                    villages attack.
+                  </p>
+                ) : (
+                  <p>
+                    {" "}
+                    This sector is claimed by{" "}
+                    <strong>{sectorData.sectorData.village.name}</strong>.{" "}
+                    {isProtected
+                      ? "However, because it is a protected sector, it cannot be attacked."
+                      : "The leader of your village or faction can attack the shrine to try and defeat it and claim the sector."}
+                  </p>
+                )
               ) : (
                 <p>
-                  This sector is unclaimed. The leader of your village or faction can
-                  attack the shrine to try and defeat it and claim the sector.
+                  This sector is unclaimed.{" "}
+                  {!isProtected &&
+                    "The leader of your village or faction can attack the shrine to try and defeat it and claim the sector."}
+                </p>
+              )}
+              {isProtected && (
+                <p className="mt-2 text-amber-600 font-semibold">
+                  {isReserved
+                    ? "This sector is reserved and cannot be attacked."
+                    : "This is a village home sector and its shrine cannot be attacked."}
                 </p>
               )}
             </div>
@@ -166,20 +215,34 @@ export default function Shrine() {
   const userWars = activeWars.filter(
     (war) =>
       war.attackerVillageId === userData.villageId ||
-      war.defenderVillageId === userData.villageId,
+      war.defenderVillageId === userData.villageId ||
+      war.warAllies.some((wa) => wa.villageId === userData.villageId),
   );
   const competingWars = activeWars.filter(
     (war) =>
       war.attackerVillageId !== userData.villageId &&
-      war.defenderVillageId !== userData.villageId,
+      war.defenderVillageId !== userData.villageId &&
+      !war.warAllies.some((wa) => wa.villageId === userData.villageId),
   );
 
   // Check if user is an attacker in any of their wars (for Team Battle tab)
   const userIsAttacker = userWars.some(
-    (war) => war.attackerVillageId === userData.villageId,
+    (war) =>
+      war.attackerVillageId === userData.villageId ||
+      war.warAllies.some(
+        (wa) =>
+          wa.villageId === userData.villageId &&
+          wa.supportVillageId === war.attackerVillageId,
+      ),
   );
   const userIsDefender = userWars.some(
-    (war) => war.defenderVillageId === userData.villageId,
+    (war) =>
+      war.defenderVillageId === userData.villageId ||
+      war.warAllies.some(
+        (wa) =>
+          wa.villageId === userData.villageId &&
+          wa.supportVillageId === war.defenderVillageId,
+      ),
   );
 
   return (
@@ -212,6 +275,7 @@ export default function Shrine() {
                       sector={userData.sector}
                       onAttack={() => attack({ sector: userData.sector })}
                       isAttacking={isAttacking}
+                      isProtected={isProtected}
                     />
                   ))}
                 </div>
@@ -229,22 +293,32 @@ export default function Shrine() {
                         onAttack={() => {}}
                         isAttacking={false}
                         hideAttackButton={true}
+                        isProtected={isProtected}
                       />
                     ))}
                   </div>
                 )}
                 {userIsAttacker ? (
                   <>
-                    <div className="mb-4 text-sm text-muted-foreground">
-                      Form a team to attack this shrine together! Up to 3 attackers can
-                      join the assault, and defenders from the owning village can queue
-                      to defend.
-                    </div>
+                    {isProtected ? (
+                      <div className="mb-4 text-sm text-amber-600 font-semibold">
+                        {isReserved
+                          ? "This sector is reserved and cannot be attacked."
+                          : "This is a village home sector and its shrine cannot be attacked."}
+                      </div>
+                    ) : (
+                      <div className="mb-4 text-sm text-muted-foreground">
+                        Form a team to attack this shrine together! Up to 3 attackers
+                        can join the assault, and defenders from the owning village can
+                        queue to defend.
+                      </div>
+                    )}
                     <ShrineBattleLobby
                       sectorNumber={userData.sector}
                       userId={userData.userId}
                       userVillageId={userData.villageId}
                       defenderVillageId={sectorOwnerVillageId}
+                      isProtected={isProtected}
                     />
                   </>
                 ) : userIsDefender ? (
@@ -258,6 +332,7 @@ export default function Shrine() {
                       userId={userData.userId}
                       userVillageId={userData.villageId}
                       defenderVillageId={sectorOwnerVillageId}
+                      isProtected={isProtected}
                     />
                   </>
                 ) : (
@@ -286,6 +361,7 @@ export default function Shrine() {
                 sector={userData.sector}
                 onAttack={() => attack({ sector: userData.sector })}
                 isAttacking={isAttacking}
+                isProtected={isProtected}
               />
             ))}
           </div>
@@ -302,6 +378,7 @@ const WarCard = ({
   onAttack,
   isAttacking,
   hideAttackButton = false,
+  isProtected = false,
 }: {
   war: War & {
     attackerVillage: { name: string; villageGraphic: string };
@@ -312,10 +389,11 @@ const WarCard = ({
   onAttack: () => void;
   isAttacking: boolean;
   hideAttackButton?: boolean;
+  isProtected?: boolean;
 }) => {
   const isAttacker = war.attackerVillageId === villageId;
   // Only attackers can attack the shrine - defenders shouldn't attack their own shrine
-  const canAttack = isAttacker && war.shrineHp > 0 && !hideAttackButton;
+  const canAttack = isAttacker && war.shrineHp > 0 && !hideAttackButton && !isProtected;
 
   return (
     <div className="flex flex-col items-center gap-4 p-4">
--- a/app/src/layout/ShrineBattleLobby.tsx
+++ b/app/src/layout/ShrineBattleLobby.tsx
@@ -31,6 +31,7 @@ interface ShrineBattleLobbyProps {
   userId: string;
   userVillageId: string;
   defenderVillageId?: string | null;
+  isProtected?: boolean;
 }
 
 export const ShrineBattleLobby: React.FC<ShrineBattleLobbyProps> = ({
@@ -38,6 +39,7 @@ export const ShrineBattleLobby: React.FC<ShrineBattleLobbyProps> = ({
   userId,
   userVillageId,
   defenderVillageId,
+  isProtected = false,
 }) => {
   const utils = api.useUtils();
   const router = useRouter();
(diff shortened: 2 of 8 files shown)


MathiasGruber:
@claude please review https://github.com/studie-tech/TheNinjaRPG/issues/659, and make sure all things are implemented. e.g. I can see right now, we still have clan1id and cland2id - the point was to refactor those to use the attackerEntityId etc. Also, the frontend components also need to be done. Please update this PR to fully implement all functionality related to this feature.

MathiasGruber:
@claude I think we should fully remove the legacy fields - whenever you do something, do not leave legacy fields around. Please update the Claude.md to clearly state this for any future interactions. 

Also consider this from code rabbit:
<concern>
Initiation flow: avoid double-query, harden update guard, and fix CASE quoting

You already have shrineBattle.sector; the extra mpvpBattleQueue.findFirst(...).then(...) (Line 685-697) can be removed.

Update mpvpBattleQueue with a guard to prevent double-start races: where id = ? AND battleId IS NULL.

This CASE expression uses double-quoted literals (Line 768-769). Under MySQL ANSI_QUOTES, that can break. Prefer parameterized SQL literals.

Example hardening for the status CASE
Also: currently any queued participant can initiate (including defenders). If only attackers (or the party creator) should be allowed to start, add a guard.</concern>

<concern>
Correctness bug: queue deletion uses stale shrineBattle.queue snapshot

After removing the user (Line 629-642), you decide whether to delete mpvpBattleQueue based on the pre-fetched shrineBattle.queue (Line 644-650). If another user joins between the fetch and this delete, you can delete an active queue out from under them.

Safer approach: after deletion, query remaining count from DB and only delete when count is truly 0 (and ideally only when battleId IS NULL).

Proposed fix (guarded cleanup)
       await Promise.all([
         ctx.drizzle
           .delete(mpvpBattleUser)
@@
         ctx.drizzle
           .update(userData)
           .set({ status: "AWAKE" })
           .where(eq(userData.userId, user.userId)),
       ]);

-      // If no one left in queue, delete the battle queue
-      const remainingQueue = shrineBattle.queue.filter((q) => q.userId !== user.userId);
-      if (remainingQueue.length === 0) {
-        await ctx.drizzle
-          .delete(mpvpBattleQueue)
-          .where(eq(mpvpBattleQueue.id, input.shrineBattleId));
-      }
+      // If no one left in queue, delete the battle queue (DB-truth, not snapshot)
+      const [{ remaining }] = await ctx.drizzle
+        .select({ remaining: sql<number>`count(*)` })
+        .from(mpvpBattleUser)
+        .where(eq(mpvpBattleUser.clanBattleId, input.shrineBattleId));
+
+      if (remaining === 0) {
+        await ctx.drizzle
+          .delete(mpvpBattleQueue)
+          .where(
+            and(
+              eq(mpvpBattleQueue.id, input.shrineBattleId),
+              isNull(mpvpBattleQueue.battleId),
+            ),
+          );
+      }</concern>

<concern>
Queue capacity checks are race-prone (can exceed max per side under concurrent joins)

The side capacity checks (Line 558-573) are based on a snapshot read, then an insert happens later. Two concurrent joins can both pass the check and overfill a side.

Without transactions (per guidelines), consider a “slot” model: store slot (0..max-1) on MpvpBattleUser, enforce unique(clanBattleId, side, slot), and allocate the first free slot with retry on conflict. Alternatively, accept overfill but add server-side cleanup/denial on insert failure + a unique constraint strategy.</concern>

And finally:
In @app/src/app/shrine/page.tsx around lines 150 - 191, The defenderVillageId
computation in the ShrineBattleLobby props currently uses userWars[0]
(non-deterministic); replace that ternary with the existing sectorOwnerVillageId
value instead so defenderVillageId={sectorOwnerVillageId}, mirroring the pattern
used at line 91 and ensuring the lobby always receives the shrine owner's
village id rather than an arbitrary war's opponent.

MathiasGruber:
@claude when I go to the shrine in my own village, it shows me that I can attack the shrine in solo battle (see screenshot) 
image. For team battle it tells me there is no shrine battle for this sector. 

When creating a sector war, the text informs you that it'll cost 150000 points to start the war. But I believe, when trying to start a sector war, it'd be nice if the dialog said that this is a sector war being initiated (if there are nobody else owning the sector), instead of sounding like it's attacking another village. 

Finally, when I queued for the sector war, my screen goes to "This sector is unclaimed. The leader of your village or faction can attack the shrine to try and defeat it and claim the sector." screen. And I cannot see myself in the travel sector anymore, since my status was (correctly) changed to queued. In this state, I'd expect I can see the queue when I go to the shrine, and also be able to exit the queue if I wanted to.

MathiasGruber:
@claude when I declare a sector war, and go to the shrine in the sector, even though I'm the leader, I get the message "This sector is unclaimed. The leader of your village or faction can attack the shrine to try and defeat it and claim the sector.", but I don't get the option to actually attack it

MathiasGruber:
@claude Please consider this comment from coderabbit: 
<concern>
Verify queue sector matching to prevent confusing empty lobby UX.

The isUserQueued check (line 66) displays team battle tabs regardless of which sector the user is queued for. Users can travel between sectors while status === "QUEUED" (no travel restrictions prevent this), so a user queued for Sector 5 can navigate to Sector 7. The tabs will appear, but ShrineBattleLobby queries battles for the current sector, resulting in an empty lobby.

Consider either:

Preventing sector changes while queued, or
Validating that the queued battle is for the current sector before showing tabs
</concern>

Other concerns:
- If I start a solo battle, it asks me to wait in the lobby. This doesn't make sense, since when it's a solo battle (only one AI), we usually do not show lobby in combat. This should only be for PvP scenarios.
- When I go to team battle for an unoccupied sector, and nobody is in the attack party yet, it says "Battle can start". Even though clearly it cannot. 
- When I go to team battle, and nobody is in the attack party yet, it says "Battle can start". Even though clearly it cannot. 
- When I go to team battle, and nobody is in the attack party yet, it both shows the "Create shrine attack party", but also the "Join attack". The latter doesn't make much sense if no attack is initiated yet.
- When I go to team battle, and initiate an attack party, I see the following (see screenshot). There are several issues here - the avatar of my character is too big - I believe it looks better in our other mpvp queue components. Also, I see two "attackers and defenders" components, where the latter is empty. 

Please review the UI components thoroughly, to make sure overall their flow, logic, and visuals are correct and work across different scenarios.

MathiasGruber:
@claude Please address this concern from code rabbit:
In @app/drizzle/migrations/0323_exotic_true_believers.sql around lines 7 - 9,
The migration adds a UNIQUE constraint on MpvpBattleUser(clanBattleId, side,
slot) but leaves side and slot nullable; change the ALTER TABLE statements to
make side and slot NOT NULL (i.e., ALTER TABLE `MpvpBattleUser` ADD `side`
enum('ATTACKER','DEFENDER') NOT NULL; and ADD `slot` tinyint NOT NULL;) so NULLs
cannot bypass the uniqueness check, and also update the corresponding fields in
app/drizzle/schema.ts to call notNull() for side and slot to keep schema and
migration in sync.

Also remove the restrictions on git commands from Claude.md

Also, see screenshot, which is what I see when I go to an un-occupied sector 
image

I expect, when I'm in a given sector (202 in this case), I would only see sector wars relating to this sector. In this case I see two entries, of which both have zero users. I expect this may be because I've started a shrine party, and then abandoned it, leaving it with 0 users. When this happens, the queue should be deleted (i.e. if nobody is left in it). No need to run migrations for previous entries.

Please also fix the issue where if I start a solo battle, it's expected to be counted as an "AI battle", since it's not PvP, and therefore we do not wish to wait in the combat lobby.

MathiasGruber:
@claude In @app/src/server/api/routers/shrine.ts around lines 795 - 811, The code
re-queries the sector into targetSector even though shrineBattle already
contains sector info; remove the ctx.drizzle.query.sector.findFirst block and
instead update the initial shrineBattle query (the one that produces
shrineBattle) to include the village relation (with: { village: true }) so you
can read AI defenders from
shrineBattle.sector.village.shrineSettings.activeAiIds and slice up to
SHRINE_BATTLE_MAX_USERS_PER_SIDE, keeping the same null/length checks and
returning errorResponse("No defenders available (AI defenders not configured)")
if still empty.

MathiasGruber:
@theeneon I've yet to review the code thoroughly, there are a few weird things in there I need to validate, but does this cover the intended functionality & work as expected on the latest preview deployment?

theeneon:
@MathiasGruber So far so good, I'm waiting on phros to test out the MPVP functionality better. 

Also, can localized notifications be sent so that people are aware when their sector is under attack?

MathiasGruber:
@claude can you add a feature such that when a sector owned by a village/faction is attacked, all its members are sent notifications that the sector in question is under attack

MathiasGruber:
@claude please review and implement the following fixes:

Issue 1: On the team tab, I believe we should still show the war card in question, letting the user know the current state of the sector war. Maybe it's even so that the ShrineBattleLobby should be within the WarCard, given that it's relating to a specific war.

Issue 2: 
Critical: double-start race can spawn multiple combat battles (orphaned battles) because the “claim” happens after initiateBattle().

Two attackers can call initiateShrineBattle concurrently:

both see battleId IS NULL,
both call initiateBattle(...) (creating two combat battles),
only one wins the subsequent UPDATE mpvpBattleQueue ... WHERE battleId IS NULL.
You need an atomic “claim/lock” step before creating the combat battle (then only the winner proceeds). One minimal pattern is: update the queue row first to a “locked/starting” state (or a separate lockId/startedAt column), check rowsAffected === 1, then call initiateBattle, then persist the real battleId.

Also: the AI fallback currently errors if the village has no configured activeAiIds, but issue https://github.com/studie-tech/TheNinjaRPG/issues/659 describes a fallback-to-AI behavior (ideally always possible, not only when configured). Consider selecting default shrine AIs (userData.isAi && userData.inShrines) if none are configured.

Issue 3:
leaveShrineBattle: race between “battle not started” check and deletes can corrupt queue/battle state.

You read shrineBattle.battleId and then delete mpvpBattleUser + set userData.status to AWAKE. If the battle starts after the read (or another request initiates), a user can “leave” a started battle queue and get force-set to AWAKE.

Mitigation: add DB-guarded writes (e.g., only delete the mpvpBattleUser row if the parent mpvpBattleQueue.battleId IS NULL, and only set userData.status back to AWAKE if the delete actually removed a row).

Issue 4:
In @app/src/server/api/routers/shrine.ts around lines 569 - 677, In
joinShrineBattle, narrow the catch around the ctx.drizzle.insert(mpvpBattleUser)
to only handle unique/constraint violations (slot-taken) and avoid
unconditionally reverting the user's status; on that specific error, verify the
user record is still status "QUEUED" (query userData by user.userId) and confirm
there is no mpvpBattleUser row for that user in this shrineBattleId before
issuing the update to set status back to "AWAKE"; for other errors rethrow or
return a generic DB error without changing status.

Issue 5:
In @app/src/server/api/routers/shrine.ts around lines 471 - 567, The code sets
userData.status = "QUEUED" before inserting mpvpBattleQueue/mpvpBattleUser and
lacks rollback, and it sends per-user hard-coded notifications; fix by making
the status update and both inserts atomic (use a DB transaction via
ctx.drizzle.transaction or wrap in try/catch and revert userData.status back to
"AWAKE" on any insert failure, or perform the update as the last step after
successful inserts), and change notifications to use a village channel or batch
trigger (e.g., channel "village-{targetSector.villageId}") sending a
localization key plus params instead of raw English text to avoid per-member
calls and support i18n.

Issue 6: In getUserQueuedShrineBattle  if mpvpBattleUser can ever contain >1 row per userId (bugs/races/manual DB ops), this can return an arbitrary battle/sector. Consider enforcing uniqueness at DB level (one queue entry per user) and/or ordering explicitly (e.g., newest first) plus filtering to “active” (battleId null) if that’s the intent.

Issue 7:
In @app/src/server/api/routers/shrine.ts around lines 408 - 443, The
getShrineBattles procedure currently orders by newest-first and only filters out
empty queues in-memory, leaving “ghost” mpvpBattleQueue rows in the DB; change
orderBy from desc(mpvpBattleQueue.createdAt) to asc(mpvpBattleQueue.createdAt)
to implement FIFO, and modify the DB query to only return mpvpBattleQueue rows
that have at least one related queue entry (use an EXISTS or count>0
subquery/join instead of post-filtering). Additionally, add a cleanup step using
ctx.drizzle.delete on mpvpBattleQueue with the same WHERE conditions (battleType
= "SHRINE_BATTLE", sector = input.sectorNumber, battleId IS NULL) plus a
condition that no related queue rows exist, so empty queues are removed from the
database rather than just filtered out in getShrineBattles.

Issue 8:
I currently have a mpvpQueue without any participants. If I join that battle, and then leave it, I'd expect it to get deleted (no participants). This doesn't happen though.

MathiasGruber:
@claude now getting the error on the shrine page:

Rendered more hooks than during the previous render.

Error: Rendered more hooks than during the previous render.
    at updateWorkInProgressHook (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:4766:23)
    at updateEffectImpl (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:5329:20)
    at Object.useEffect (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:15233:13)
    at exports.useEffect (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_16962a53._.js:1723:36)
    at useClearResetErrorBoundary (http://localhost:3000/_next/static/chunks/node_modules_30bf43ff._.js:13489:175)
    at useBaseQuery (http://localhost:3000/_next/static/chunks/node_modules_30bf43ff._.js:13604:224)
    at useQuery (http://localhost:3000/_next/static/chunks/node_modules_30bf43ff._.js:13897:211)
    at Object.useQuery$1 [as useQuery] (http://localhost:3000/_next/static/chunks/node_modules_30bf43ff._.js:12857:213)
    at http://localhost:3000/_next/static/chunks/node_modules_30bf43ff._.js:12343:30
    at Object.apply (http://localhost:3000/_next/static/chunks/node_modules_%40trpc_server_dist_be1055dd._.js:434:20)
    at Shrine (http://localhost:3000/_next/static/chunks/src_7cd7c92d._.js:1850:209)
    at Object.react_stack_bottom_frame (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:14827:24)
    at renderWithHooks (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:4652:24)
    at updateFunctionComponent (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:6113:21)
    at beginWork (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:6709:24)
    at runWithFiberInDEV (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:966:74)
    at performUnitOfWork (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:9563:97)
    at workLoopSync (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:9457:40)
    at renderRootSync (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:9441:13)
    at performWorkOnRoot (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:9106:47)
    at performSyncWorkOnRoot (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:10239:9)
    at flushSyncWorkAcrossRoots_impl (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:10155:316)
    at processRootScheduleInMicrotask (http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:10176:106)
    at http://localhost:3000/_next/static/chunks/node_modules_next_dist_compiled_react-dom_1e674e59._.js:10250:158
[ToS](https://app.termly.io/document/terms-of-service/71d95c2f-d6eb-4e3c-b480-9f0b9bb87830) - [Privacy](https://app.termly.io/document/privacy-policy/9fea0bba-1061-47c0-8f28-0f724f06cc0e) - Cookie [Policy](https://app.termly.io/document/cookie-policy/971fe8a9-3613-41a0-86ad-8e08e7be93d7) / [Consent](http://localhost:3000/consent) - [Rules](http://localhost:3000/rules) - [Staff](http://localhost:3000/manual/staff)


Please review all frontend component for such render issues. And add to the Claude.md that frontent UI should always be checked for these types of errors

MathiasGruber:
@claude

Issue 1: For the WarCard, please add the same border around on the solo battle as the team battle, to align the two tabs visuals.

Issue 2: In getShrineBattles, I believe we can do one query to fetch all the battles, then based on these figure out which should be removed (which we do not have to await), and those can be filtered off before returning. That will make the endpoint faster.

Issue 3: Comments have been left in the code relating to e.g. "issue X: something".. I don't want comments left like this, ever. Please remove those present (of course, leave the comment if it makes sense, but remove mention of issue number), and add to claude.md a section about not leaving un-natural (wrt. how we usually comment) comments in the code.

Issue 4: When sending the message about the sector being attacked, we should send it on a villageId channel (i.e. to all villagers of that villageId). If we don't have such a channel yet, we should establish it. Right now it's being sent on the sector, which is not what we want.

Issue 5:
In @app/src/server/api/routers/shrine.ts around lines 520 - 613, The
challengeShrine mutation can leave partial state if one insert succeeds and the
other fails; change the insert logic to be atomic: either run both inserts
inside a single DB transaction (using ctx.drizzle.transaction or equivalent) or
sequence them and perform compensating deletes on failure. Specifically, in
challengeShrine, create the mpvpBattleQueue first (mpvpBattleQueue.insert), then
insert the mpvpBattleUser (mpvpBattleUser.insert); if the second insert fails,
delete the inserted mpvpBattleQueue row and restore user status
(userData.update) before rethrowing the error; alternatively wrap both inserts
in ctx.drizzle.transaction so both succeed or both are rolled back.

Issue 6:
Join flow can leave users “QUEUED but not in any queue” on non-duplicate insert errors
You update userData.status to QUEUED before inserting mpvpBattleUser. If the insert fails for reasons other than duplicate-slot, you currently rethrow and do not revert status—this can strand the user in QUEUED with no queue entry.

Recommend: on any insert failure after the status update, do the same “verify no queue entry exists, then revert to AWAKE” rollback (not just for duplicate errors).

Issue 7:
Critical: raw SQL references MpvpBattleQueue (case-sensitive table name risk)
Line 779 references MpvpBattleQueue directly. On many MySQL installs (notably Linux), table name casing can be case-sensitive—this can make the EXISTS subquery fail at runtime.

Prefer expressing this with Drizzle (or at least ensure the identifier matches the actual table name exactly).

Safer approach: use an `IN (subquery)` guard (no raw table name)
-const deleteResult = await ctx.drizzle
-  .delete(mpvpBattleUser)
-  .where(
-    and(
-      eq(mpvpBattleUser.clanBattleId, input.shrineBattleId),
-      eq(mpvpBattleUser.userId, user.userId),
-      sql`EXISTS (
-        SELECT 1 FROM MpvpBattleQueue
-        WHERE MpvpBattleQueue.id = ${input.shrineBattleId}
-        AND MpvpBattleQueue.battleId IS NULL
-      )`,
-    ),
-  );
+const activeQueueSubquery = ctx.drizzle
+  .select({ id: mpvpBattleQueue.id })
+  .from(mpvpBattleQueue)
+  .where(
+    and(eq(mpvpBattleQueue.id, input.shrineBattleId), isNull(mpvpBattleQueue.battleId)),
+  );
+
+const deleteResult = await ctx.drizzle
+  .delete(mpvpBattleUser)
+  .where(
+    and(
+      eq(mpvpBattleUser.userId, user.userId),
+      inArray(mpvpBattleUser.clanBattleId, activeQueueSubquery),
+    ),
+  );

Issue 8: raised by coderabbit
Concern 1 is valid: claiming-* state can permanently wedge the queue with no recovery mechanism

If the server crashes after setting battleId = claiming-* but before updating to the real battleId, the queue is stuck. The leaveShrineBattle endpoint checks battleId IS NULL, so users cannot leave a queue with a claiming-* placeholder. No background cleanup job was found to recover from stale claims. Consider adding either:

A TTL/expiry on queue entries with automatic cleanup
A recovery mechanism to detect and clear claiming-* entries after a timeout
Explicit cleanup job for stale claims
Concern 2 is valid: Setting status="AWAKE" after initiateBattle() undoes the correct status set by that function

The initiateBattle() function uses CASE WHEN isAi = false THEN "BATTLE" ELSE "AWAKE" END, setting human players to status="BATTLE" and AI defenders to status="AWAKE". However, initiateShrineBattle then immediately overwrites this with status="AWAKE" for users matching status="QUEUED". This resets human attackers back to AWAKE when they should remain in BATTLE status to properly enter combat. Remove the post-init status update or update only the players who weren't already transitioned to BATTLE by initiateBattle().

theeneon:
@MathiasGruber Should i add the other systems from 4.1 in git hub now?

MathiasGruber:
> @MathiasGruber Should i add the other systems from 4.1 in git hub now?

Please do.. preferably chunked into smaller collected issues, and then you can tag "claude" and ask something along these line: "please create a thorough implementation plan for this feature, and post the full plan into this issue. Make sure to ask any follow-up questions wrt. to any features or implementations that are not clear am ambiguous"

That will let it go over the codebase and come up with an implementation plan, as well as find any gaps in understanding. Once the implementation plan from claude is approved by you guys, I can take over, have claude implement it, and review it for security/performance/logic issues.

MathiasGruber:
@claude Look into this possible optimization: app/src/server/api/routers/shrine.ts (1)
464-500: LGTM - Correct active queue detection with minor optimization opportunity.

The endpoint correctly:

Filters to active shrine battles (battleId === null)
Returns sector information for UI validation
Handles the case of no active queue
Optional: Add database-level filtering to avoid fetching completed battles:

⚡ Optional optimization
     const queueEntries = await ctx.drizzle.query.mpvpBattleUser.findMany({
-      where: eq(mpvpBattleUser.userId, ctx.userId),
+      where: eq(mpvpBattleUser.userId, ctx.userId),
       with: {
         clanBattle: {
+          where: and(
+            eq(mpvpBattleQueue.battleType, "SHRINE_BATTLE"),
+            isNull(mpvpBattleQueue.battleId),
+          ),
           columns: {
However, the current in-memory filtering works correctly for typical use cases.

MathiasGruber:
@claude if I try to start a team sector battle with two attackers, but nobody signed up to defend, I'm told "No defenders available (no AI defenders found)". In this case, it should fill up the defenders with the same AIs as we fill into the solo battle.

Phrosfire:
When the attacker attacks the shrine, there is no notifications going out to the village whose shrine is being attacked.  When a defending villager goes to the shrine, they also cannot see that there is anyone grouped for the raid and there is no option to join as a defender.  This is the view you get when there is a raid forming on the defending villages shrine.
image

Phrosfire:
We no longer get the defender not available error, but the fight fails to start.  Looking at the MpvpBattleQueue table, we can see that the row for that shrines queue is created.  Viewing the MpvpBattleUser table, we can see the list of users who joined that queue as an attacker.  This is the response when trying to start the battle

image

MathiasGruber:
> When the attacker attacks the shrine, there is no notifications going out to the village whose shrine is being attacked. When a defending villager goes to the shrine, they also cannot see that there is anyone grouped for the raid and there is no option to join as a defender. This is the view you get when there is a raid forming on the defending villages shrine. 

Are you sure? I'm trying on sector 202, and it looks good as defender to me. I'm pushing the fix to start battle now, I encountered that about half an hour ago as well. Should work in next preview.

image

theeneon:
@MathiasGruber I think i need permissions to tag claude. I just tried and it failed

MathiasGruber:
Added a few small final tweaks and fixes, and also did some testing just with two users, which seemed to work well. I've gone over the code, and I'm happy with the performance / security of it as well.. from my point of view ready for merge, although before doing so need to do DB migrations on prod, otherwise we break things. I'll wait for approval from you guys first though @theeneon and @Phrosfire

theeneon:
@MathiasGruber 

- [x] So found an issue You should not be able to attack the Shrine that is located in the villages home sector or reserved sectors. 

- [x] Also the defender view is only showing for Sectors wars but not when someone has attacked the sector to claim it while during normal wars. 

- [x] While a village has a Sector war active, you can't declare a village war. 
- [x] Notifications only show up for Sector Wars not when the Shrine is attacked. 
- [x] You can also create a Shrine attack party regardless if your Kage has declared a sector war or if you're at war with that village. 

- [x] On a side note for the notifications can it be like one of these:
image

MathiasGruber:
Resolved all of them (as far as I can tell from my testing) locally.. I'll merge, and then I think any left-over inconsistencies we can fix based on user feedback.

PR #1010, opened 2026-01-29
fix: make friendly fire settings work in raid battles

### LINKED ISSUE #1007: Friendly Fire Settings Don't work in Raids
When in a Raid, friendly fire settings don't work as you're taking damage from your own Jutsu and you're jutsus are able to hit your teammates.

MathiasGruber:
## Summary
- Fixed friendly fire settings not working in raid battles
- Uses `direction` property to determine team membership instead of `villageId`
- Updated targeting logic (OPPONENT/ALLY) for RAID battles

Closes #1007

---

> [!NOTE]
> **Medium Risk**
> Changes core combat targeting/friendly-fire logic and team assignment, which could affect multiple battle types if `direction` is missing or mis-set; scope is localized but behavior-impacting.
> 
> **Overview**
> Fixes RAID battles where friendly-fire rules were evaluated incorrectly by switching team detection from `villageId` (and intra-village heuristics) to `direction` (left/right) in `checkFriendlyFire`.
> 
> Updates targeting validation so `OPPONENT`/`ALLY` checks use `direction` for move/target legality, and ensures `processUsersForBattle` assigns `direction` from `leftSideUserIds` while combat state updates preserve a user’s original `direction` (and initiative) across loadout hydration. Also removes an unused import in the combat router.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0c010a28a3979134844bf1a0fc58c7ae051412e6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **Refactor**
  * Combat system now determines teams by player direction (left/right) instead of village affiliation.
  * Friendly-fire detection and move validation updated to use direction-based comparisons.
  * Target selection behavior on shared tiles adjusted to use direction for opponent/ally resolution.
  * Battle processing now preserves and restores players' direction alongside initiative values.

### DIFF
--- a/app/src/libs/combat/actions.ts
+++ b/app/src/libs/combat/actions.ts
@@ -966,7 +966,7 @@ export const getTargetUser = (
     if (target === "SELF") {
       result = users.find((u) => u.userId === user.userId && u.hex === tile);
     } else if (target === "OPPONENT") {
-      result = users.find((u) => u.villageId !== user.villageId && u.hex === tile);
+      result = users.find((u) => u.direction !== user.direction && u.hex === tile);
     } else if (target === "ALLY") {
       result = users.find((u) => u.villageId === user.villageId && u.hex === tile);
     } else if (target === "OTHER_USER") {
--- a/app/src/libs/combat/process.ts
+++ b/app/src/libs/combat/process.ts
@@ -53,6 +53,7 @@ import type { BattleUserState } from "./types";
 import type { GroundEffect, UserEffect, ActionEffect, BattleEffect } from "./types";
 import type { CompleteBattle, Consequence, CombatAction } from "./types";
 import type { ShieldTagType } from "@/validators/combat";
+
 /**
  * Minimal user type for checkFriendlyFire
  */
@@ -60,11 +61,14 @@ type FriendlyFireUser = {
   userId: string;
   isSummon?: boolean;
   controllerId: string;
-  villageId: string | null;
+  direction: "left" | "right";
 };
 
 /**
- * Check whether to apply given effect to a user, based on friendly fire settings
+ * Check whether to apply given effect to a user, based on friendly fire settings.
+ * Uses the 'direction' property to determine teams - users on the same side (left/right)
+ * are allies, users on opposite sides are enemies. This works for all battle types
+ * because direction is set based on userIds (attackers=left) vs targetIds (defenders=right).
  */
 export const checkFriendlyFire = (
   effect: BattleEffect,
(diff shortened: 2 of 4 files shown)


theeneon:
@MathiasGruber Not sure if you saw my message most Jutsus in the game use targetting other_user because Opponent would break in Spars/Clan/Kage Battles anything that is happening with players from the same village while it is objectively the correct thing to use. 

Opponet would fail in these cases. Does this also resolve that issue?

MathiasGruber:
@theeneon This would likely break it.. can you test on preview deployment? I think this new evaluation of friendly fire is a lot more intuitive from a code perspective. Adapted the AI solution a bit

theeneon:
> @theeneon This would likely break it.. can you test on preview deployment? I think this new evaluation of friendly fire is a lot more intuitive from a code perspective. Adapted the AI solution a bit

Yea when fighting someone of the same village, the opponent targetting doesn't work. 

A solution that you might can use here is auto assign whoever you're fighting the freedom state village ID so no matter what it would always count as an enemy seeing as we used village ID to decide who's ally vs enemy. 
but that was a roundabout way of doing it

MathiasGruber:
> > @theeneon This would likely break it.. can you test on preview deployment? I think this new evaluation of friendly fire is a lot more intuitive from a code perspective. Adapted the AI solution a bit
> 
> Yea when fighting someone of the same village, the opponent targetting doesn't work.
> 
> A solution that you might can use here is auto assign whoever you're fighting the freedom state village ID so no matter what it would always count as an enemy seeing as we used village ID to decide who's ally vs enemy. but that was a roundabout way of doing it

This new solution on the preview domain shouldn't be using village at all to determine friendly fire. Instead if uses the "direction"; i.e. who attacked who.. so setting village wouldn't change anything. Can you explain to me exactly what currently fails in this PR preview, I think I'm not following completely.

theeneon:
> > > @theeneon This would likely break it.. can you test on preview deployment? I think this new evaluation of friendly fire is a lot more intuitive from a code perspective. Adapted the AI solution a bit
> > 
> > 
> > Yea when fighting someone of the same village, the opponent targetting doesn't work.
> > A solution that you might can use here is auto assign whoever you're fighting the freedom state village ID so no matter what it would always count as an enemy seeing as we used village ID to decide who's ally vs enemy. but that was a roundabout way of doing it
> 
> This new solution on the preview domain shouldn't be using village at all to determine friendly fire. Instead if uses the "direction"; i.e. who attacked who.. so setting village wouldn't change anything. Can you explain to me exactly what currently fails in this PR preview, I think I'm not following completely.

Well I entered the battle arena, set the AI i was fighting to Current as am in Current.
Changed a Jutsu targetting to Opponent and when in the Arena the jutsu did no damage to the AI. 
I would say the assign to left probably only works in Raids.

PR #1018, opened 2026-01-29
fix: normalize tag case in getEfficiencyRatio for damage modifiers

### LINKED ISSUE #1017: Damage Increase and Damage Reduction are Incorrect
During tests with Damage Increase Given/Taken:

Expected multiplier math did not match actual damage output

Example:

253.36 × 1.9 = 481.38 expected

Actual result: 725.38

This revealed a massive hidden damage boost (~245–286%)

The samething is happening with damage reduction where a test was conducted using 70% Damage reduction 

where we got 59.7% reduction instead of 70% 
1145.51 -> 461.61 is a 59.7% reduction

There's a 10.3% reduction that is missing.

theeneon:
## Summary

- Fixes case sensitivity bug in `getEfficiencyRatio()` that caused damage modifiers to not apply correctly
- Normalizes all tag comparisons to lowercase

## Test plan

- [ ] Test 70% damage reduction - should now reduce damage by exactly 70%
- [ ] Test 90% damage increase - should increase damage by exactly 90%

Fixes #1017

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed inconsistency in damage modifier tag matching that could prevent certain combat effects from being properly applied during battles.

### DIFF
--- a/app/src/libs/combat/tags.ts
+++ b/app/src/libs/combat/tags.ts
@@ -2965,8 +2965,8 @@ const getEfficiencyRatio = (dmgEffect: UserEffect, effect: UserEffect) => {
       e.statTypes?.forEach((statType) =>
         tags.push(
           statType === "Highest" && e.highestOffence
-            ? getStatTypeFromStat(e.highestOffence)
-            : statType,
+            ? getStatTypeFromStat(e.highestOffence).toLowerCase()
+            : statType.toLowerCase(),
         ),
       );
     }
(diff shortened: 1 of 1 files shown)


theeneon:
@MathiasGruber The PR was implemented even though I didn't ask for it. Should remember to tell it to not implement PRs.

Älso I do believe that this entire PR is probably very very wrong. 

- Water Spirit: Undyne takes 673.85 damage 

After applying a 35% buff 

- Water Spirit: Undyne takes 842.32 damage


So as we can see that is incorrect. As 35% increase should have given us 909.70 

So some where in the effect processing we have a bug.

MathiasGruber:
claude is a bit difficult to control sometimes i guess. I'll abandon this PR and look at the issue once I get to it :)

PR #1254, opened 2026-03-15
Fix issue# 1141: Reset skill tree in batches

Phrosfire:
# Pull Request

When an admin attempts to reset all skill trees on the live database, they are met with "vttablet: rpc error: code = Unavailable desc = error reading from server: EOF".  In order to mitigate timeouts, I adjusted the code so it would do the deletes in batches rather than trying to delete everything at once.

This issue does not occur on the dev database, so I'm thinking the issue is a timeout due to the number of deletes.

## License

## Summary by CodeRabbit

* **Refactor**
  * Optimized skill tree reset to use batched operations for more reliable and efficient clearing.
  * Kept audit logging for mass resets; message updated to reflect batched process.

* **Bug Fixes**
  * Improved error handling and messages: explicitly reports when nothing was cleared and preserves failure reporting when some entries remain.
  * Success message now shows the exact number of entries cleared.

### DIFF
--- a/app/src/server/api/routers/skillTree.ts
+++ b/app/src/server/api/routers/skillTree.ts
@@ -476,6 +476,15 @@ export const skillTreeRouter = createTRPCRouter({
       }
 
       if (hadUndeletedRows) {
+        await ctx.drizzle.insert(actionLog).values({
+          id: nanoid(),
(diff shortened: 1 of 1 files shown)


MathiasGruber (approved):
LGTM

PR #1285, opened 2026-03-25
Feat/raid queue unstuck

ifroshty:
## What was implemented

Added a **"Ready to Queue"** gate that users must click before joining a raid queue. This button runs a self-service unstuck routine server-side before allowing the user to create or join a raid team.

### What the button does
- Resets user status to `AWAKE` and clears `battleId` / `travelFinishAt`
- Removes the user from any stuck `mpvpBattleUser` or `rankedPvpQueue` entries
- Cancels any pending spar requests the user sent
- Rejects any pending spar requests the user received (prevents being dragged into a spar while queued)
- Cleans up orphaned/stuck-claiming raid teams left behind

### Additional fixes included
- `getUserRaidQueue` now auto-detects if the raid expired while the user was queued and silently cleans up all affected users on page load
- `checkAndCleanupExpiredRaid` now also clears queue entries and resets user statuses for everyone queued on that raid when cleanup is triggered

## Why

Users were getting permanently stuck in the raid queue with `QUEUED` status after disconnecting, navigating away, or encountering edge cases during battle initialization. Previously the only fix was an admin `forceAwake`. This gives users a self-service escape hatch before every queue attempt.

## Guardrails
- Rate limited via `ratelimitMiddleware`
- Blocked if user is `HOSPITALIZED` — hospital is a game penalty, not a stuck state
- Spar requests are cancelled/rejected atomically in the same `Promise.all` as the status reset

## Breaking changes
None.

## License

## Summary by CodeRabbit

* **New Features**
  * Added a "Ready to Queue" button in the Raid Browser to let users reset and rejoin raid queues.

* **Bug Fixes**
  * Improved detection and cleanup of expired or completed raids to prevent stale queue states and reduce stuck sessions.
  * Safer cleanup logic to avoid unintended removal of active raid data.

* **Chores**
  * Appended new migration journal entries.

### DIFF
--- a/app/src/layout/RaidBrowser.tsx
+++ b/app/src/layout/RaidBrowser.tsx
@@ -57,6 +57,7 @@ const RaidBrowser: React.FC<RaidBrowserProps> = (props) => {
   // State
   const [selectedRaidId, setSelectedRaidId] = useState<string | null>(null);
   const [viewMode, setViewMode] = useState<"active" | "history">("active");
+  const [isReadyToQueue, setIsReadyToQueue] = useState(false);
   const { data: userData, pusher } = useRequiredUserData();
 
   // Queries
@@ -110,6 +111,7 @@ const RaidBrowser: React.FC<RaidBrowserProps> = (props) => {
     api.raids.leaveRaidQueue.useMutation({
       onSuccess: (data) => {
         showMutationToast(data);
+        setIsReadyToQueue(false);
         void util.raids.getUserRaidQueue.invalidate();
         void util.raids.getActiveRaidTeams.invalidate();
         void util.profile.getUser.invalidate();
@@ -139,9 +141,23 @@ const RaidBrowser: React.FC<RaidBrowserProps> = (props) => {
       },
     });
 
+  const { mutate: readyToQueue, isPending: readyPending } =
+    api.raids.readyToQueue.useMutation({
+      onSuccess: (data) => {
+        if (data.success) {
+          setIsReadyToQueue(true);
+          void util.raids.getUserRaidQueue.invalidate();
+          void util.raids.getActiveRaidTeams.invalidate();
+          void util.profile.getUser.invalidate();
+        }
+        showMutationToast(data);
+      },
+    });
+
   // Event handlers
   const handleRaidSelect = (raidId: string) => {
     setSelectedRaidId((prev) => (prev === raidId ? null : raidId));
+    setIsReadyToQueue(false);
   };
 
   const handleJoinQueue = (teamId?: string) => {
@@ -242,10 +258,11 @@ const RaidBrowser: React.FC<RaidBrowserProps> = (props) => {
   const hideSelectorCard =
     sectorFilter !== undefined && availableRaids.length === 1 && viewMode === "active";
 
-  // Handle view mode switch - clear selection
+  // Handle view mode switch - clear selection and ready state
   const handleViewModeChange = (mode: "active" | "history") => {
     setViewMode(mode);
     setSelectedRaidId(null);
+    setIsReadyToQueue(false);
   };
 
   return (
@@ -525,7 +542,7 @@ const RaidBrowser: React.FC<RaidBrowserProps> = (props) => {
                     </p>
                   </div>
                 </div>
-              ) : (
+              ) : userInSelectedRaidQueue || isReadyToQueue ? (
                 <RaidTeamsDisplay
                   activeTeams={activeTeams}
                   maxTeams={maxTeams}
@@ -540,6 +557,17 @@ const RaidBrowser: React.FC<RaidBrowserProps> = (props) => {
                   isLeaving={leavePending}
                   isStarting={startPending}
                 />
+              ) : (
+                <div className="rounded-lg border bg-card p-4 text-center">
+                  <p className="mb-3 text-muted-foreground text-sm">
+                    Click the button below to prepare for battle. This will clear any
+                    stuck status before joining the queue.
+                  </p>
+                  <Button onClick={() => readyToQueue()} disabled={readyPending}>
+                    <Swords className="mr-2 h-4 w-4" />
+                    {readyPending ? "Preparing..." : "Ready to Queue"}
+                  </Button>
+                </div>
               ))}
 
             {/* Reward Thresholds */}
--- a/app/src/server/api/routers/raids.ts
+++ b/app/src/server/api/routers/raids.ts
@@ -1,4 +1,17 @@
-import { and, desc, eq, gte, inArray, isNull, lt, lte, sql } from "drizzle-orm";
+import {
+  and,
+  desc,
+  eq,
+  gte,
+  inArray,
+  isNull,
+  like,
+  lt,
+  lte,
+  notExists,
+  or,
+  sql,
+} from "drizzle-orm";
 import { nanoid } from "nanoid";
 import { z } from "zod";
 import {
@@ -17,9 +30,11 @@ import {
   quest,
   raidDamageThreshold,
   raidParticipation,
+  rankedPvpQueue,
   sector,
   userData,
   userRaidBuff,
+  userRequest,
   war,
 } from "@/drizzle/schema";
 import { getServerPusher, updateRaidTeamsOnSector } from "@/libs/pusher";
@@ -109,7 +124,10 @@ export const raidsRouter = createTRPCRouter({
    */
   getAvailableRaids: protectedProcedure
     .meta({
-      mcp: { enabled: true, description: "Get available raids for current user" },
+      mcp: {
+        enabled: true,
+        description: "Get available raids for current user",
+      },
     })
     .input(z.object({ sector: z.number().optional() }).optional())
     .query(async ({ ctx, input }) => {
@@ -264,7 +282,9 @@ export const raidsRouter = createTRPCRouter({
    * Get detailed info about a specific raid
    */
   getRaidDetails: protectedProcedure
-    .meta({ mcp: { enabled: true, description: "Get details for a specific raid" } })
+    .meta({
+      mcp: { enabled: true, description: "Get details for a specific raid" },
+    })
     .input(z.object({ questId: z.string() }))
     .query(async ({ ctx, input }) => {
       // Query - parallel fetch
@@ -313,7 +333,9 @@ export const raidsRouter = createTRPCRouter({
    * Get the raid leaderboard
    */
   getRaidLeaderboard: protectedProcedure
-    .meta({ mcp: { enabled: true, description: "Get raid damage leaderboard" } })
+    .meta({
+      mcp: { enabled: true, description: "Get raid damage leaderboard" },
+    })
     .input(
       z.object({
         questId: z.string(),
@@ -364,7 +386,10 @@ export const raidsRouter = createTRPCRouter({
    */
   getUserRaidQueue: protectedProcedure
     .meta({
-      mcp: { enabled: true, description: "Get user's current raid queue status" },
+      mcp: {
+        enabled: true,
+        description: "Get user's current raid queue status",
+      },
     })
     .query(async ({ ctx }) => {
       // Query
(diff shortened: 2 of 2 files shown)


app/src/server/api/routers/raids.ts
```diff
@@ -392,6 +407,25 @@ export const raidsRouter = createTRPCRouter({
...
+      const raidExpired =
+        (raid?.raidEndsAt && raid.raidEndsAt < now) ||
+        (raid?.raidBossCurrentHealth ?? 1) <= 0;
+      if (raidExpired) {
+        await checkAndCleanupExpiredRaid(
+          ctx.drizzle,
+          queue.clanBattle.attackerEntityId,
+        );
+        return { inQueue: false, queue: null, isClaiming: false };
+      }
```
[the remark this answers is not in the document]
ifroshty: This pattern is intentional and pre-dates this PR. `getUserRaidQueue` is a user-polling endpoint  placing the cleanup here is the only reliable way to un-stick a user who is actively waiting on their own queue status. The "write storm" concern is mitigated because `checkAndCleanupExpiredRaid` exits early on its first query if the raid isn't expired; by the time a second concurrent caller reaches the write path, the first has already cleaned up and the early-exit fires. Promoting this to a client-triggered mutation would add round-trips and a separate failure mode (client never calls it). Out of scope for this PR.

app/src/server/api/routers/raids.ts
```diff
@@ -1407,27 +1572,72 @@ export const checkAndCleanupExpiredRaid = async (
...
+  if (timeExpired && !bossDefeated) {
+    const objective = raid.content?.objectives?.[0];
+    if (objective?.task === "exclusive_raid") {
+      const raidSector = (objective as RaidObjectiveType).sector;
+      if (raidSector !== null && raidSector !== undefined) {
+        // Delete sector - returns it to neutral ownership (syndicate)
+        const result = await client.delete(sector).where(eq(sector.sector, raidSector));
+        return { wasExpired: true, sectorCleaned: result.rowsAffected > 0 };
+      }
     }
```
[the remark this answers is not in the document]
ifroshty: The `if (timeExpired && !bossDefeated)` sector cleanup gate was not touched by this PR, it existed before and is unchanged. This PR only extended the raid fetch predicate to also match boss-defeated raids so their queued teams get cleaned up. Whether an exclusive sector should be reclaimed when a boss is killed simultaneously as the timer expires is a game design question that should be addressed in a dedicated PR with input from game design. No regression introduced here.

app/src/server/api/routers/raids.ts
```diff
@@ -633,6 +653,84 @@ export const raidsRouter = createTRPCRouter({
...
+        fetchUser(ctx.drizzle, ctx.userId),
+        ctx.drizzle.query.mpvpBattleUser.findFirst({
+          where: eq(mpvpBattleUser.userId, ctx.userId),
+        }),
+      ]);
+      // Guard
+      if (!user) return errorResponse("User not found");
+      if (user.status === "HOSPITALIZED") {
+        return errorResponse("You cannot ready up while hospitalized");
+      }
```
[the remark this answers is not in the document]
ifroshty: already fixed in the current PR head. `readyToQueue` now blocks `user.status === "BATTLE"` in [raids.ts]

app/src/layout/RaidBrowser.tsx
```diff
@@ -139,9 +140,22 @@ const RaidBrowser: React.FC<RaidBrowserProps> = (props) => {
...
+        }
+        showMutationToast(data);
+      },
+    });
+
   // Event handlers
   const handleRaidSelect = (raidId: string) => {
     setSelectedRaidId((prev) => (prev === raidId ? null : raidId));
+    setIsReadyToQueue(false);
   };
```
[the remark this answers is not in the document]
ifroshty: already fixed in the current PR head. `handleViewModeChange` now resets `isReadyToQueue` in [RaidBrowser.tsx]

app/src/server/api/routers/raids.ts
```diff
@@ -1407,27 +1572,84 @@ export const checkAndCleanupExpiredRaid = async (
...
+        ? [
+            client
+              .update(userData)
+              .set({ status: "AWAKE", battleId: null })
+              .where(
+                and(eq(userData.status, "QUEUED"), inArray(userData.userId, userIds)),
+              ),
+          ]
+        : []),
+    ]);
```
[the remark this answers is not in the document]
ifroshty: already fixed in the current PR head. `checkAndCleanupExpiredRaid` now deletes `mpvpBattleUser` using snapshot `teamIds` from `queuedTeams`, avoiding the live subquery race in [raids.ts]

app/src/server/api/routers/raids.ts
```diff
@@ -1407,27 +1548,97 @@ export const checkAndCleanupExpiredRaid = async (
...
+                  and(
+                    eq(mpvpBattleQueue.battleType, "RAID_BATTLE"),
+                    eq(mpvpBattleQueue.attackerEntityId, raidId),
+                    sql`(${mpvpBattleQueue.battleId} IS NULL OR ${mpvpBattleQueue.battleId} LIKE 'claiming-%')`,
+                  ),
+                ),
+            ),
+          ),
+        ),
+    ]);
```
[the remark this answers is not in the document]
ifroshty: Already fixed in the current PR head. `checkAndCleanupExpiredRaid` now derives snapshot `userIds` from `queuedTeams` and uses those IDs directly in the `userData` update, so the wake-up path no longer depends on a runtime subquery over rows that may already have been deleted.

app/src/server/api/routers/raids.ts
```diff
@@ -392,6 +394,24 @@ export const raidsRouter = createTRPCRouter({
...
+      const now = new Date();
+      const raidExpired =
+        (raid?.raidEndsAt && raid.raidEndsAt < now) ||
+        (raid?.raidBossCurrentHealth ?? 1) <= 0;
+      if (raidExpired) {
+        await checkAndCleanupExpiredRaid(
+          ctx.drizzle,
+          queue.clanBattle.attackerEntityId,
+        );
+        return { inQueue: false, queue: null, isClaiming: false };
```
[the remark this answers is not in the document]
ifroshty: Not taking this change in this PR. The suggested `quest` relation on `mpvpBattleQueue` is only valid for `RAID_BATTLE` rows, and we already had review feedback calling out that broad relation as misleading for other battle types. The explicit `quest.findFirst` here keeps the read path correct and scoped to the confirmed raid queue case without changing shared schema behavior.

app/drizzle/migrations/meta/_journal.json
```diff
@@ -22,6 +22,20 @@
       "when": 1771398851340,
       "tag": "0002_fat_the_spike",
       "breakpoints": false
+    },
+    {
+      "idx": 3,
+      "version": "5",
+      "when": 1774291788478,
+      "tag": "0003_brave_black_cat",
+      "breakpoints": false
```
MathiasGruber: What's up with this _journal file change?

app/src/server/api/routers/raids.ts
```diff
@@ -633,6 +667,137 @@ export const raidsRouter = createTRPCRouter({
...
+      if (user.status === "BATTLE") {
+        return errorResponse("You cannot ready up while in an active battle");
+      }
+      const userId = user.userId;
+      // CAS: only transition from the status we read to avoid racing concurrent requests.
+      // Skip CAS for already-AWAKE users — MySQL reports rowsAffected = 0 on a no-op update,
+      // which is indistinguishable from a concurrent status change.
+      if (user.status !== "AWAKE") {
+        const updateResult = await ctx.drizzle
+          .update(userData)
```
MathiasGruber, line 733: The endpoint sets battleId: null and travelFinishAt: null even for AWAKE users. If a user is AWAKE but traveling (has travelFinishAt set), calling ready-to-queue would cancel their travel. The endpoint only guards against HOSPITALIZED and BATTLE statuses. A user who is QUEUED for something else, or TRAVEL, could have their state corrupted. Consider adding guards for other non-AWAKE statuses

app/src/server/api/routers/raids.ts
```diff
@@ -633,6 +667,137 @@ export const raidsRouter = createTRPCRouter({
...
+        if (updateResult.rowsAffected === 0) {
+          return errorResponse(
+            "Failed to update status - your status may have changed",
+          );
+        }
+      } else {
+        // Already AWAKE: still clear any dirty fields without a CAS check
+        await ctx.drizzle
+          .update(userData)
+          .set({ travelFinishAt: null, battleId: null })
```
MathiasGruber, line 752: We have to think carefully that this will not further corrupt user state. I acknowledge we have an issue now, which this seeks to help users with, but I can't quite wrap my head around if there could be consequences downstream, resulting in even more messed up user states. This is why I've always opted to try to fix the root issue.. If we move forward with this, please help me with considering all the other things this could mess up

app/src/server/api/routers/raids.ts
```diff
@@ -633,6 +667,137 @@ export const raidsRouter = createTRPCRouter({
...
+          .where(eq(userData.userId, userId));
```
MathiasGruber: I think we should do `.where(and(eq(userData.userId, userId), eq(userData.status, user.status)));`here as well, in case we have race mutations running at the same time

MathiasGruber (changes requested):

app/src/server/api/routers/raids.ts
theeneon: > We have to think carefully that this will not further corrupt user state. I acknowledge we have an issue now, which this seeks to help users with, but I can't quite wrap my head around if there could be consequences downstream, resulting in even more messed up user states. This is why I've always opted to try to fix the root issue.. If we move forward with this, please help me with considering all the other things this could mess up

As a  fix I know it's a bit ugly and fixing the root cause would be best but its such a hard issue to nail down where its not easily replicable in testing but each time we do a raid, players need to have their states cleared after every fight. 

It could be because of how raids are where its a prolong battle that is causing the issues. 

Off the top of my head, this feature is only available when raids are active; it can't be used to dodge other forms of combat. There might be other cases that I've over looked I'll let the others speak on that as well.

app/drizzle/migrations/meta/_journal.json
ifroshty: These were orphaned entries left over from an earlier point in development where a `quest` relation was temporarily added to `mpvpBattleQueueRelations` in `schema.ts`. That relation was later removed since `attackerEntityId` is only a quest ID for `RAID_BATTLE` rows , advertising it as a universal relation on a polymorphic table was wrong. The migration tool was run while that relation was still present, which wrote the journal entries, but since Drizzle relations are TypeScript-only and generate no SQL, the actual `.sql` files were never created. The journal ended up with two entries pointing at non-existent files.

Removed both orphaned entries (idx 3 & 4) from `_journal.json`.

app/src/server/api/routers/raids.ts
ifroshty: I went in and double checked and found the following statuses:

| Status | Treatment | Reason |
|--------|-----------|--------|
| `HOSPITALIZED` | Blocked (existing) | Game penalty, not a stuck state |
| `BATTLE` | Blocked (existing) | Actively in a fight |
| `TRAVEL` | Blocked (existing) | Legitimately traveling |
| `ASLEEP` | **Blocked (new)** | Legitimate game state, resetting would wake them without consent |
| `KAGE_QUEUED` | **Blocked (new)** | The KAGE `userRequest` row isn't cancelled here; leaving it dangling could drag the user into a Kage battle after they've joined a raid queue |
| `QUEUED` (clan/shrine/ranked) | **Blocked (new)** | Non-raid `mpvpBattleUser` rows would be deleted, corrupting their team; ranked queue entry would be silently dropped |
| `QUEUED` (raid) | Allowed | The exact stuck state this endpoint is meant to fix |
| `AWAKE` + active travel | **Blocked (new)** | Guards against `travelFinishAt` being in the future even when status reads `AWAKE` |

Also tightened the `mpvpBattleUser` query to fetch `battleType` from `clanBattle`, so the `QUEUED` guard can distinguish a stuck raid queue entry from a legitimate clan/shrine queue entry before touching anything.

What the endpoint does **not** touch (and never did): KAGE/SPAR userRequests beyond SPAR cancellation, clan/shrine `mpvpBattleQueue` records, hospital timers, or any other game state outside of the user's own status, `travelFinishAt`, `battleId`, their specific `mpvpBattleUser` row, and `rankedPvpQueue`.

Worth noting as well: to reach this button the user has to travel to the raid sector, open the raid browser, and actively click "Ready to Queue", so there's an inherent layer of intention before any state is touched.

app/src/server/api/routers/raids.ts
ifroshty: Made the change. good call

app/src/server/api/routers/raids.ts
ifroshty: Fixed. Added two guards for this:

1. **`TRAVEL` status blocked** so users actively traveling are now rejected with "You cannot ready up while traveling"

2. **AWAKE + future `travelFinishAt` blocked** so users whose status is `AWAKE` but still have `travelFinishAt` set in the future (edge case where status and travel state diverged), the endpoint now checks `travelFinishAt > now` and returns an error rather than silently clearing it

Also added guards for `ASLEEP`, `KAGE_QUEUED`, and `QUEUED` users not in a raid queue. Please see the reply on the [thread below ](https://github.com/studie-tech/TheNinjaRPG/pull/1285#discussion_r3005803592)for the full breakdown of all 7 statuses.

app/src/server/api/routers/raids.ts
```diff
@@ -392,6 +407,22 @@ export const raidsRouter = createTRPCRouter({
...
+        columns: { raidEndsAt: true, raidBossCurrentHealth: true },
+      });
+      const now = new Date();
+      const raidExpired =
+        (raid?.raidEndsAt && raid.raidEndsAt < now) ||
+        (raid?.raidBossCurrentHealth ?? 1) <= 0;
+      if (raidExpired) {
+        void checkAndCleanupExpiredRaid(ctx.drizzle, queue.clanBattle.attackerEntityId);
+        return { inQueue: false, queue: null, isClaiming: false };
+      }
```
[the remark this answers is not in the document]
ifroshty: Fixed latest commit

MathiasGruber:
## "Completed" label removed

This PR has **3 unresolved review thread(s)**. The "Completed" label has been automatically removed.

Please resolve all review threads before re-applying the label.

app/src/server/api/routers/raids.ts
theeneon: @MathiasGruber is the following above to your liking or is there anything else you think we might be missing/didn't cover? 

Also is there anything you can implement so that we can run raids atm and see if we can track down why  and where players are getting stuck.

MathiasGruber: Thanks for addressing this! :)

PR #1296, opened 2026-03-27
fix: clean up pending shrine battle queues during battle and war reso…

ifroshty:
## What was implemented

Fixed a bug where **pending shrine battle queue entries were left behind** after shrine battles ended or after a Village War / Raid ended. This could leave users stuck in `QUEUED` status and keep stale `MpvpBattleUser` / `MpvpBattleQueue` rows around for shrine battles that would never start.

### What the fix does

- Extended battle-end cleanup in `app/src/libs/combat/database.ts` so `SHRINE_BATTLE` now follows the same queued-user cleanup path as `RAID`
- Deletes `MpvpBattleUser` rows for completed/cleaned-up shrine battles before deleting the matching `MpvpBattleQueue` rows, preserving the existing dependency order
- Added Village War / Raid end cleanup in `app/src/libs/war.ts` to reset users in pending shrine battle queues back to `AWAKE`
- Added matching deletion of stale pending `SHRINE_BATTLE` records from both `MpvpBattleUser` and `MpvpBattleQueue` when a `VILLAGE_WAR` or `WAR_RAID` ends
- Kept the cleanup ordering explicit: status reset/update runs first, then battle-user rows are deleted, then queue rows are deleted

### Root cause

The existing cleanup logic handled:
- sector-war pending shrine queues when a `SECTOR_WAR` ended
- raid battle queues when a `RAID` battle ended

But it did **not** fully cover:
- `SHRINE_BATTLE` cleanup when an individual shrine battle resolved
- pending shrine queues created under `VILLAGE_WAR` / `WAR_RAID` when the parent war ended

That left orphaned queue/user records and could keep affected players in `QUEUED` state even though the associated shrine fight or war was already over.

### Guard conditions

| Guard | Purpose |
|---|---|
| `["RAID", "SHRINE_BATTLE"].includes(newBattle.battleType)` in `database.ts` | Restricts battle-level queue cleanup to the battle types that use `MpvpBattleQueue` / `MpvpBattleUser` |
| `mbq.battleType = 'SHRINE_BATTLE'` in `war.ts` | Ensures war-end cleanup only touches pending shrine battle queues |
| `mbq.battleId IS NULL` in `war.ts` | Only cleans up queued shrine battles that never started; active/linked battles are untouched |
| attacker/defender village filters in `war.ts` | Limits cleanup to shrine queues involving the villages from the war that just ended |
| `ud.status = 'QUEUED'` in the reset query | Prevents overwriting unrelated user states |

## Why

When a war ends, any pending shrine battles tied to that war are no longer valid. If those queue entries remain, they become stale state that can block users from normal play and create inconsistent battle queue data. The same applies when a shrine battle itself resolves but its queue records are not removed.

## Guardrails

- Cleanup order remains intentional: update user status first, then delete `MpvpBattleUser`, then delete `MpvpBattleQueue`
- Village War / Raid cleanup only targets pending shrine queues for the villages involved in the ended war
- Existing sector-war-specific cleanup remains unchanged and separate from the Village War / Raid path

## Breaking changes

None.

## License

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cleanup so shrine battles and raids conclude correctly, preventing stuck or inconsistent battle states.
  * Improved queue handling at war end to ensure player statuses return to active and pending entries are removed in the right order.
  * Ensured allied/participant villages are properly included in post-war cleanup to keep progression and histories consistent.

### DIFF
--- a/app/src/libs/combat/database.ts
+++ b/app/src/libs/combat/database.ts
@@ -109,7 +109,7 @@ export const updateBattle = async (
           ]
         : []),
       // Clean up battle users in parallel - safe since they only read from mpvpBattleQueue
-      ...(["RAID", "SHRINE_BATTLE"].includes(newBattle.battleType)
+      ...(["RAID", "SHRINE_WAR"].includes(newBattle.battleType)
         ? [
             client
               .delete(mpvpBattleUser)
@@ -128,7 +128,7 @@ export const updateBattle = async (
 
     // Delete queue entries AFTER the Promise.all to ensure mpvpBattleUser
     // subquery has completed (it references mpvpBattleQueue)
-    if (["RAID", "SHRINE_BATTLE"].includes(newBattle.battleType)) {
+    if (["RAID", "SHRINE_WAR"].includes(newBattle.battleType)) {
       await client
         .delete(mpvpBattleQueue)
         .where(eq(mpvpBattleQueue.battleId, newBattle.id));


app/src/libs/war.ts
```diff
@@ -493,10 +492,26 @@ export const handleWarEnd = async (activeWar: FetchActiveWarsReturnType) => {
...
+            WHERE mbq.battleType = 'SHRINE_BATTLE'
+              AND mbq.battleId IS NULL
+              AND (
+                mbq.attackerEntityId IN (${activeWar.attackerVillageId}, ${activeWar.defenderVillageId})
+                OR mbq.defenderEntityId IN (${activeWar.attackerVillageId}, ${activeWar.defenderVillageId})
+              )
+              AND ud.status = 'QUEUED'
+          `),
+        ]
+      : []),
```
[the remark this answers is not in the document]
ifroshty: Handled. The war-end shrine cleanup now includes all villages involved in the war, not just the two primary villages.

I added a shared `involvedVillageIds` list from the attacker, defender, and all `warAllies`, and reused it in the queued-user reset, `MpvpBattleUser` delete, and `MpvpBattleQueue` delete paths in `app/src/libs/war.ts`.

That covers ally vs ally shrine queues that were previously missed.

app/src/libs/combat/database.ts
```diff
@@ -108,8 +108,8 @@ export const updateBattle = async (
               }),
           ]
         : []),
-      // Clean up raid battle users - can run in parallel since it only reads from mpvpBattleQueue
-      ...(newBattle.battleType === "RAID"
+      // Clean up battle users in parallel - safe since they only read from mpvpBattleQueue
+      ...(["RAID", "SHRINE_BATTLE"].includes(newBattle.battleType)
```
MathiasGruber: Was this tested? the enum is SHRINE_WAR, not SHRINE_BATTLE

ifroshty: Copy and paste problem, adjusted it to the proper string

MathiasGruber (approved):

PR #1321, opened 2026-04-13
Fix/shrine attack notifications 1291

ifroshty:
# Pull Request

Adds a stale-lobby cleanup pass to the existing `/api/shrine-maintenance` cron so shrine-battle lobbies that never progressed into a real battle get purged, and the users stuck in them get released back to `AWAKE`. Also closes a race in that cleanup path that the peer review caught.

**What was implemented**

- New `runStaleShrineLobbyCleanup` step inside the shrine-maintenance cron (runs every minute alongside the existing boost tick). It deletes `mpvpBattleQueue` rows whose `battleType = SHRINE_BATTLE`, `battleId IS NULL`, and `createdAt` is older than `SHRINE_BATTLE_LOBBY_SECONDS + SHRINE_BATTLE_STALE_LOBBY_SECONDS` (60s lobby window + 5 min grace = 6 min). For any user still sitting in those stale lobbies with `status = QUEUED`, it flips them back to `AWAKE`.
- New `SHRINE_BATTLE_STALE_LOBBY_SECONDS = 300` constant in `drizzle/constants.ts` so the grace window is a single source of truth.
- The `mpvpBattleUser` delete is gated on a Drizzle subquery that re-checks `battleId IS NULL` at delete time. Without this guard, a concurrent `initiateShrineBattle` CAS-claim landing between our initial SELECT and our DELETE would strand a now-live battle with zero `mpvpBattleUser` rows, breaking the `fetchActiveUserMpvpBattles` join.
- Unit test coverage for the cleanup function, including a regression guard that partial-mocks `drizzle-orm` so any future change reverting the user delete to the raw `inArray(..., staleIds)` form fails the assertion.

**Why**

Players were reporting shrine-attack lobbies that stayed visible long after their creator had wandered off, blocking their own users from joining new battles (`QUEUED` status never cleared). The cron pass fixes the persistent-instance symptom. The subquery guard fixes a race condition Codex peer review flagged: since PlanetScale has no transactions, the initial SELECT + later DELETE in the cleanup opens a window where `initiateShrineBattle` can claim the queue row between the two, and without the guard the cleanup would still strip the user rows of the newly-live battle.

**Screenshots**

N/A — backend cron change, no UI surface.

**Breaking changes**

None. The new constant is additive, the cleanup job runs in parallel with the existing boost tick via `Promise.all`, and the success-message string adds new fields without removing old ones.

**Test plan**

- [x] `vitest run tests/app/api/shrine-maintenance/route.test.ts` — 2 tests pass
- [x] `tsc --noEmit` clean
- [x] `biome lint` clean on the route file
- [ ] Observe a full cron cycle in staging and confirm the `stale lobbies cleared N (M users reset)` log line
- [ ] Induce a stuck lobby in staging (create shrine attack, abandon past the 6-minute window) and confirm it is cleared on the next tick without disturbing any concurrent live shrine battle

## License

## Summary by CodeRabbit

* **Bug Fixes**
  * Stale shrine lobbies are now auto-cleared and player statuses reset.
  * Sector lobby UI and join/leave/initiate actions immediately exclude expired lobbies, preventing joins and surfacing expired-state responses.

* **New Features**
  * Maintenance runs now report counts of cleared lobbies and users reset.

* **Tests**
  * Added tests covering stale-lobby cleanup and the empty-case behavior.

### DIFF
--- a/app/src/app/api/shrine-maintenance/route.ts
+++ b/app/src/app/api/shrine-maintenance/route.ts
@@ -276,53 +276,55 @@ export async function runStaleShrineLobbyCleanup(
     return { lobbiesCleared: 0, usersReset: 0 };
   }
 
-  const staleIds = staleLobbies.map((row) => row.id);
+  const deletedQueueIds = staleLobbies.map((row) => row.id);
 
-  // Build the subquery once and reuse it for both the status reset and the
-  // delete, so both are guarded against a concurrent initiateShrineBattle
-  // CAS-claim (which sets battleId before initiateBattle transitions users
-  // to BATTLE — see the two-step sequence in initiateShrineBattle).
-  const stillStaleQueues = db
-    .select({ id: mpvpBattleQueue.id })
-    .from(mpvpBattleQueue)
-    .where(
-      and(
-        inArray(mpvpBattleQueue.id, staleIds),
-        isNull(mpvpBattleQueue.battleId),
-        lt(mpvpBattleQueue.createdAt, cutoff),
-      ),
-    );
-
-  const stillQueuedUsers = db
-    .select({ userId: mpvpBattleUser.userId })
-    .from(mpvpBattleUser)
-    .where(inArray(mpvpBattleUser.clanBattleId, stillStaleQueues));
-
-  const [resetResult] = await Promise.all([
-    db
-      .update(userData)
-      .set({ status: "AWAKE" })
-      .where(
-        and(inArray(userData.userId, stillQueuedUsers), eq(userData.status, "QUEUED")),
-      ),
-    db
-      .delete(mpvpBattleUser)
-      .where(inArray(mpvpBattleUser.clanBattleId, stillStaleQueues)),
-  ]);
-  const usersReset = resetResult.rowsAffected ?? 0;
-
+  // Delete parent rows FIRST — this is the atomic gate that closes the race
+  // with concurrent initiateShrineBattle calls. Any claim CAS that runs after
+  // this point will find the row gone (rowsAffected=0) and abort. Queues
+  // already claimed before this delete have battleId set and are skipped by
+  // the isNull guard; their users will have been transitioned to BATTLE by
+  // initiateBattle, so the status CAS on the update below protects them.
   const deleteResult = await db
     .delete(mpvpBattleQueue)
     .where(
       and(
-        inArray(mpvpBattleQueue.id, staleIds),
+        inArray(mpvpBattleQueue.id, deletedQueueIds),
         eq(mpvpBattleQueue.battleType, "SHRINE_BATTLE"),
         isNull(mpvpBattleQueue.battleId),
         lt(mpvpBattleQueue.createdAt, cutoff),
       ),
     );
 
-  return { lobbiesCleared: deleteResult.rowsAffected ?? 0, usersReset };
+  if (deleteResult.rowsAffected === 0) {
+    return { lobbiesCleared: 0, usersReset: 0 };
+  }
+
+  // Clean up children of the deleted queues. Scope userData reset through a
+  // subquery on mpvpBattleUser so we only reset users actually in those queues.
+  const deletedUsersSubquery = db
+    .select({ userId: mpvpBattleUser.userId })
+    .from(mpvpBattleUser)
+    .where(inArray(mpvpBattleUser.clanBattleId, deletedQueueIds));
+
+  const [resetResult] = await Promise.all([
+    db
+      .update(userData)
+      .set({ status: "AWAKE" })
+      .where(
+        and(
+          inArray(userData.userId, deletedUsersSubquery),
+          eq(userData.status, "QUEUED"),
+        ),
+      ),
+    db
+      .delete(mpvpBattleUser)
+      .where(inArray(mpvpBattleUser.clanBattleId, deletedQueueIds)),
+  ]);
+
+  return {
+    lobbiesCleared: deleteResult.rowsAffected ?? 0,
+    usersReset: resetResult.rowsAffected ?? 0,
+  };
 }
 
 /** Creates an updated shrineSettings object with new activeBoosts */
--- a/app/src/app/api/shrine-maintenance/route.ts
+++ b/app/src/app/api/shrine-maintenance/route.ts
@@ -1,8 +1,7 @@
 import { and, eq, gt, inArray, isNull, lt, lte } from "drizzle-orm";
 import { cookies } from "next/headers";
 import {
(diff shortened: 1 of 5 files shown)


app/src/server/api/routers/shrine.ts
```diff
@@ -728,12 +729,19 @@ export const shrineRouter = createTRPCRouter({
...
+          (SHRINE_BATTLE_LOBBY_SECONDS + SHRINE_BATTLE_STALE_LOBBY_SECONDS) * 1000,
+      );
       // Fetch all battles with FIFO ordering (oldest first)
       const battles = await ctx.drizzle.query.mpvpBattleQueue.findMany({
         where: and(
           eq(mpvpBattleQueue.battleType, "SHRINE_BATTLE"),
           eq(mpvpBattleQueue.sector, input.sectorNumber),
           isNull(mpvpBattleQueue.battleId),
+          gt(mpvpBattleQueue.createdAt, shrineLobbyStaleBefore),
         ),
```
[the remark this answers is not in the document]
ifroshty: newest commit addresses this

app/src/app/api/shrine-maintenance/route.ts
```diff
@@ -237,6 +252,80 @@ async function runShrineBoostTick(now: Date = new Date()) {
...
 
+/**
+ * Deletes shrine battle lobbies that never progressed into a real battle and
+ * resets users who are still marked as queued for those stale lobbies.
+ */
+export async function runStaleShrineLobbyCleanup(
+  now: Date,
+  db: ShrineMaintenanceDb = drizzleDB,
+) {
+  const cutoffSeconds = SHRINE_BATTLE_LOBBY_SECONDS + SHRINE_BATTLE_STALE_LOBBY_SECONDS;
```
MathiasGruber: The same computation is repeated 6 times across 3 files:

```
route.ts
Lines 263-263
  const cutoffSeconds = SHRINE_BATTLE_LOBBY_SECONDS + SHRINE_BATTLE_STALE_LOBBY_SECONDS;

profile.ts
Lines 2237-2240
  const shrineLobbyStaleBefore = new Date(
    now.getTime() -
      (SHRINE_BATTLE_LOBBY_SECONDS + SHRINE_BATTLE_STALE_LOBBY_SECONDS) * 1000,
  );

shrine.ts
Lines 734-737
      const shrineLobbyStaleBefore = new Date(
        Date.now() -
          (SHRINE_BATTLE_LOBBY_SECONDS + SHRINE_BATTLE_STALE_LOBBY_SECONDS) * 1000,
      );
...and 3 more in shrine.ts at lines 813, 1027, 1252.
```

Suggestion: Add a helper in @/drizzle/constants.ts next to the related constants:

```
export function shrineLobbyFreshAfter(now: Date = new Date()): Date {
  return new Date(
    now.getTime() -
      (SHRINE_BATTLE_LOBBY_SECONDS + SHRINE_BATTLE_STALE_LOBBY_SECONDS) * 1000,
  );
}
```

Then all 6 call sites become just shrineLobbyFreshAfter(now) or shrineLobbyFreshAfter().

app/src/server/api/routers/shrine.ts
MathiasGruber, line 1159: It's not a bug (the subquery at line 1185 already gates on isNull(battleId), and the cleanup cron will handle the rest), but for consistency with the other endpoints and to give a better error message ("Shrine battle expired" vs the generic "Not in this shrine battle queue"), consider adding the same guard to the activeQueueSubquery. e.g. something like:

```typescript
      const shrineLobbyStaleBefore = shrineLobbyFreshAfter();
      const activeQueueSubquery = ctx.drizzle
        .select({ id: mpvpBattleQueue.id })
        .from(mpvpBattleQueue)
        .where(
          and(
            eq(mpvpBattleQueue.id, input.shrineBattleId),
            isNull(mpvpBattleQueue.battleId),
            gt(mpvpBattleQueue.createdAt, shrineLobbyStaleBefore),
          ),
        );
```

app/src/server/api/routers/shrine.ts
```diff
@@ -802,11 +810,17 @@ export const shrineRouter = createTRPCRouter({
         orderBy: desc(mpvpBattleUser.createdAt),
       });
 
+      const shrineLobbyStaleBefore = new Date(
+        Date.now() -
+          (SHRINE_BATTLE_LOBBY_SECONDS + SHRINE_BATTLE_STALE_LOBBY_SECONDS) * 1000,
+      );
+
       // Find the first active shrine battle (battleId IS NULL means not started)
       const activeEntry = queueEntries.find(
```
MathiasGruber: Instead of doing the filtering here, lets calculate the shrineLobbyStaleBefore before the query, and push the filtering to the SQL where-claude instead?

app/src/server/api/routers/shrine.ts
ifroshty: shrine.ts:1155-1168 fix

app/src/app/api/shrine-maintenance/route.ts
```diff
@@ -237,6 +251,80 @@ async function runShrineBoostTick(now: Date = new Date()) {
...
+    .where(
+      and(
+        inArray(mpvpBattleQueue.id, staleIds),
+        eq(mpvpBattleQueue.battleType, "SHRINE_BATTLE"),
+        isNull(mpvpBattleQueue.battleId),
+        lt(mpvpBattleQueue.createdAt, cutoff),
+      ),
+    );
+
+  return { lobbiesCleared: deleteResult.rowsAffected ?? 0, usersReset };
```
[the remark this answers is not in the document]
ifroshty: confirm if this is resolved

app/tests/app/api/shrine-maintenance/route.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    expect(result).toEqual({ lobbiesCleared: 2, usersReset: 2 });
+    expect(db.update).toHaveBeenCalledWith(userData);
+    expect(updateSet).toHaveBeenCalledWith({ status: "AWAKE" });
+    expect(db.delete).toHaveBeenNthCalledWith(1, mpvpBattleUser);
+    expect(db.delete).toHaveBeenNthCalledWith(2, mpvpBattleQueue);
+    // Both the status reset and the user-row delete are gated by subqueries
+    // (calls 2 and 3). Reverting either to a flat inArray(..., staleIds)
+    // would reduce db.select to fewer calls.
+    expect(db.select).toHaveBeenCalledTimes(3);
```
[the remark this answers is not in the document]
ifroshty: confirm if this is resolved

ifroshty:
@greptileai

MathiasGruber (approved):

PR #1322, opened 2026-04-13
fix: dedupe affected combat tiles

### LINKED ISSUE #1228: Method: aoe_circle_spawn Jutsu can hit same tile multiple times
**Describe the bug**
Some abilities with shapes like aoe_circle_spawn cause `getAffectedTiles` to return duplicate tiles. This can result in the same tile of a barrier getting hit multiple times for 1 ability.

**To Reproduce**
In battle, one player summons a wall/barrier.
Another player uses an aoe_circle_spawn ability centered on the barrier (not the player I think).

**Expected behavior**
Abilities that hit multiple tiles should only trigger effects on each tile once, so `getAffectedTiles` should always return a unique list of tiles with no duplicates.

**Further Information**
https://github.com/studie-tech/TheNinjaRPG/blob/d81ae6974cbfcd34bdf2bcf4077e4c673d2a8618/app/src/libs/combat/util.ts#L2427

Can maybe filter out duplicates like this:
```ts
const visited = new Set<string>();

const tiles = grid.traverse(
  spiral<TerrainHex>({ start: [b.q, b.r], radius: 1 })
).filter(t => {
  const key = `${t.q},${t.r}`;
  if (visited.has(key)) return false;
  visited.add(key);
  return true;
});
```

**Screenshots**

Image

TheNinja-RPG:
## Summary
- dedupe affected combat tiles by battlefield coordinates before they are added to the returned sets
- keep the fix inside getAffectedTiles so duplicate traversal output cannot hit the same barrier more than once
- add a regression test covering duplicate coordinate entries for AOE_CIRCLE_SPAWN

Fixes #1228

## Test plan
- cd app && NODE_ENV=test NEXT_PUBLIC_PUSHER_APP_KEY=test NEXT_PUBLIC_PUSHER_APP_CLUSTER=test NEXT_PUBLIC_BASE_URL=https://example.com ~/.bun/bin/bun run test --run tests/libs/combat/getAffectedTiles.test.ts

## Summary by CodeRabbit

* **Bug Fixes**
  * Area-of-effect targeting now deduplicates overlapping tiles and consistently classifies each unique tile as valid (green) or invalid (red), including cases with overlapping traversals and barriers.

* **Tests**
  * Added tests covering deduplication and AOE scenarios to ensure single-instance targets and correct classification.

[Review Change Stack](https://app.coderabbit.ai/change-stack/studie-tech/TheNinjaRPG/pull/1322)


app/tests/libs/combat/getAffectedTiles.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      grid,
+      users: [createUser()],
+      ground: [createBarrier(3, 4)],
+      userId: "actor-id",
+    });
+
+    expect(red.size).toBe(0);
+    expect(green.size).toBe(1);
+    expect([...green].map((tile) => [tile.col, tile.row])).toEqual([[3, 4]]);
+  });
```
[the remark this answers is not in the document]
MathiasGruber: Added a test case for `AOE_WALL_SHOOT` that simulates two line traversals both starting from the same tile `b`, verifying that the shared start tile is deduplicated in the result set (3 unique tiles instead of 4).

app/src/libs/combat/util.ts
```diff
@@ -2547,19 +2560,19 @@ export const getAffectedTiles = (info: {
     if (tiles) tiles = tiles.filter((t) => t !== a);
   } else if (action.method === "ALL") {
     grid.forEach((target) => {
-      if (isValidMove({ action, target, user, users, barriers, clicked: b })) {
-        green.add(target);
-      }
+      addAffectedTile(
+        target,
+        isValidMove({ action, target, user, users, barriers, clicked: b }),
```
MathiasGruber: For the ALL method, we don't want to add it to the red list. It should retain the functionality as before visually for the user

MathiasGruber (changes requested):

app/src/libs/combat/util.ts
MathiasGruber: Fixed. The ALL method now only adds valid tiles to the green set and never routes anything to red, restoring the original visual behavior. Deduplication via seenTiles is still applied.

app/tests/libs/combat/getAffectedTiles.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+describe("getAffectedTiles", () => {
+  it("deduplicates the shared start tile for aoe wall shoot", () => {
+    // AOE_WALL_SHOOT traverses two lines both starting at b, so b appears twice
+    const b = createTile(3, 4);
+    const northTile = createTile(3, 3);
+    const southTile = createTile(3, 5);
+    // Simulate two line traversals that both include b
+    const grid = {
+      traverse: () => [b, northTile, b, southTile],
```
MathiasGruber: This wall-shoot test does not actually exercise the new coordinate-key dedup code path. The mocked `grid.traverse` returns `[b, northTile, b, southTile]` where `b` is the same JS object reference in both positions. Since `green` is a `Set<TerrainHex>` and JS `Set` uses reference equality, the duplicate would already be filtered out by the `Set` itself even without the `seenTiles` coordinate-key check.

To genuinely validate the fix, use two distinct objects at the same coordinates, e.g.:

```ts
const b = createTile(3, 4);
const bDup = createTile(3, 4); // distinct object, same coords
const northTile = createTile(3, 3);
const southTile = createTile(3, 5);
const grid = {
  traverse: () => [b, northTile, bDup, southTile],
} as unknown as Grid<TerrainHex>;
```

That way the `green.size === 3` assertion only passes if the new coordinate-keyed `seenTiles` set is doing the work.

MathiasGruber: Good catch — switched to a second `createTile(3, 4)` instance (`bDup`) so the assertion only passes when the coordinate-keyed `seenTiles` dedup is doing the work. Reference-equality on the `Set` alone no longer filters it. Pushed in f7e17d9c7.

MathiasGruber (approved):
LGTM

MathiasGruber (approved):
LGTM

app/src/libs/combat/util.ts
```diff
@@ -2555,11 +2571,10 @@ export const getAffectedTiles = (info: {
...
   tiles?.forEach((target) => {
-    if (isValidMove({ action, target, user, users, barriers, clicked: b })) {
-      green.add(target);
-    } else {
-      red.add(target);
-    }
+    addAffectedTile(
+      target,
+      isValidMove({ action, target, user, users, barriers, clicked: b }),
+    );
```
MathiasGruber: `isValidMove` is evaluated eagerly here for every tile in the traversal, even when `addAffectedTile` is going to immediately discard the tile as a duplicate. This is asymmetric with the `ALL` branch above (lines 2562-2569), which correctly checks `seenTiles` before invoking `isValidMove`.

This matters for `AOE_WALL_SHOOT` / `AOE_LARGE_WALL_SHOOT`, where the starting tile `b` is included in both `line` traversals (N+S or W+E), so `isValidMove` runs twice on the same tile. `isValidMove` does two linear scans of `barriers` and `users` plus an effect-array scan, so it isn't free.

Either move the dedup check into `addAffectedTile` before the `isValid` argument is computed (e.g. accept a thunk `() => isValidMove(...)`, or inline the seen-check before the call), or have `addAffectedTile` itself call `isValidMove` so the early-return on duplicate happens first. That keeps the dedup behavior consistent across both branches and avoids the redundant compute.

MathiasGruber: Good catch. Converted `addAffectedTile`'s second argument into a thunk so `isValidMove` is only invoked after the `seenTiles` dedup check passes. This eliminates the redundant `isValidMove` call on the shared starting tile in `AOE_WALL_SHOOT` / `AOE_LARGE_WALL_SHOOT` traversals while keeping the existing red/green semantics. Fixed in 4424f5b54.

Phrosfire:
The issue has not been fixed.  Neon and I ran a test and the damage still hits the barriers multiple times

image

app/src/libs/combat/util.ts
```diff
@@ -2457,9 +2457,22 @@ export const getAffectedTiles = (info: {
...
   const green = new Set<TerrainHex>();
   const red = new Set<TerrainHex>();
+  const seenTiles = new Set<string>();
   const user = users.find((u) => u.userId === userId);
   let tiles: Grid<TerrainHex> | undefined;
 
+  const addAffectedTile = (target: TerrainHex, getIsValid: () => boolean) => {
+    const key = `${target.col},${target.row}`;
+    if (seenTiles.has(key)) return;
+    seenTiles.add(key);
```
MathiasGruber: Re Phrosfire's report ("Neon and I ran a test and the damage still hits the barriers multiple times"): that report is correct, and this PR does not fix it. The coord-key dedup added here prevents the SAME tile coordinate from being returned twice, but the in-game complaint is about a SINGLE barrier being hit multiple times when it lies on the path from the actor to multiple DIFFERENT affected tiles (e.g. AOE_CIRCLE_SPAWN with a barrier at the spawn center -> every outer-ring tile's path crosses that barrier).

The relevant dedup is in `app/src/libs/combat/actions.ts` (around line 830 on main):

```ts
barriers.forEach((barrier) => {
  const idx = `${barrier.id}-${effect.id}`;
  if (!barrierAttacks.includes(idx)) {
    barrierAttacks.push(idx);
    ...
  }
});
```

This is inside an `affectedTiles.forEach -> action.effects.forEach` loop, and `effect` comes from `realizeTag`. `realizeTag` in `app/src/libs/combat/tags.ts` line 70 unconditionally regenerates `tag.id = nanoid();` on every call. So `effect.id` is unique per (tile, tag) iteration, which means `${barrier.id}-${effect.id}` is unique per tile -- the dedup never matches across different tiles and a given barrier gets re-hit once per affected tile.

Suggested fix (in `actions.ts`, not this file): drop `effect.id` from the key. Either

- `const idx = barrier.id;` for one hit per barrier per ability cast (matches the apparent intent of `barrierAttacks` being scoped to a single `performAction` invocation), or
- `const idx = `${barrier.id}-${tag.type}`;` if a damage tag and a pierce tag on the same ability should each get one hit on the barrier.

Would be good to land that change in this PR (or a follow-up) and add a regression test along the lines of the new ones in `getAffectedTiles.test.ts` so this bug stays fixed.

MathiasGruber (commented):
Prior-feedback verification: 5 prior threads check out as resolved in code, but Phrosfire's bug report is real and not fixed by this PR. See inline comment.

app/src/libs/combat/util.ts
MathiasGruber: Good catch — applied the conservative variant. Keyed barrierAttacks by `${barrier.id}-${tag.type}` (instead of `${barrier.id}-${effect.id}`), so each barrier on the path absorbs at most once per (barrier, tag) per cast across all affected tiles. Confirmed realizeTag in tags.ts:70 regenerates effect.id on every iteration, which was defeating the dedup. The tag-type granularity preserves the existing behavior where an ability with both damage and pierce tags lands one of each on the barrier. Pushed in 52804548b.

MathiasGruber:
@Phrosfire how about now?

MathiasGruber (approved):
LGTM

app/src/libs/combat/actions.ts
```diff
@@ -827,7 +827,11 @@ export const insertAction = (info: {
             // Extra: If no target, check if there is a barrier & apply damage only
             if (["damage", "pierce"].includes(tag.type)) {
               barriers.forEach((barrier) => {
-                const idx = `${barrier.id}-${effect.id}`;
+                // Key by tag.type (not effect.id) so a single barrier is hit
+                // at most once per (barrier, tag) per cast. realizeTag assigns
+                // a fresh nanoid to effect.id on every iteration, so including
+                // it in the key would defeat the dedup across affected tiles.
+                const idx = `${barrier.id}-${tag.type}`;
```
MathiasGruber: Keying by `tag.type` collapses across distinct tags of the same type. If an action's `effects` array contains more than one `damage` (or `pierce`) tag, only the first one will ever land on a given barrier per cast — every subsequent same-typed tag short-circuits on `barrierAttacks.includes(idx)` and is silently dropped. Consider keying by the tag's identity within this cast (e.g. the array index from `action.effects.forEach((tag, tagIndex)` -> `` `${barrier.id}-${tagIndex}` ``) so each distinct tag still applies once per barrier while still deduping across affected tiles.

MathiasGruber (commented):
One concern about the barrier dedup key granularity.

app/src/libs/combat/actions.ts
MathiasGruber: Good catch — switched the key to use `tagIndex` from `action.effects.forEach((tag, tagIndex) => ...)` so each distinct tag (even multiple of the same `tag.type`) still applies once per barrier, while the cross-tile dedup is preserved. Pushed in 46aa2557a.

MathiasGruber (approved):
LGTM

app/src/libs/combat/util.ts
```diff
@@ -2547,6 +2560,9 @@ export const getAffectedTiles = (info: {
     if (tiles) tiles = tiles.filter((t) => t !== a);
   } else if (action.method === "ALL") {
     grid.forEach((target) => {
+      const key = `${target.col},${target.row}`;
+      if (seenTiles.has(key)) return;
+      seenTiles.add(key);
```
MathiasGruber: The ALL branch reimplements the same `seenTiles` add/has check that `addAffectedTile` already encapsulates. Could be collapsed by routing through the helper (e.g. `addAffectedTile(target, () => isValidMove(...))`); the only behavioral difference today is that the ALL path silently drops invalid tiles instead of adding them to `red`. Either fold that into the helper (optional flag to skip red) or accept the small duplication — having two copies of the coordinate-key logic is easy to drift.

MathiasGruber (commented):
Coordinate-keyed dedup looks correct overall, and the barrier-key change to tagIndex achieves stable cross-tile dedup while still distinguishing tags within a cast. One DRY nit on the ALL branch noted inline.

app/src/libs/combat/util.ts
MathiasGruber: Folded the dedup into the helper. `addAffectedTile` now takes an optional `{ skipRed: true }` flag, and the ALL branch routes through it instead of reimplementing the `seenTiles` add/has check. Coordinate-key logic lives in one place now — no risk of drift between the two paths, behavior unchanged. Pushed in 44bca93f5.

MathiasGruber (approved):
LGTM

app/src/libs/combat/actions.ts
```diff
@@ -827,7 +827,12 @@ export const insertAction = (info: {
...
             if (["damage", "pierce"].includes(tag.type)) {
               barriers.forEach((barrier) => {
-                const idx = `${barrier.id}-${effect.id}`;
+                // Key by tag's array position within this cast so distinct tags
+                // (even of the same type) each land once per barrier, while
+                // still deduping across affected tiles. realizeTag assigns a
+                // fresh nanoid to effect.id every iteration, so it cannot be
+                // used as part of the key.
+                const idx = `${barrier.id}-${tagIndex}`;
                 if (!barrierAttacks.includes(idx)) {
```
MathiasGruber: `barrierAttacks` is a `string[]` and `Array.prototype.includes` is O(N), so this membership check runs inside `affectedTiles.forEach -> action.effects.forEach -> barriers.forEach`. For a max-radius AOE with multiple damage tags and several barriers, this becomes O(tiles * effects^2 * barriers^2) per cast in the hot `performAction` path. Cheap fix: declare `barrierAttacks` as `new Set<string>()` and use `.has` / `.add`. Same dedup semantics, constant-time lookup.

MathiasGruber (commented):
Performance review

app/src/libs/combat/actions.ts
MathiasGruber: Good catch — switched `barrierAttacks` from `string[]` to `Set<string>` so the dedup check is O(1) instead of O(N). Removes the quadratic membership scan inside the `affectedTiles -> action.effects -> barriers` loop in the `performAction` hot path. Pushed in ae0fb9016.

MathiasGruber (approved):
Approved.

MathiasGruber (approved):
Approved. All previous review threads resolved; no new logic, readability, DRY, or performance concerns surfaced on re-review.

MathiasGruber (approved):
LGTM

MathiasGruber (approved):
LGTM

app/src/libs/combat/actions.ts
```diff
@@ -827,9 +827,14 @@ export const insertAction = (info: {
...
             if (["damage", "pierce"].includes(tag.type)) {
               barriers.forEach((barrier) => {
-                const idx = `${barrier.id}-${effect.id}`;
-                if (!barrierAttacks.includes(idx)) {
-                  barrierAttacks.push(idx);
+                // Key by tag's array position within this cast so distinct tags
+                // (even of the same type) each land once per barrier, while
+                // still deduping across affected tiles. realizeTag assigns a
+                // fresh nanoid to effect.id every iteration, so it cannot be
+                // used as part of the key.
```
MathiasGruber: This comment is conversational/historical — it explains why the previous key (`effect.id`) didn't work rather than describing the current code's purpose. Per CLAUDE.md ("Natural Comments Only"), comments should describe what the code does, not reference prior bugs or review feedback. Consider trimming to something like: `// Key by barrier + tag position so each distinct tag lands once per barrier, even across multiple affected tiles.`

MathiasGruber: Removed the historical/why-not-effect.id comment. Code now stands on its own with the descriptive key string.

MathiasGruber (approved):
LGTM

app/src/libs/combat/actions.ts
```diff
@@ -827,9 +827,9 @@ export const insertAction = (info: {
...
               barriers.forEach((barrier) => {
-                const idx = `${barrier.id}-${effect.id}`;
-                if (!barrierAttacks.includes(idx)) {
-                  barrierAttacks.push(idx);
+                const idx = `${barrier.id}-${tagIndex}`;
+                if (!barrierAttacks.has(idx)) {
+                  barrierAttacks.add(idx);
                   targetUsernames.push("barrier");
                   targetGenders.push("it");
                   const barrierEffect = structuredClone(effect);
```
MathiasGruber: `structuredClone(effect)` here is redundant work in the `performAction` hot path. `realizeTag` in `tags.ts:96` already returns a `structuredClone(tag)`, so `effect` is itself a fresh deep copy unique to this `(tile, tag)` iteration. The clone at this line therefore deep-copies the same object a second time on every (tile, tag, barrier) triple that passes the dedup.

Since the only mutations after this point are top-level field assignments (`targetType`, `targetId`, `id`, `barrierAbsorb`), a shallow spread is sufficient and cheaper:

```ts
const barrierEffect = { ...effect, targetType: "barrier", targetId: barrier.id, id: nanoid() };
if ("absorbPercentage" in barrier) {
  barrierEffect.barrierAbsorb = barrier.absorbPercentage;
}
```

structuredClone is significantly slower than a spread for objects this shape, and this runs inside `affectedTiles -> action.effects -> barriers` for every damage/pierce tag.

app/src/libs/combat/actions.ts
```diff
@@ -827,9 +827,9 @@ export const insertAction = (info: {
             // Extra: If no target, check if there is a barrier & apply damage only
             if (["damage", "pierce"].includes(tag.type)) {
```
MathiasGruber: Minor: `["damage", "pierce"].includes(tag.type)` allocates a fresh 2-element array on every iteration of `action.effects.forEach`, which itself runs inside `affectedTiles.forEach` — so the literal is re-allocated O(tiles * effects) times per cast in the `performAction` hot path. Hoist it to a module-level constant (e.g. `const BARRIER_DAMAGE_TYPES = new Set(["damage", "pierce"] as const)`) and use `.has`. The identical pattern lives in `isValidMove` at `util.ts:2423`/`2415`, which itself runs per-tile from `getAffectedTiles`.

app/src/libs/combat/actions.ts
MathiasGruber: Good catch — replaced `structuredClone(effect)` with a shallow spread `{ ...effect, targetType, targetId, id }`. `realizeTag` already returns a `structuredClone(tag)`, so the extra deep clone was duplicate work; only top-level fields are mutated afterwards (plus the optional `barrierAbsorb`), so the spread is sufficient. Pushed in 63b42855c.

app/src/libs/combat/actions.ts
MathiasGruber: Good catch — hoisted `["damage", "pierce"]` to a module-level `BARRIER_DAMAGE_TAG_TYPES` `Set<string>` in `combat/constants.ts` and switched both call sites (`actions.ts` barrier branch and `isValidMove` in `util.ts:2423`) to `.has`. Removes the per-iteration array allocation in the `performAction` and `getAffectedTiles` hot paths. Pushed in 63b42855c.

MathiasGruber (approved):
Approved. All 11 prior review threads resolved in code; Phrosfire's barrier-multi-hit report is correctly fixed by keying `barrierAttacks` by `${barrier.id}-${tagIndex}` outside the affected-tile loop. Logic/readability/DRY and performance re-passes surfaced no new in-scope concerns. /tnr-review-now (latest) recommends approve.

MathiasGruber (approved):
Approved. All previous review threads resolved in code, no logic/readability/DRY or performance concerns surfaced on re-review, and the latest /tnr-review-now run completed successfully with no in-scope findings.

MathiasGruber (approved):
Approved. Prior review threads remain addressed in code, no new logic/readability/DRY or performance concerns found on re-review, and the latest /tnr-review-now run completed with no in-scope findings.

app/src/libs/combat/constants.ts
```diff
@@ -18,6 +18,8 @@ export const COMBAT_LOBBY_SECONDS = 15;
 
 export const SPAR_EXPIRY_SECONDS = 120;
 
+export const BARRIER_DAMAGE_TAG_TYPES = new Set<string>(["damage", "pierce"]);
```
MathiasGruber, line 21: `process.ts:1016` still uses the inline `["damage", "pierce"].includes(effect.type)` pattern. Since this constant is meant to centralize the check, migrate that callsite to `BARRIER_DAMAGE_TAG_TYPES.has(effect.type)` too, otherwise the two checks can drift apart.

MathiasGruber (commented):
DRY finding on barrier tag-type check.

app/src/libs/combat/constants.ts
MathiasGruber: Done in 68a7e1697 — `process.ts:1016` now imports `BARRIER_DAMAGE_TAG_TYPES` from `./constants` and uses `BARRIER_DAMAGE_TAG_TYPES.has(effect.type)` instead of the inline array. Both call sites (process.ts barrier branch + actions.ts/util.ts) now go through the same constant, so the check cannot drift.

MathiasGruber (approved):
LGTM

MathiasGruber (approved):
Reviewed. All prior feedback addressed; logic, readability, and performance look clean; tests cover the dedup invariants. Cross-tile barrier dedup correctly keys on (barrier.id, tagIndex), preventing the multi-hit reported in #1228.

MathiasGruber (approved):
Approved. All prior review threads remain addressed in the current code; no new logic, readability, DRY, or performance concerns surfaced on re-review; and the latest successful /tnr-review-now run validated the Blast Shield + Kinetic Nova barrier-multi-hit scenario in live preview (each barrier hit once, not multiple times).

MathiasGruber (approved):

MathiasGruber (approved):

MathiasGruber (approved):
LGTM. All prior feedback addressed, no new logic/readability/DRY or performance concerns, TNR live-preview review passed.

MathiasGruber (approved):
Approved. All prior review threads remain addressed in code, no new logic/readability/DRY or performance concerns surfaced on re-review, and the latest /tnr-review-now live-preview run completed successfully with no in-scope findings (recommendation: approve).

MathiasGruber (approved):
Approved. All prior review threads remain addressed in code, including the Phrosfire/Neon barrier-multi-hit report (fixed via `barrierAttacks` Set keyed by `${barrier.id}-${tagIndex}` outside the affected-tile loop). Re-review surfaced no new logic, readability, DRY, or performance concerns; the latest /tnr-review-now live-preview run completed successfully with all-Pass findings.

MathiasGruber (approved):
LGTM. All review threads addressed, latest TNR review run recommended approve, no migrations.

MathiasGruber (approved):

MathiasGruber (approved):
LGTM

MathiasGruber (approved):
LGTM

MathiasGruber (approved):
Approved. All prior review threads verified as addressed in code, including the cross-tile barrier dedup via `${barrier.id}-${tagIndex}`. No new logic, readability, DRY, or performance concerns. Latest successful /tnr-review-now validated the AOE_CIRCLE_SPAWN + barrier behavior in browser.

MathiasGruber (approved):
Reviewed: prior feedback addressed, logic/readability/DRY clean, no perf concerns, /tnr-review-now smoke run flagged no code-level regressions.

MathiasGruber (approved):
LGTM. All previous feedback addressed, no new logic/perf concerns.

MathiasGruber (approved):
LGTM

MathiasGruber (approved):
LGTM

MathiasGruber (approved):
LGTM. All review threads resolved, checks green, /tnr-review-now ran successfully with no PR-specific regressions found.

MathiasGruber (approved):
LGTM. All prior review threads remain addressed in code, including the Phrosfire/Neon barrier-multi-hit report (fixed via `barrierAttacks` Set keyed by `${barrier.id}-${tagIndex}` outside the affected-tile loop). Re-review surfaced no new logic, readability, DRY, or performance concerns. The latest /tnr-review-now run was inconclusive due to test-harness issues (broker endpoint failures, stale battle state) — not code defects — and prior successful runs validated the AOE_CIRCLE_SPAWN + barrier behavior in browser with each barrier hit exactly once.

MathiasGruber (approved):
LGTM

MathiasGruber (approved):
LGTM

MathiasGruber (approved):

MathiasGruber (approved):
LGTM. All prior review threads remain addressed in code, no new logic/readability/DRY or performance concerns surfaced on re-review, and the latest /tnr-review-now run completed with only out-of-scope preview-environment blockers.

MathiasGruber (approved):

MathiasGruber (commented):
One performance note on the dedup helper.

MathiasGruber (approved):
LGTM. All prior review threads remain addressed in code, including the Phrosfire/Neon barrier-multi-hit report (fixed via `barrierAttacks` Set keyed by `${barrier.id}-${tagIndex}` outside the affected-tile loop). Re-review of logic, readability, DRY, and performance surfaced no in-scope concerns. Most recent /tnr-review-now had only preview-environment blockers (broker rejected tRPC paths, stale battle state) — prior successful runs validated the AOE_CIRCLE_SPAWN + barrier behavior in-browser with each barrier hit exactly once.

MathiasGruber (approved):

app/src/libs/combat/actions.ts
```diff
@@ -825,17 +825,19 @@ export const insertAction = (info: {
...
-                  const barrierEffect = structuredClone(effect);
-                  barrierEffect.targetType = "barrier";
-                  barrierEffect.targetId = barrier.id;
-                  barrierEffect.id = nanoid();
+                  const barrierEffect: UserEffect = {
+                    ...effect,
+                    targetType: "barrier",
+                    targetId: barrier.id,
+                    id: nanoid(),
+                  };
```
MathiasGruber: Replacing `structuredClone(effect)` with a shallow spread means `barrierEffect` aliases any nested objects on `effect` (e.g. `timeTracker`). When the parent `effect` is also pushed to `usersEffects` (lines 816/823), both entries share the same `timeTracker` map, so later mutations in process.ts/util.ts (`effect.timeTracker[targetId] = battle.round`) leak across the user-targeted and barrier-targeted copies. Keep `structuredClone(effect)` (or clone the relevant nested fields) to preserve isolation.

MathiasGruber: Good catch — reverted to `structuredClone(effect)` in 2bf3454cd. The shallow spread aliased `timeTracker` between the parent `effect` (pushed at lines 816/823) and `barrierEffect`, so later mutations in process.ts/util.ts leaked across both. Deep clone preserves isolation.

MathiasGruber (approved):
LGTM

MathiasGruber (approved):

MathiasGruber (approved):

MathiasGruber (approved):
Approved. All prior review feedback addressed; no new logic, readability, DRY, or performance findings. Latest successful TNR reviewer run confirmed live spawn-dedupe behavior; remaining 'needs follow-up' was a scope-coverage gap (could not stage a valid attacker cast against the spawned barrier in-browser), not a code defect.

app/src/libs/combat/util.ts
```diff
@@ -2457,9 +2458,26 @@ export const getAffectedTiles = (info: {
...
+    const key = `${target.col},${target.row}`;
+    if (seenTiles.has(key)) return;
+    seenTiles.add(key);
+
+    if (getIsValid()) {
+      green.add(target);
+    } else if (!options?.skipRed) {
+      red.add(target);
+    }
+  };
```
MathiasGruber: Perf: `seenTiles` dedup runs on every `getAffectedTiles` call, but duplicate tile references are only produced by the two-line traversals (`AOE_WALL_SHOOT`, `AOE_LARGE_WALL_SHOOT`). For all other methods (`SINGLE`, `AOE_CIRCLE_SPAWN`, `AOE_LINE_SHOOT`, `AOE_CIRCLE_SHOOT`, `AOE_SPIRAL_SHOOT`, `ALL`) and the grid traversal in the final loop, a single `grid.traverse(...)` call cannot yield the same hex object twice, so we pay a `Set.has`/`Set.add` + per-tile `${col},${row}` string allocation for every tile (hundreds per call) with no possible dedup benefit. This function is also invoked from `libs/threejs/combat.ts` during rendering/targeting, so the overhead lands on a hot path. Suggest gating the dedup to only the wall-shoot methods (where it's actually needed), or dedup inside those branches before joining the lines, leaving the common cases zero-overhead.

MathiasGruber: Good catch — gated the `seenTiles` dedup to `AOE_WALL_SHOOT` and `AOE_LARGE_WALL_SHOOT` only in 97f6d7714. The flag is checked once per call; the helper skips the `Set.has`/`add` + `${col},${row}` string allocation entirely for `SINGLE`, `AOE_CIRCLE_SPAWN`, `AOE_LINE_SHOOT`, `AOE_CIRCLE_SHOOT`, `AOE_SPIRAL_SHOOT`, `ALL`, and the final `grid.traverse` loop, so the rendering/targeting hot path in `libs/threejs/combat.ts` pays zero dedup overhead for the common cases. Dropped the `AOE_CIRCLE_SPAWN` dedup test since it asserted behavior that doesn't occur with real spiral traversal — the `AOE_WALL_SHOOT` test still verifies the gated path.

MathiasGruber (approved):
LGTM. All prior review threads resolved in code. No new logic, readability, DRY, or performance concerns surfaced. /tnr-review-now found no confirmed player-visible bug.

app/src/libs/combat/util.ts
```diff
@@ -2457,9 +2458,34 @@ export const getAffectedTiles = (info: {
   const radius = action.range;
   const green = new Set<TerrainHex>();
   const red = new Set<TerrainHex>();
+  // Only the two-line wall traversals can yield the same coordinate twice
+  // (both lines start at `b`). For every other method a single `grid.traverse`
+  // call visits each hex at most once, so we skip the coord-key dedup overhead
+  // on the hot rendering/targeting path.
+  const needsDedup =
+    action.method === "AOE_WALL_SHOOT" || action.method === "AOE_LARGE_WALL_SHOOT";
+  const seenTiles = needsDedup ? new Set<string>() : null;
```
MathiasGruber: `seenTiles` looks like dead code in production. `grid.traverse(...)` from honeycomb-grid returns a `Grid`, not an array, and `Grid` stores hexes in an internal `Map` keyed by coordinate string (`hex.toString()`). When the wall-shoot path calls `grid.traverse([line N, line S])`, both lines emit `b` but the resulting `Grid` only retains one canonical reference per coordinate, so the downstream `tiles?.forEach` already sees each coord at most once. Verified locally with honeycomb-grid 4.1.5: `grid.traverse([line({start:[3,4], direction:N, length:2}), line({start:[3,4], direction:S, length:2})]).size === 3`, not 4. The comment block above is also inaccurate — the two-line wall traversal does not yield the same coordinate twice on the consumer side. Recommend dropping `needsDedup`/`seenTiles`/`addAffectedTile` and reverting `getAffectedTiles` to the simple inline `isValidMove` branches; the real dedup fix lives in `actions.ts` (the `${barrier.id}-${tagIndex}` key) and stands on its own.

app/tests/libs/combat/getAffectedTiles.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  it("deduplicates the shared start tile for aoe wall shoot", () => {
+    // AOE_WALL_SHOOT traverses two lines both starting at b. Use a second
+    // distinct object at the same coords so reference equality on Set<TerrainHex>
+    // does NOT filter it — only the coordinate-keyed seenTiles dedup can.
+    const b = createTile(3, 4);
+    const bDup = createTile(3, 4);
+    const northTile = createTile(3, 3);
+    const southTile = createTile(3, 5);
+    const grid = {
+      traverse: () => [b, northTile, bDup, southTile],
```
MathiasGruber: This test does not exercise the production code path. The mocked `traverse: () => [b, northTile, bDup, southTile]` returns a raw array, but the real `Grid.traverse` returns a `Grid` whose internal `Map` already deduplicates by coord-string — duplicates like `bDup` would never reach `getAffectedTiles`'s `tiles?.forEach` in production. As a result, the assertion `green.size === 3` passes here but would pass equally if `seenTiles` were removed, since real traversal output is already deduped. The test gives false confidence in the new dedup mechanism; it should use a real `Grid` (e.g. `new Grid(defineHex(...), [...])`) so that whatever it asserts reflects production behavior.

app/src/libs/combat/util.ts
MathiasGruber: Good catch — you are right. `Grid.traverse` already deduplicates by coord-string key, so `seenTiles`/`needsDedup`/`addAffectedTile` were dead code. Reverted `getAffectedTiles` to the inline `isValidMove` branches and removed the misleading wall-traversal comment. The real fix lives in `actions.ts`'s `${barrier.id}-${tagIndex}` dedup, which is unchanged. Commit: 9980115c9.

app/tests/libs/combat/getAffectedTiles.test.ts
MathiasGruber: Agreed — the mocked `traverse` returning a raw array did not exercise real `Grid` behavior, and the assertion was only validating the dead dedup path. Removed the test entirely. The genuine per-cast barrier dedup in `actions.ts` (`${barrier.id}-${tagIndex}`) stands on its own and is unaffected. Commit: 9980115c9.

MathiasGruber (approved):
Approved. All prior review threads resolved in code; no new logic, readability, DRY, or performance concerns surfaced on re-review. Latest /tnr-review-now run on commit 9980115c9 successfully reproduced the AOE-on-barrier scenario and confirmed no duplicate-hit symptom — the `${barrier.id}-${tagIndex}` dedup keyed across affected tiles behaves correctly. No SQL migrations in scope.

MathiasGruber (approved):
Approved. All prior review threads resolved; logic, DRY, and performance re-review surfaced no concerns; /tnr-review-now validated the fix in live combat.

MathiasGruber (approved):
Approved. All prior review threads resolved in code (including Phrosfire's barrier-multi-hit report, fixed via `barrierAttacks` Set keyed by `${barrier.id}-${tagIndex}` outside the affected-tile loop). No new logic, readability, DRY, or performance concerns surfaced on re-review. Latest /tnr-review-now run validated the fix in live combat against placed barriers.

MathiasGruber (approved):
LGTM

MathiasGruber (approved):
LGTM. All prior review threads remain addressed in code, no new logic/readability/DRY or performance concerns surfaced on re-review, and the latest /tnr-review-now run completed successfully with no in-scope findings.

MathiasGruber (approved):
LGTM.

MathiasGruber (approved):
LGTM

MathiasGruber (approved):
Approved. All prior review threads remain addressed in code (Phrosfire's barrier multi-hit report fixed by `barrierAttacks` Set keyed by `${barrier.id}-${tagIndex}` outside the affected-tile loop, `BARRIER_DAMAGE_TAG_TYPES` centralized in `combat/constants.ts` and used in actions.ts/process.ts/util.ts, dead `seenTiles` dedup correctly reverted since `Grid.traverse` already dedups). No new logic, readability, DRY, or performance concerns surfaced on re-review. Latest `/tnr-review-now` run on commit 9980115c9 validated live behavior in Arena combat (single cast → one action entry → barriers each hit once at correct HP) with no PR-scoped findings.

MathiasGruber (approved):

MathiasGruber (approved):
Approved. All prior review threads resolved, no new logic/DRY/readability or performance concerns surfaced on re-review, and the /tnr-review-now run from 2026-05-16 reproduced the fix in player-facing combat testing.

MathiasGruber (approved):

MathiasGruber (approved):

MathiasGruber (approved):
Reviewed. All previous feedback addressed, code is clean and the dedup fix is correct. Approving.

MathiasGruber (approved):
Approved. All prior review threads resolved in code (including Phrosfire's barrier multi-hit report, fixed via `barrierAttacks` Set keyed by `${barrier.id}-${tagIndex}` outside the affected-tile loop). No new logic, readability, DRY, or performance concerns surfaced on re-review. Latest successful /tnr-review-now validated the AOE_CIRCLE_SPAWN-on-barrier fix in live combat.

MathiasGruber (approved):

MathiasGruber (approved):
Approved. All prior review threads resolved in code (including Phrosfire's barrier multi-hit report, fixed via `barrierAttacks` Set keyed by `${barrier.id}-${tagIndex}` outside the affected-tile loop, and `BARRIER_DAMAGE_TAG_TYPES` centralized in `combat/constants.ts`). No new logic, readability, DRY, or performance concerns surfaced on re-review. Latest `/tnr-review-now` run on commit 9980115c9 validated the AOE_CIRCLE_SPAWN-on-barrier fix in live combat with no duplicate barrier hits. No SQL migrations in scope.

MathiasGruber (approved):

MathiasGruber (approved):

MathiasGruber (approved):

MathiasGruber (approved):

MathiasGruber (approved):
Approved. All previous review threads verified resolved; no new logic, readability, DRY, or performance concerns surfaced on re-review; /tnr-review-now ran on the current HEAD and reported a clean pass with no PR bugs.

MathiasGruber (approved):
Approved. All prior review threads remain addressed in code, including the Phrosfire/Neon barrier-multi-hit report (fixed via `barrierAttacks` Set keyed by `${barrier.id}-${tagIndex}` outside the affected-tile loop) and `BARRIER_DAMAGE_TAG_TYPES` centralized in `combat/constants.ts`. Re-review of logic, readability, DRY, and performance surfaced no in-scope concerns. The most recent `/tnr-review-now` run was inconclusive only due to preview-environment scope-coverage gaps (level-1 scaling making Howitzer Crash uncastable in the default arena, no valid barrier placement in the Academy Dummy battle) — not code defects. Prior successful runs validated the fix in live combat. No SQL migrations in scope.

MathiasGruber (approved):

MathiasGruber (approved):
Approved. All prior review threads remain addressed in code (including Phrosfire's barrier-multi-hit report, fixed via `barrierAttacks` Set keyed by `${barrier.id}-${tagIndex}` outside the affected-tile loop, and `BARRIER_DAMAGE_TAG_TYPES` centralized in `combat/constants.ts` used by actions.ts/process.ts/util.ts). No new logic, readability, DRY, or performance concerns surfaced on re-review. Latest /tnr-review-now run completed with only preview-environment scope gaps (level scaling, default arena dummy battle) — prior successful runs validated the AOE_CIRCLE_SPAWN + barrier behavior in live combat. No SQL migrations in scope.

MathiasGruber (approved):
Approved. All prior review threads remain addressed in code (including Phrosfire's barrier-multi-hit report, fixed via `barrierAttacks` Set keyed by `${barrier.id}-${tagIndex}` outside the affected-tile loop, and `BARRIER_DAMAGE_TAG_TYPES` centralized in `combat/constants.ts` used by actions.ts/process.ts/util.ts). Re-review surfaced no new logic, readability, DRY, or performance concerns. Latest /tnr-review-now run reported only out-of-scope preview-environment blockers (player level/chakra/stamina too low to cast the AOE_CIRCLE_SPAWN jutsu against a live barrier in preview), not code defects. No migrations.

MathiasGruber (approved):
Approved. All prior review threads remain addressed in code (Phrosfire's barrier-multi-hit fixed via `barrierAttacks` Set keyed by `${barrier.id}-${tagIndex}` outside the affected-tile loop; `BARRIER_DAMAGE_TAG_TYPES` centralized in `combat/constants.ts` and consistently used in actions.ts/process.ts/util.ts). No new logic, readability, DRY, or performance concerns surfaced on re-review. Latest /tnr-review-now run was inconclusive only due to preview-env scope limitations (level-1 player scaling, chakra/stamina insufficient for casting Howitzer Crash, Academy Dummy battle disallowing barrier placement) — not code defects; the prior successful run validated AOE_CIRCLE_SPAWN dedup via dedicated probe jutsus with an Approve recommendation. No SQL migrations.

MathiasGruber (approved):

MathiasGruber (approved):
Approved. All prior review threads resolved and verified in current code (including Phrosfire’s multi-hit barrier report — barrier dedup key now keyed by `${barrier.id}-${tagIndex}` correctly prevents repeat hits on a single barrier across AOE tiles). No new logic, readability, DRY, or performance concerns surfaced on re-review. Latest /tnr-review-now ran after HEAD and produced no code-level findings.

app/src/libs/combat/actions.ts
```diff
@@ -825,11 +825,11 @@ export const insertAction = (info: {
               }
             }
             // Extra: If no target, check if there is a barrier & apply damage only
-            if (["damage", "pierce"].includes(tag.type)) {
+            if (BARRIER_DAMAGE_TAG_TYPES.has(tag.type)) {
```
MathiasGruber, line 829: Outside this diff hunk, but related to the barrier hot path: at line 730 above, `getBarriersBetween` is invoked here once per affected tile, even when the battle has zero barriers on the field. Internally it runs `aStar.getShortestPath(userHex, tile)` for each target, which performs A* pathfinding on the first call per (origin, target) pair (the `PathCalculator` cache only helps on repeat lookups within the same action). For a large AOE with N affected tiles and an empty barrier set, that is N wasted A* searches plus the path walking and `findBarrier` scans inside `getBarriersBetween`, all producing `{ barriers: [], totalAbsorb: 0 }`.

Since `totalAbsorb` is 0 and `barriers` is empty whenever no barriers exist on the battlefield, the entire call can be skipped with a single up-front check before the loop, e.g.:

```ts
const hasBarriers = groundEffects.some((g) => g.type === "barrier");
// ...
affectedTiles.forEach((tile) => {
  const { barriers, totalAbsorb } = hasBarriers
    ? getBarriersBetween(actorId, aStar, groundEffects, userHex, tile)
    : { barriers: [] as BattleEffect[], totalAbsorb: 0 };
  // ...
});
```

Non-barrier fights are the common case; this removes per-tile A* pathfinding from `performAction` on AOE actions without changing behavior.

MathiasGruber (commented):
Performance review

app/src/libs/combat/actions.ts
MathiasGruber: Good catch — added an up-front `hasBarriers = groundEffects.some((g) => g.type === "barrier")` check and gated `getBarriersBetween` on it, falling back to an empty `{ barriers: [], totalAbsorb: 0 }` constant when no barriers are on the field. Non-barrier AOEs (the common case) now skip the per-tile A\* pathfinding entirely. Pushed in 4b698ec2b.

MathiasGruber (approved):
Reviewed: prior feedback all addressed, no logic/readability/DRY/perf findings on the current diff, and the TNR live-preview pass on AOE_CIRCLE_SPAWN confirms the dedupe path. Blast Shield spawning multiple barrier ground effects is the ability's intended AOE behavior — the per-tile barrier-multi-hit bug from #1228 was already verified fixed in the prior live-preview run.

MathiasGruber (approved):
Approved. All 17 prior review threads verified addressed in code. No logic, readability, DRY, or performance concerns introduced by the diff. tnr-review-now findings (Blast Shield multi-barrier observation, checkout-mismatch and scenario-coverage notes) reviewed and confirmed false positives — the ring-spawn count is the intended behavior of AOE_CIRCLE_SPAWN jutsu, unrelated to the per-cast barrier/tile dedup this PR fixes.

MathiasGruber (approved):
Approved. All previous review threads resolved; no new logic, readability, DRY, or performance concerns surfaced on re-review.

MathiasGruber (approved):
LGTM

MathiasGruber (approved):
Approved.

MathiasGruber (approved):
Approving — all prior review feedback verified addressed, logic/perf re-review found nothing, latest /tnr-review-now Fail was a false positive (ring-of-barriers from Blast Shield is intended AOE_CIRCLE_SPAWN behavior, not a dedup bug).

MathiasGruber (approved):
All previous feedback verified addressed in current code. Reviews for logic/readability/DRY and performance found no actionable concerns. Latest /tnr-review-now report's lone 'fail' on Blast Shield is a false positive — multi-tile barrier spawning is the intended behavior of that jutsu (AOE_WALL_SHOOT with barrier tag, target INHERIT). No SQL migrations included.

MathiasGruber (approved):

MathiasGruber (approved):
All review threads resolved. Logic, readability, DRY, performance, and prior /tnr-review-now feedback reviewed — no outstanding concerns. The Blast Shield observation in the latest /tnr-review-now is by design (AOE_CIRCLE_SPAWN spawns a radius-1 spiral of barrier tiles). Approving.

MathiasGruber (approved):
Verified prior feedback addressed, logic/readability/DRY clean, performance clean, /tnr-review-now findings evaluated as false positives. Approving.

MathiasGruber (approved):
LGTM

MathiasGruber (approved):
LGTM

MathiasGruber (commented):
Logic / readability / DRY review: no actionable findings beyond those already addressed in the existing review thread.

MathiasGruber (approved):
All checks pass, no unresolved threads, prior feedback verified addressed, logic/readability/DRY/performance reviews clean, /tnr-review-now findings evaluated as non-issues.

MathiasGruber (approved):
Looks good. Prior feedback fully addressed, no remaining concerns from logic/readability/DRY, performance, or live-preview review.

MathiasGruber (approved):
All review feedback addressed. Code looks clean — barrier dedup correctly keyed by tagIndex, hot-path A* gated on hasBarriers, BARRIER_DAMAGE_TAG_TYPES consistently applied. TNR review pass on the AOE_CIRCLE_SPAWN probe path.

MathiasGruber (approved):

MathiasGruber (approved):

MathiasGruber (approved):
LGTM

MathiasGruber:
AoE were using A* on the backend as well, which was an issue for spawning-like tags. Fixed manually

PR #1326, opened 2026-04-13
Feature/item variants

### LINKED ISSUE #1304: Item Variants
Item Variants is a feature that will allow content to add multiple images for an item and allow players to select different looks for their items. 

Item Variant will be a toggle in the Item Creation menu when ticked, Content can add up to 7 Variants. 
Each Variant will need:

1. Image
2. Cost(Ryo,Prestige,Silver,Rep, Variant Token)

Players can select their Variants when they select the item in their inventory. The interface will be in a modal view where player can select the image to see the preview of it, once the image is bought or a variant token is used to unlock the variant the player can select it. 

Once the image is selected, this is the image that will show in the players Inventory 

VariantTag

Similar to the Reskin Token. This is a special that can be placed on an item to allow a palyer to unlock Variants that require tokens.

ifroshty:
Pull Request

  Item Variants — Closes #1304

  This implements the Item Variants feature, allowing content creators to attach up to 7 cosmetic variants (alternate images) to any item. Players
  can browse, purchase, and select variants from their inventory.

  ---
  What was implemented

  Database
  - New ItemVariant table storing variant definitions per item (image, name, cost type, cost, display order)
  - New UserItemVariant table tracking per-user variant unlocks (keyed by userId + variantId unique constraint)
  - activeVariantId nullable column on UserItem to track the selected variant per stack
  - Two migration files generated

  Backend (tRPC — routers/item.ts)
  - getItemVariants — list variants for an item (output-typed for OpenAPI compatibility)
  - getUserUnlockedVariants — fetch which variants the current user has unlocked for an item
  - upsertItemVariant / deleteItemVariant — admin CRUD (requires canChangeContent role); delete cascades to unlock records and clears
  activeVariantId on any affected user items
  - purchaseVariant — deducts currency (Ryo / Rep / Seichi Silver / Village Prestige) with a compare-and-swap guard; deduct-first pattern to prevent
   double-spend
  - selectVariant — sets activeVariantId on a user item; validates the variant belongs to that item
  - consumeVariantToken — consumes a UnlockItemVariantTag consumable item and unlocks a token-gated variant; CAS guard on token quantity

  Validators
  - ItemVariantValidator Zod schema (validators/item.ts)
  - UnlockItemVariantTag effect tag added to AllTags in validators/combat.ts — marks a consumable item as a Variant Token

  Constants
  - VARIANT_COST_TYPES — ["MONEY", "REPUTATION", "SEICHI_SILVER", "VILLAGE_PRESTIGE", "VARIANT_TOKEN"]
  - MAX_ITEM_VARIANTS = 7

  Admin UI (/manual/item/edit/[itemid])
  - ItemVariantsEditor section below the main item form
  - Lists existing variants with image preview, cost info, edit and delete (with confirmation dialog) buttons
  - Inline "Add / Edit Variant" form with validation error display on all fields

  Player Inventory UI (/items)
  - "Variants" button appears on any item that has variants configured
  - ItemVariantModal shows: base look (always free), each variant with its image and unlock state
    - Unlocked → "Select" / "Active" indicator
    - Currency cost → "Buy: X COST_TYPE" button
    - Token-gated → "Use Token" (uses first available Variant Token from inventory) or "Need Variant Token" if none owned

  Display helper
  - getItemDisplayImage(source) added to libs/combat/util.ts — returns the active variant image if set, otherwise falls back to the base item image

  ---
  Why

  Implements studie-tech/TheNinjaRPG#1304 — a monetisation feature allowing cosmetic customisation of items without affecting game balance. The
  variant unlock system supports multiple currency types including a dedicated Variant Token consumable.

  ---
  Breaking changes

  None. All new columns are nullable, all new tables are additive, and existing item/inventory flows are unchanged. The fetchUserItems helper remains lean (no variant data loaded) for all combat/non-inventory paths.

  ---
  Known limitations / follow-up work

  - Variant images are not shown during active battles — BattleUserItem does not carry activeVariantId. This is a deliberate deferral; the getItemDisplayImage helper is ready for when that is added.
  - "Prestige" currency maps to userData.villagePrestige — confirm with team if a separate player prestige field is intended.
  - Variant image upload in the admin form is a plain URL field (paste from UploadThing). A full file-picker widget is a follow-up improvement.

  ---
  License

  By making this pull request, I confirm that I have the right to waive copyright and related rights to my contribution, and agree that all
  copyright and related rights in my contributions are waived, and I acknowledge that the Studie-Tech ApS organization has the copyright to use and
  modify my contribution for perpetuity.

## Summary by CodeRabbit

* **New Features**
  * Introduce item variants system allowing users to unlock, purchase, and select different versions of items
  * Add variant selection interface in inventory and character screens
  * Enhance combat system to display and utilize active item variants
  * Expand item editor with variant creation and management capabilities
  * Increase image upload size limit from 64KB to 512KB

* **Tests**
  * Add test coverage for variant display logic and validation


app/src/app/items/page.tsx
```diff
@@ -1029,6 +1070,11 @@ const Character: React.FC<CharacterProps> = (props) => {
             <ItemWithEffects
               item={{
                 ...useritem.item,
+                image: getItemDisplayImage(useritem as UserItemWithVariants),
+                name: getItemDisplayName(useritem as UserItemWithVariants),
+                description: getItemDisplayDescription(
+                  useritem as UserItemWithVariants,
+                ),
```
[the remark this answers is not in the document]
ifroshty: can you confirm this is resolved?

app/src/app/manual/item/edit/[itemid]/page.tsx
```diff
@@ -197,3 +216,310 @@ const SingleEditItem: React.FC<SingleEditItemProps> = (props) => {
...
+        cost: 0,
+        order: (variants?.length ?? 0) + 1,
+        description: "",
+        battleDescription: "",
+      });
+    }
+    // variants?.length intentionally omitted: order is computed once at form-open time,
+    // not on every background refetch
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [editingVariant, form]);
```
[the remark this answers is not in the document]
Comment (ifroshty) [reply app/src/app/manual/item/edit/[itemid]/page.tsx]: can you confirm this is resolved?

app/src/server/api/routers/item.ts
```diff
@@ -429,7 +447,372 @@ export const itemRouter = createTRPCRouter({
...
+        // Roll back the unlock we just inserted
+        await ctx.drizzle
+          .delete(userItemVariant)
+          .where(
+            and(
+              eq(userItemVariant.userId, ctx.userId),
+              eq(userItemVariant.variantId, input.variantId),
+            ),
+          );
+        return errorResponse("Token item was modified concurrently — please try again");
```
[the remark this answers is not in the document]
ifroshty: can you confirm this is resolved?

app/src/validators/combat.ts
```diff
@@ -797,6 +797,13 @@ export const NonCombatGainSkill = z.object({
   skillId: z.string().prefault(""),
 });
 
+export const UnlockItemVariantTag = z.object({
+  ...BaseAttributes,
+  type: z.literal("unlockitemvariant").prefault("unlockitemvariant"),
+  description: msg("Unlocks a token-locked item variant when used on an item"),
+});
+export type UnlockItemVariantTagType = z.infer<typeof UnlockItemVariantTag>;
```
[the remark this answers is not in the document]
ifroshty: can you confirm this is resolved?

app/tests/libs/combat/util.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    expect(getItemDisplayBattleDescription(source)).toBe("%user swings a blade");
+  });
+
+  it("pre-resolved value takes priority when passed as flat source (BattleUserItem simulation)", () => {
+    // Once variantBattleDescription is stored on BattleUserItem and used directly
+    // in userItemToAction, it bypasses the helper entirely. This test simulates
+    // the flat-object path that action processing uses.
+    const preResolved = "%user swings the Crimson Blade";
+    const source = { battleDescription: preResolved };
+    expect(getItemDisplayBattleDescription(source)).toBe(preResolved);
```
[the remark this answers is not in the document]
ifroshty: can you confirm this is resolved?

app/src/server/api/routers/item.ts
```diff
@@ -1829,7 +2229,24 @@ export const fetchItemWithCraftingRequirements = async (
...
+
+export const fetchUserItemsWithVariants = async (
+  client: DrizzleClient,
+  userId: string,
+) => {
+  const useritems = await client.query.userItem.findMany({
+    where: and(eq(userItem.userId, userId)),
+    with: {
+      item: { with: { variants: true } },
+      imbuements: { with: { item: true } },
```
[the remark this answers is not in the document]
ifroshty: confirm this is resolved

MathiasGruber (commented):
Submitting pending review

---

MathiasGruber (commented):
## Review: Feature/item-variants

### Open threads

**Thread A — `deleteItemVariant` delete order (line 589)**
MathiasGruber's reply ("FK is not an issue in PlanetScale") is correct about FK enforcement, but the original CodeRabbit concern was actually about a race condition window: between Step 1 (clearing `userItemVariant` / nulling `activeVariantId`) and Step 2 (deleting `itemVariant`), a concurrent `purchaseVariant` or `consumeVariantToken` call could re-insert a `userItemVariant` row for the variant that is about to be deleted. This is an admin-only endpoint so real-world risk is very low, but the two-step sequential approach does leave a small window. Since `deleteItemVariant` is admin-only, this is acceptable as-is; just worth noting the thread as "addressed by intent, low risk".

**Thread B — `consumeVariantToken` using `fetchUserItemWithVariants` (line 767)**
MathiasGruber is correct: `consumeVariantToken` never inspects `tokenItem.item.variants` — it only reads `tokenItem.item.effects` (line 796) and `tokenItem.quantity` (lines 807/814/823). `fetchUserItemWithVariants` fetches an extra join (variants with ordered results) that is unused. Replace with `fetchUserItem` to avoid the unnecessary query work:

```ts
// before
fetchUserItemWithVariants(ctx.drizzle, ctx.userId, input.tokenUserItemId),

// after
fetchUserItem(ctx.drizzle, ctx.userId, input.tokenUserItemId),
```

**Thread C — `fetchUserItemWithVariants` missing imbuements (line 2288)**
MathiasGruber's question is valid. `fetchUserItemsWithVariants` (the bulk version, line 2263) already includes `imbuements: { with: { item: true } }`. The single-item `fetchUserItemWithVariants` should match that shape for consistency and to avoid surprises if future callers need imbuements:

```ts
export const fetchUserItemWithVariants = async (
  client: DrizzleClient,
  userId: string,
  userItemId: string,
) => {
  return await client.query.userItem.findFirst({
    where: and(eq(userItem.userId, userId), eq(userItem.id, userItemId)),
    with: {
      item: { with: { variants: { orderBy: (v, { asc }) => [asc(v.order)] } } },
      imbuements: { with: { item: true } },
    },
  });
};
```

---

### Logic — `unlockitemvariant` not in `nonCombatConsume` (`app/src/libs/item.ts` line 32)

`nonCombatConsume` lists every recognised effect type that makes a consumable usable outside combat. `unlockitemvariant` is **not** in that list, so calling the generic `consume` endpoint on a Variant Token item returns the opaque "Not consumable" error rather than a message directing the user to the dedicated `consumeVariantToken` flow.

This is safe (the item cannot be consumed the wrong way), but the error is unhelpful. The UI at `page.tsx:1239` correctly identifies Variant Token items and never shows the generic consume button for them, so the happy path is protected. Consider adding an explicit guard or a better error message in `consume` for items with only the `unlockitemvariant` effect.

---

### Minor — `purchaseVariant` refund arithmetic (line 691)

The refund block adds `variant.cost` back via `` sql`${userData.money} + ${variant.cost}` ``. A network retry could in theory trigger two refunds. The `onDuplicateKeyUpdate` on the variant insert (line 686) makes this path extremely unlikely. Low priority given it sits in an error-handling branch.


---

MathiasGruber:

- `app/drizzle/migrations/meta/_journal.json`, `app/drizzle/migrations/meta/0006_snapshot.json`, `app/drizzle/migrations/meta/0007_snapshot.json`: took `main`'s version.
- PR's original migrations (`0006_rapid_hiroim.sql`, `0007_mean_vector.sql`, `0008_giant_susan_delgado.sql`, `0009_little_sue_storm.sql`): deleted; migration regenerated after rebase as `0008_omniscient_black_queen.sql` containing the full item-variant schema (`ItemVariant` table, `UserItemVariant` table, `UserItem.activeVariantId` column, and the supporting indexes).
- All other files (`schema.ts`, `constants.ts`, combat/item routers and validators, libs, etc.) auto-merged cleanly.

app/src/server/api/routers/item.ts
```diff
@@ -434,6 +443,413 @@ export const itemRouter = createTRPCRouter({
...
+      return { success: true, message: "Variant deleted" };
+    }),
+  // Purchase a variant with in-game currency
+  purchaseVariant: protectedProcedure
+    .input(z.object({ variantId: z.string() }))
+    .output(baseServerResponse)
+    .mutation(async ({ ctx, input }) => {
+      // Query — ownership check joins userItem→itemVariant in parallel (variantId known upfront)
+      const [userResult, variant, ownershipRows, existingUnlock] = await Promise.all([
+        fetchUpdatedUser({ client: ctx.drizzle, userId: ctx.userId }),
```
MathiasGruber: `fetchUpdatedUser` is unnecessarily heavy here — internally it runs ~7 parallel queries (achievements, settings, user with bloodline/clan/village/structures/relationships/sectors/anbu, unvoted polls, active wars, shrine battles, active raids) plus full stat regeneration, just to read `isBanned` and four currency fields. The CAS `gte()` guard on the deduct already protects against stale balances. Swap to the lightweight `fetchUser` (used by all blackmarket purchase mutations) to cut a large fixed cost off every variant purchase. Same applies to `consumeVariantToken` at line 769.

MathiasGruber (commented):
Performance review pass.

app/src/app/items/page.tsx
```diff
@@ -872,6 +907,14 @@ const Backpack: React.FC<BackpackProps> = (props) => {
         onRepairItem={mutateRepairItem}
         isPending={isUsingRepairItem}
       />
+      {/* Variant Modal */}
+      {variantItem && (
+        <ItemVariantModal
+          userItem={variantItem}
+          allUserItems={useritems}
```
MathiasGruber: Token-gated variants still only inspect the current Backpack tab slice. `allUserItems={useritems}` here is the filtered list passed into Backpack (e.g. only normal items where `equipped === "NONE"`), so `variantTokens` will miss Variant Tokens stored under a different tab (event/materials) or in an equipped slot. Pass the page-level `userItems` (the full inventory from `getUserItemsWithVariants`) here instead, the way the Character modal already does at line 1150.

MathiasGruber (commented):
Re-opening prior review feedback that does not appear fully addressed in the current code.

app/src/server/api/routers/item.ts
```diff
@@ -434,6 +443,413 @@ export const itemRouter = createTRPCRouter({
...
+        }
+      }
+
+      // Insert unlock — pre-check above ensures this is not a duplicate.
+      // On unexpected DB error, refund if currency was deducted.
+      try {
+        await ctx.drizzle
+          .insert(userItemVariant)
+          .values({ id: nanoid(), userId: ctx.userId, variantId: input.variantId })
+          .onDuplicateKeyUpdate({ set: { id: sql`id` } });
```
MathiasGruber: Race condition: two concurrent `purchaseVariant` calls can both pass the `existingUnlock` pre-check and both deduct currency, but only one row is inserted (the other hits the unique index and is silently swallowed by `onDuplicateKeyUpdate({ set: { id: sql`id` } })`). The user is double-charged for a single unlock. Check `insertResult.rowsAffected` and refund if 0, mirroring the catch branch.

app/src/server/api/routers/item.ts
```diff
@@ -434,6 +443,413 @@ export const itemRouter = createTRPCRouter({
...
+              : gte(userData.villagePrestige, variant.cost);
+
+      const currencySet =
+        variant.costType === "MONEY"
+          ? { money: sql`${userData.money} - ${variant.cost}` }
+          : variant.costType === "REPUTATION"
+            ? { reputationPoints: sql`${userData.reputationPoints} - ${variant.cost}` }
+            : variant.costType === "SEICHI_SILVER"
+              ? { seichiSilver: sql`${userData.seichiSilver} - ${variant.cost}` }
+              : { villagePrestige: sql`${userData.villagePrestige} - ${variant.cost}` };
```
MathiasGruber: DRY: the same MONEY/REPUTATION/SEICHI_SILVER/VILLAGE_PRESTIGE ternary chain is repeated 4 times in this mutation (pre-check, `currencyWhere`, `currencySet`, and `refundSet`). Extract a single helper keyed off `costType` that returns `{ balance, where, decrementSet, incrementSet }` so a new currency only needs to be added in one place.

app/src/libs/combat/util.ts
```diff
@@ -80,6 +80,108 @@ import type {
...
+  }
+  if ("item" in source) return source.item.description;
+  return source.description;
+};
+
+/**
+ * Returns the battle description for a user item, preferring the active variant battleDescription if set.
+ * Falls back to the base item battleDescription.
+ */
+export const getItemDisplayBattleDescription = (
```
MathiasGruber: `getItemDisplayBattleDescription` is unused outside of its own test file. Combat call sites in `actions.ts` inline the equivalent `useritem.variantBattleDescription ?? item.battleDescription` directly, and no other module imports this helper. Either delete this function (and the associated tests) or refactor `userItemToAction` to call it.

MathiasGruber:

**Conflicts resolved:**
- **Migrations**: The PR carried several migrations (`0007`/`0008`/`0009` across its history) that collided with new migrations merged into `main` (`0007`–`0010`). Per the migration-conflict policy, all of the PR's own migration files were dropped, the migration metadata (`_journal.json` / snapshots) was reset to match `main`, and a single fresh migration (`0011_sparkling_thaddeus_ross.sql`) was regenerated via `make makemigrations`. It cleanly creates the `ItemVariant` and `UserItemVariant` tables and adds `UserItem.activeVariantId` plus the associated indexes.
- **schema.ts / combat util**: Auto-merged; the PR's item-variant schema and logic are preserved on top of the latest `main`.

ifroshty:
@greptileai

app/src/server/api/routers/item.ts
```diff
@@ -434,6 +443,382 @@ export const itemRouter = createTRPCRouter({
...
+  // Consume a Variant Token item to unlock a VARIANT_TOKEN-gated variant
+  consumeVariantToken: protectedProcedure
+    .input(z.object({ tokenUserItemId: z.string(), variantId: z.string() }))
+    .output(baseServerResponse)
+    .mutation(async ({ ctx, input }) => {
+      // Query — ownership check joins userItem→itemVariant in parallel (variantId known upfront)
+      const [user, tokenItem, variant, ownershipRows, existingUnlock] =
+        await Promise.all([
+          fetchUser(ctx.drizzle, ctx.userId),
+          fetchUserItemWithVariants(ctx.drizzle, ctx.userId, input.tokenUserItemId),
```
MathiasGruber: The token's userItem only needs effects, quantity, and base item fields here — the code never reads tokenItem.item.variants. Using fetchUserItemWithVariants triggers an unnecessary join to ItemVariant on every token consumption. Swap to fetchUserItem (no variant join) to save data on the hot path.

MathiasGruber (commented):
Performance review

MathiasGruber (commented):

app/src/server/api/routers/item.ts
```diff
@@ -434,6 +443,382 @@ export const itemRouter = createTRPCRouter({
...
+          where: eq(itemVariant.id, input.variantId),
+        }),
+        ctx.drizzle
+          .select({ id: userItem.id })
+          .from(userItem)
+          .innerJoin(itemVariant, eq(userItem.itemId, itemVariant.itemId))
+          .where(
+            and(eq(userItem.userId, ctx.userId), eq(itemVariant.id, input.variantId)),
+          )
+          .limit(1),
```
MathiasGruber: DRY: this 8-line ownership join is byte-for-byte identical to the one in `consumeVariantToken` (lines 743-750). Both are checking "does the caller own at least one userItem whose itemId matches the variant's itemId". Extracting a small helper like `fetchVariantOwnership(client, userId, variantId)` keeps the two mutations from drifting apart if the ownership rule changes (e.g. excluding items `storedAtHome`, `isInAuction`, or items with `craftingFinishedAt` in the future — none of which are currently filtered out here).

app/src/server/api/routers/item.ts
```diff
@@ -434,6 +443,382 @@ export const itemRouter = createTRPCRouter({
...
+            where: eq(itemVariant.id, input.variantId),
+          }),
+          ctx.drizzle
+            .select({ id: userItem.id })
+            .from(userItem)
+            .innerJoin(itemVariant, eq(userItem.itemId, itemVariant.itemId))
+            .where(
+              and(eq(userItem.userId, ctx.userId), eq(itemVariant.id, input.variantId)),
+            )
+            .limit(1),
```
MathiasGruber: Same ownership join as `purchaseVariant` (lines 606-613). Worth pulling into a shared helper so both mutations apply the exact same filters. Right now neither one excludes items that are `storedAtHome`, `isInAuction`, or still crafting — that may or may not be intended, but at least it would be enforced in one place.

app/src/app/items/page.tsx
```diff
@@ -661,6 +683,7 @@ const Backpack: React.FC<BackpackProps> = (props) => {
   const items = useritems?.map((useritem) => ({
     ...useritem.item,
     ...useritem,
+    image: getItemDisplayImage(useritem as UserItemWithVariants),
```
MathiasGruber: The grid only overrides `image` but not `name`/`description`. `ActionSelector` consumes `item.name` for the `alt` attribute (CombatActions.tsx:299) and for the visible label when `showLabels` is enabled, so a user with an active variant sees the variant image but the base name in the tooltip / alt text. The detail modal on line 740-742 already applies all three getters; consider doing the same here for consistency:

```ts
const items = useritems?.map((useritem) => ({
  ...useritem.item,
  ...useritem,
  image: getItemDisplayImage(useritem as UserItemWithVariants),
  name: getItemDisplayName(useritem as UserItemWithVariants),
  description: getItemDisplayDescription(useritem as UserItemWithVariants),
}));
```

Same issue at line 953 in `Character`.

app/src/app/items/page.tsx
```diff
@@ -893,13 +940,17 @@ const Character: React.FC<CharacterProps> = (props) => {
...
+    undefined,
+  );
 
   // The item on the current slot
 
   // Collapse UserItem and Item
   const items = useritems?.map((useritem) => ({
     ...useritem.item,
     ...useritem,
+    image: getItemDisplayImage(useritem as UserItemWithVariants),
```
MathiasGruber: Mirror of the issue at line 686: `image` is variant-aware but `name`/`description` aren't. The character-screen tile shows the active variant's sprite but its `alt` text (and any hover label) still reads the base item name. Apply `getItemDisplayName` and `getItemDisplayDescription` here too so the inventory grid matches the detail modal at line 1083-1086.

ifroshty:
@greptileai

MathiasGruber:

**Conflicts resolved:**
- The only conflicts were in the Drizzle migration files (`app/drizzle/migrations/meta/_journal.json` and the branch's migration `.sql`). `main` had advanced to migration `0011` while this branch carried its own migration created against an older base, causing journal index collisions.
- Following the standard procedure, the branch's stale migration files were dropped, the migrations directory was reset to match `main`, and a fresh migration (`0012_secret_lord_hawal.sql`) was regenerated from `schema.ts` via `make makemigrations`. It cleanly creates the `ItemVariant` and `UserItemVariant` tables, adds `UserItem.activeVariantId`, and the associated indexes — no other schema changes.

No application/source code conflicts occurred; all 51 commits replayed cleanly. The branch is now up to date with `main`.

MathiasGruber:
Rebased this branch onto the latest `main` and force-pushed the result.

**Conflicts encountered:** only in the Drizzle migration metadata — `migrations/meta/_journal.json` and the `0012` snapshot — because both `main` and this branch introduced a new `0012` migration.

**Resolution:** dropped this branch's migration (`0011_sparkling_thaddeus_ross` / `0012_secret_lord_hawal`), reset the migration meta to `main`'s state (latest = `0012_little_master_chief`), and regenerated a fresh migration from the schema with `make makemigrations`. The item-variant changes now land as `0013_yummy_cerise` (creates `ItemVariant`, `UserItemVariant`, and adds `UserItem.activeVariantId` with supporting indexes).

No application/source conflicts occurred. `tsc --noEmit` passes locally after the rebase.

app/drizzle/schema.ts
```diff
@@ -2476,6 +2549,7 @@ export const userItem = mysqlTable(
     storedAtHome: boolean("storedAtHome").default(false).notNull(),
     craftingFinishedAt: datetime("craftingFinishedAt", { mode: "date", fsp: 3 }),
     isInAuction: boolean("isInAuction").default(false).notNull(),
+    activeVariantId: varchar("activeVariantId", { length: 191 }),
```
MathiasGruber, line 2552: Since `activeVariantId` lives on `UserItem`, can we clear or revalidate it when the row changes owner? `completeAuctionInternal` currently transfers the auctioned `UserItem` by setting `userId` to the buyer and leaves `activeVariantId` intact. That lets a sold item carry the seller’s selected variant into the buyer’s inventory even if the buyer has not unlocked that variant, and the display/combat paths trust `ui.activeVariantId`. I think auction completion should either null this field or preserve it only when the buyer has the matching `UserItemVariant` unlock.

app/src/server/api/routers/item.ts
```diff
@@ -454,6 +493,382 @@ export const itemRouter = createTRPCRouter({
...
+          fetchVariantOwnership(ctx.drizzle, ctx.userId, input.variantId),
+          ctx.drizzle.query.userItemVariant.findFirst({
+            where: and(
+              eq(userItemVariant.userId, ctx.userId),
+              eq(userItemVariant.variantId, input.variantId),
+            ),
+          }),
+        ]);
+
+      // Guards
```
MathiasGruber: Can we add the same availability checks here before consuming the token? This currently rejects home storage and empty stacks, but it still allows a token to be consumed while the user is not `AWAKE`, and while the token stack is still crafting. It also does not reject `isInAuction`, so the direct mutation can burn/delete a token row that is currently tied to an auction. Since `consumeUserItemAtomically` only guards on `id`, `userId`, and `quantity`, I think those state checks need to happen before this call.

app/src/server/api/routers/item.ts
```diff
@@ -454,6 +493,382 @@ export const itemRouter = createTRPCRouter({
...
+        );
+        if (!variantBelongsToItem) {
+          return errorResponse("Variant does not belong to this item");
+        }
+      }
+
+      // Mutate
+      const updateResult = await ctx.drizzle
+        .update(userItem)
+        .set({ activeVariantId: input.variantId })
```
MathiasGruber, line 775: Once the active variant is stored on each `UserItem` stack, the stack helpers need to treat it as part of the stack state. `splitItemStack` creates the new row without copying `activeVariantId`, so splitting a selected stack drops the cosmetic from the split-off stack. `mergeStacksBucketKey` also groups only by `storedAtHome` and `equipped`, so two stacks of the same item with different active variants can be merged and one selected variant silently wins. Can we copy `activeVariantId` on split and include it in the merge payload/bucket key?

app/src/server/api/routers/item.ts
```diff
@@ -454,6 +493,382 @@ export const itemRouter = createTRPCRouter({
...
+        .where(and(eq(userItem.id, input.userItemId), eq(userItem.userId, ctx.userId)));
```
MathiasGruber: This still validates the unlock and item/variant relationship before the update, but the write itself only checks `userItem.id` and `userItem.userId`. If the unlock or variant is deleted between the precheck and this update, the row can still be set to an `activeVariantId` that is no longer valid for this user/item. Can we make the update conditional on the variant/unlock still existing, or re-check immediately after the write and clear/fail if it no longer does?

app/src/server/api/routers/item.ts
```diff
@@ -59,13 +75,19 @@ import { getRandomElement } from "@/utils/array";
 import { calculateContentDiff } from "@/utils/diff";
 import { fedItemLoadouts } from "@/utils/paypal";
 import { canAwardReputation, canChangeContent } from "@/utils/permissions";
+import sanitize, { sanitizeVariantText } from "@/utils/sanitize";
```
MathiasGruber: Leftover from the variant refactor — only `sanitizeVariantText` is used in this file now, so the default `sanitize` import is dead (biome flags it). Drop it to `import { sanitizeVariantText } from "@/utils/sanitize";`.

app/src/app/items/page.tsx
```diff
@@ -26,7 +26,14 @@ import {
 import { Input } from "@/components/ui/input";
 import { Label } from "@/components/ui/label";
 import { COST_EXTRA_ITEM_SLOT, IMG_EQUIP_SILHOUETTE } from "@/drizzle/constants";
-import type { Item, ItemSlot, UserItem, UserItemWithRelations } from "@/drizzle/schema";
+import type {
+  Item,
+  ItemSlot,
+  ItemVariant,
```
MathiasGruber: `ItemVariant` isn't referenced anywhere in this file — the variant-aware paths go through `UserItemWithVariants` instead — so this import is unused and biome warns on it. Please drop it from the type import.

app/src/libs/combat/util.ts
```diff
@@ -88,6 +88,80 @@ import type {
...
 
+// =============================================================================
+// ITEM DISPLAY HELPERS
+// =============================================================================
+
+/**
+ * Returns the display image for a user item, preferring the active variant image if set.
+ * Falls back to the base item image.
+ */
+export const getItemDisplayImage = (
```
MathiasGruber: For jutsu reskins we resolve the cosmetic in a single `getReskinnedUserJutsu(userJutsu)` that returns the fully merged object. Here we instead have three parallel helpers (`getItemDisplayImage`/`Name`/`Description`) that each redo the same active-variant lookup, and every call site then spreads all three by hand (and casts `as UserItemWithVariants`). Can we collapse these into one `applyActiveVariant(userItem)` that returns the item with name/image/description already overridden? It keeps the two cosmetic systems parallel, and it would fix the grid maps that currently only override `image` in one place rather than per call site.

app/src/server/api/routers/item.ts
```diff
@@ -454,6 +493,382 @@ export const itemRouter = createTRPCRouter({
...
+      if (user.isBanned) return errorResponse("You are banned");
+      if (!tokenItem) return errorResponse("Token item not found");
+      if (tokenItem.storedAtHome) {
+        return errorResponse("Fetch the Variant Token from home storage first");
+      }
+      if (!variant) return errorResponse("Variant not found");
+      if (variant.costType !== "VARIANT_TOKEN") {
+        return errorResponse("This variant does not require a Variant Token");
+      }
+      const hasVariantTokenEffect = tokenItem.item.effects.some(
```
MathiasGruber, line 860: This only checks that the token *has* an `unlockitemvariant` effect — there's no binding between a given token and a specific item or variant, so any Variant Token unlocks any token-gated variant on any item the user owns. If tokens are meant to be fully generic that's fine, but let's make that explicit (a code comment here at least); if we ever want per-item or per-variant tokens, there's nothing on the token to key off of yet.

app/src/server/api/routers/item.ts
```diff
@@ -1876,6 +2362,33 @@ export const fetchUserItem = async (
...
+  return await client.query.userItem.findFirst({
+    where: and(eq(userItem.userId, userId), eq(userItem.id, userItemId)),
+    with: { item: { with: { variants: { orderBy: (v, { asc }) => [asc(v.order)] } } } },
+  });
+};
+
+// Does the caller own at least one userItem whose itemId matches the variant's
+// itemId? Shared by purchaseVariant and consumeVariantToken so both apply the
+// exact same ownership filter. Returns an array; callers check `.length`.
+export const fetchVariantOwnership = async (
```
MathiasGruber, line 2429: Ownership is checked here at purchase/consume time, but the unlock itself lives on `UserItemVariant` keyed by `(userId, variantId)` and outlives the item — a player can buy a variant, sell every copy of the item, later re-acquire it, and the variant is still unlocked. I think "you keep what you paid for" is the behaviour we want, just flagging it so it's a deliberate choice rather than something that surprises us later.

app/drizzle/schema.ts
ifroshty: Addressed — completeAuctionInternal now sets activeVariantId: null on ownership
  transfer to the buyer. Went with null over preserve-if-unlocked to avoid an extra
  lookup in the auction-completion path; the buyer can re-select the variant if they
  hold the unlock.

app/src/server/api/routers/item.ts
ifroshty: Done — splitItemStack now copies activeVariantId onto the split-off row, and
  mergeStacksBucketKey includes activeVariantId (plus the bucket/eligible row types)
  so stacks with different active variants never merge. The per-bucket UPDATE/DELETE
  don't need it in their WHERE since same-variant stacks share a bucket; added a
  comment noting the concurrent-selectVariant case as a pre-existing PlanetScale
  (no-transaction) limitation rather than a regression.

app/src/server/api/routers/item.ts
ifroshty: Confirmed — "you keep what you paid for" is intended, matching the jutsu Reskin
  model. Documented it on fetchVariantOwnership: ownership is only checked at
  purchase/consume time, and the (userId, variantId) unlock intentionally persists
  even after selling every copy of the item

PR #1327, opened 2026-04-14
Auction house qol

Phrosfire:
# Pull Request

**Improvements:**
- Cleaner UI
- Added multiple tabs for better organization
- Added a level range to bid/purchase the items (Neon request)
- Added the auction house to syndicate
- If an auction contains a stack size larger than one, it displays the cost per unit as well.
- Notification is sent to the highest bidder if they were outbid or the item was bought out by someone else.
- Allow sellers to cancel their auction if nobody has bid on it yet.
- The auctions tab displays auctions in your level range (Display all for staff regardless)
- Direct Sales tab only displays items where you are the buyer or seller (Display all for staff regardless)
- My auctions displays all active auctions you have open
- My bids displays all open auctions that you have bid on.  Also displays if you are winning or not.

## License

## Summary by CodeRabbit

* **New Features**
  * Seller search and filtering for auction listings
  * Tabs to view your own auctions and active bids separately
  * Bidder level eligibility restrictions on auction listings
  * Cancel active listings with no bids
  * Visual improvements including item rarity indicators and stack quantities
  * Bid status badges showing if you're winning, outbid, or have lost

[Review Change Stack](https://app.coderabbit.ai/change-stack/studie-tech/TheNinjaRPG/pull/1327)


Phrosfire:
Can my permissions be updated so that my PRs can be deployed for testing on vercel?

Phrosfire:
The auction houses are available in the villages on live from a previous update, the sql change must not have been pushed in into the dev server. I did all my testing while in wake Island when building this update.

MathiasGruber (commented):
## Code Review — Auction House QoL

Three findings from reading the changed files:

---

### P2 — `validators/auction.ts` line 76: `.refine()` uses `error:` key instead of `message:`

Zod's `.refine()` only recognises `message` (and `path`) in its options object. The `error` key is silently ignored, so the buyout-price validation currently fires a generic Zod error message rather than the intended human-readable one.

```ts
// Current — broken, error key is ignored by Zod
.refine(
  (data) => {
    if (data.buyoutPrice) {
      return data.buyoutPrice > data.startingPrice;
    }
    return true;
  },
  {
    path: ["buyoutPrice"],
    error: "Buyout price must be higher than the starting price",  // ← wrong key
  },
);

// Fix
  {
    path: ["buyoutPrice"],
    message: "Buyout price must be higher than the starting price",
  },
```

The three other `.refine()` calls in the same file already use `message:` correctly, so this is a straightforward one-line fix.

---

### P2 — `routers/auction.ts` `createAuctionListing`: target-user fetch is sequential after the parallel initial fetch

The `targetUserId` validation block (lines 401–414) runs a separate `await ctx.drizzle.query.userData.findFirst(...)` only for `DIRECT` listings. This is a third sequential round-trip that could be collapsed into the opening `Promise.all` (with a conditional include) or simply moved earlier. Given CLAUDE.md's strict "minimise DB round-trips" policy this should be parallelised with the existing `[userItemData, user]` fetch when `listingType === "DIRECT"`:

```ts
const [userItemData, user, targetUserData] = await Promise.all([
  ctx.drizzle.query.userItem.findFirst({ ... }),
  targetUserId
    ? ctx.drizzle.query.userData.findFirst({
        where: eq(userData.userId, targetUserId),
        columns: { userId: true },
      })
    : Promise.resolve(undefined),
]);
```

Then remove the sequential fetch block later in the mutation.

---

### P3 — `page.tsx` `getAuctionBidStanding`: "Outbid" badge shown even when the viewer has no bid

`getAuctionBidStanding` is only called for the `MY_BIDS` tab (which requires `onlyBidOn: true`), so in practice the user always has a bid on every displayed listing — the server-side join ensures this. The badge logic itself is therefore not broken in the current flow.

However, the function itself has a latent inaccuracy: for an active listing it returns `"outbid"` whenever the viewer is *not* the current leader, including the edge case where the viewer's bid has the same amount as the leader (a tied bid that lost ordering). If the function is ever reused outside the `MY_BIDS` tab the label would be misleading. Worth a guard or a comment clarifying the assumption.

---

### Confirmed fixed / not an issue

- **DIRECT listing privacy bypass (thread about omitting `listingType`)**: Fixed. `applyUserListingPrivacy` is set to `true` whenever `!onlyMine && !onlyBidOn`, meaning the DIRECT privacy filter runs regardless of whether `listingType` is passed in the query params. The `OR` predicate (`ne(listingType, "DIRECT") OR sellerId = me OR targetUserId = me`) is applied at the SQL level via a join on the caller's row, so the bypass is closed.
- **`caller` parallelisation in `getAuctionListing`**: Already parallelised — `[listing, caller]` are fetched in a single `Promise.all` (lines 51–97).
- **Available-funds display for reputation listings**: Correct — `spendBalanceForListing` reads `userData.reputationPoints` when `currencyType === "REPUTATION"`, and `availableFunds` adds back the user's existing bid amount on top, so the "Available reputation" figure shown to the bidder is accurate.

---

MathiasGruber:

- PR's original migration `0007_misty_deathbird.sql`: deleted; migration regenerated after rebase as `0008_busy_dreaming_celestial.sql` adding `AuctionListing.bidderMinLevel`, `AuctionListing.bidderMaxLevel` and the corresponding index.
- Other files (`schema.ts`, `constants.ts`, auction router/validator, auctionhouse page, layout components) auto-merged cleanly.

app/src/server/api/routers/auction.ts
```diff
@@ -489,6 +659,28 @@ export const auctionRouter = createTRPCRouter({
...
+      if (existingBid) {
+        existingBid.amount = amount;
+      }
+
+      if (outbidUserId) {
+        const itemLabel = auction.userItem?.item?.name ?? "An auction lot";
+        const content = isBuyoutBid
+          ? `Someone bought out "${itemLabel}" — you were the previous high bidder.`
+          : `You were outbid on "${itemLabel}" in the auction.`;
+        await Promise.all([
```
MathiasGruber: This outbid-notification `Promise.all` runs as a separate sequential step after the bid-update `Promise.all` block above (the existing-bid branch at lines ~595-621 and new-bid branch at ~625-650). The two operations have no dependency on each other — `outbidUserId`, `content`, and `auction.userItem.item.name` are all computed beforehand, and the `notification` insert + `unreadNotifications` increment target a different user than the bidder. Folding the `notification` insert and the `unreadNotifications` increment into the same `Promise.all` as the bid insert/update + listing update + actionLog insert removes an extra round-trip from every bid that outbids someone. Per `CLAUDE.md`: "ALWAYS prefer `Promise.all()` for parallel queries... latency matters more than bandwidth."

MathiasGruber (commented):
Performance review — one spot where DB calls can be parallelized further.

MathiasGruber (commented):
Code-quality review (logic, readability, DRY).

app/src/server/api/routers/auction.ts
```diff
@@ -380,6 +515,22 @@ export const auctionRouter = createTRPCRouter({
...
+          user.level > auction.bidderMaxLevel
+        ) {
+          return errorResponse(
+            `This auction is only open to characters between level ${auction.bidderMinLevel} and ${auction.bidderMaxLevel}.`,
+          );
+        }
+      }
+      const buyoutPrice = auction.buyoutPrice;
+      const isListedBuyoutAmount =
+        buyoutPrice != null && Math.abs(amount - buyoutPrice) < 1e-6;
```
MathiasGruber: `isListedBuyoutAmount` uses `Math.abs(amount - buyoutPrice) < 1e-6` to treat a bid as a buyout, but prices are now integer-only end-to-end (schema, DB column, `createAuctionListingSchema`). The float tolerance is unnecessary, and it lets `amount = buyoutPrice + 1e-7` bypass the `Number.isInteger` check below — the stored `currentPrice` (int) then disagrees with the `auctionBid.amount` that was recorded. Either tighten `placeBid`'s input schema to `z.number().int().min(1)` and drop this carve-out, or at minimum compare with `amount === buyoutPrice` so the integer invariant holds.

app/src/server/api/routers/auction.ts
```diff
@@ -114,18 +164,85 @@ export const auctionRouter = createTRPCRouter({
...
+        }
+        if (applyUserListingPrivacy) {
+          q = q.innerJoin(listingCallerUser, eq(listingCallerUser.userId, ctx.userId));
+        }
+        listings = await q
+          .where(whereClause)
+          .orderBy(desc(auctionListing.expiresAt))
+          .limit(limit)
+          .offset(skip);
+      }
```
MathiasGruber: The two branches that build the listings query duplicate ~10 lines just to add the `auctionBid` inner-join when `onlyBidOn` is true — same select, joins, where, orderBy, limit, offset. Build a single query and add the bid join conditionally beforehand (similar to how the seller/caller joins are already conditional). As written, any future tweak to the listing query has to be made in both places.

app/src/app/auctionhouse/page.tsx
```diff
@@ -170,159 +254,265 @@ const AuctionListing: React.FC<AuctionListingProps> = ({ selectedStatus }) => {
...
+                "flex flex-col items-center justify-center gap-0.5 rounded-lg px-1 py-2.5 font-semibold text-[11px] transition-all sm:flex-row sm:gap-2 sm:py-3 sm:text-sm",
+                browseTab === "MY_BIDS"
+                  ? "bg-background text-foreground shadow-sm ring-1 ring-border/60"
+                  : "text-muted-foreground hover:bg-background/60 hover:text-foreground",
+              )}
+              onClick={() => setBrowseTab("MY_BIDS")}
+            >
+              <Hammer className="h-4 w-4 shrink-0 opacity-90" />
+              <span className="text-center leading-tight">My bids</span>
+            </button>
```
MathiasGruber: The four tab buttons are 15-line near-duplicates differing only by icon, id, label, and tab key. Extract a small `BrowseTabButton` (or map over a `[{ key, icon, label, id }]` config) so the shared aria/role/className logic lives once. As is, any styling tweak has to be repeated four times and it's easy for one to drift.

MathiasGruber (commented):
Logic review test.

app/src/server/api/routers/auction.ts
```diff
@@ -28,47 +49,63 @@ export const auctionRouter = createTRPCRouter({
...
+          columns: { role: true },
+        }),
+      ]);
+
+      if (
+        listing &&
+        caller?.role === "USER" &&
+        listing.listingType === "DIRECT" &&
+        listing.sellerId !== ctx.userId &&
+        listing.targetUserId !== ctx.userId
```
MathiasGruber, line 105: Returning `null` here for denied USER callers will break the dialog in `app/src/app/auctionhouse/page.tsx` (`AuctionDetailsDialog`). That component's existing guard is `if (isLoading || !listing) return <Loader/>` (around line 783) — so a denied caller resolves with `isLoading=false, listing=null` and is shown a permanent spinner with no way to recover except closing the dialog. Either throw a tRPC error (e.g. `FORBIDDEN`) so the client error handler kicks in, or update the dialog to distinguish denied-vs-loading and render an explicit "not available" message.

MathiasGruber (commented):

MathiasGruber (approved):
Looks good. I've run the SQL migration on prod database

PR #335, opened 2025-01-27
Fix issue #334: Medical Ninja

TheNinja-RPG:
This pull request fixes #334.

The PR successfully implements all required features from the issue description:

1. Added new Occupation menu button with Medical Ninja signup option
2. Implemented occupation system that only allows one occupation at a time
3. Added new Legendary Medical Ninja rank requiring 700,000 EXP
4. Implemented 30% item discount for all Medical Ninjas
5. Added special ability for Legendary Medical Ninjas to heal Chakra and Stamina
6. Created Medical Ninja Squad system with:
   - Squad management UI in hospital
   - 5 squad limit per village
   - 10 member limit per squad
   - Proper permissions (Kage can create/delete, leaders can manage members)
   - Squad features blocked for Syndicate
   - Full CRUD operations for squads

The changes include all necessary database schema updates, new API endpoints, frontend components, and proper integration with existing systems like the hospital and item shop. The code shows proper permission handling, limit enforcement, and integration with existing game mechanics. All core functionality appears to be properly implemented and working as specified in the issue.

Automatic fix generated by [OpenHands](https://github.com/All-Hands-AI/OpenHands/) 🙌

PR #1328, opened 2026-04-14
Feature/724 combat new player immunity

### LINKED ISSUE #724: Combat Changes for New Players
**### I. Immunity Threshold for Players:**

To foster a more welcoming environment for new participants, an immunity threshold is proposed:
Standard Immunity for Newer Players: Players with less than 500,000 Experience Points (XP) are protected from attacks by more experienced players. This "Standard Immunity" is in place to allow new users to acclimate to the game, learn its mechanics, and gain XP without constant harassment.

Immunity Exception: This immunity is not absolute. If a player below the 500,000 XP threshold attacks an ally of a higher-XP player within a 5-minute window, their immunity will be temporarily lifted. This immunity exception prevents newer players from exploiting the system by repeatedly attacking lower XP users 

The proposed combat experience brackets are as follows:

- Bracket 1: 0 - 500,000 XP
- Bracket 2: 500,001 - 1,000,000 XP
- Bracket 3: 1,000,001 - 1,500,000 XP
- Bracket 4: 1,500,001 - 2,000,000 XP
- Bracket 5: 2,000,001 - 2,500,000 XP
- Bracket 6: 2,500,001 - 3,000,000 XP
- Bracket 7: 3,000,000 XP+  users should have all defense, generals, and one offense capped

Combat Engagement Rules:

Bracket 1 XP: Can initiate attacks against any player.
Bracket 2 XP: Can only attack Bracket 1 players after a Bracket 1 player has attacked someone within a Bracket 2 village or alliance. The protection is lifted for two hours. 

### **For Villages and Factions that are at war War:**

To prevent low-level players from being targeted and harassed while their village is at war. A sort of opt in system will be used to mark a player as war participant so higher level players can target them if they are active in the war and protect low level players that aren't doing any war activity. 

1. If a player has accepted a war mission and is currently outside of their village. 
2. If a player has killed a player from a village/faction they are at war with
3. If a player has also killed enemy Shrine AI

ifroshty:
# Pull Request

  ## What was implemented

  Closes #724

  Replaces the existing ±15 level PvP restriction with an XP bracket system that protects new and mid-level players from being targeted by
  significantly more experienced players.

  ### XP Brackets

  | Bracket | XP Range |
  |---------|----------|
  | 1 | 0 – 500,000 |
  | 2 | 500,001 – 1,000,000 |
  | 3 | 1,000,001 – 1,500,000 |
  | 4 | 1,500,001 – 2,000,000 |
  | 5 | 2,000,001 – 2,500,000 |
  | 6 | 2,500,001 – 3,000,000 |
  | 7 | 3,000,001+ |

  By default players can only initiate COMBAT against players in their own bracket.

  ### Guard Rules

  1. **Same village** — always blocked, regardless of sector or war status
  2. **Allied village** — always blocked
  3. **Sector 335 (War-Torn Battleground)** — bypasses all bracket checks; any non-village, non-allied combat is allowed
  4. **Same bracket** — allowed
  5. **Bracket immunity lift** — if the target recently initiated any attack, their bracket immunity is lifted for **5 minutes**, allowing anyone to cross-bracket attack them
  6. **War participants** — if both the attacker and all targets hold active `warParticipantUntil` status AND their villages are on opposing sides
  of an active war, cross-bracket combat is allowed for **2 hours**
  7. Otherwise — blocked with a message showing both bracket numbers

  ### War Participant Triggers

  A player gains `warParticipantUntil` (2 hours) when they:

  - Win a battle against a war enemy (player or shrine AI)
  - Accept a war mission
  - Successfully initiate a COMBAT battle against a war enemy

  ### Schema Changes

  Two new `datetime` columns added to `UserData`:

  | Column | Purpose |
  |--------|---------|
  | `bracketImmunityLiftedUntil` | Tracks the 5-minute window after a player initiates any attack — others may cross-bracket target them during this
   window |
  | `warParticipantUntil` | Tracks 2-hour war participation status — enables cross-bracket combat between active war opponents |

  Both default to `CURRENT_TIMESTAMP(3)` (effectively "no immunity / not a participant") so existing players are unaffected by the migration.

  ## Why

  New players were being farmed by significantly higher-XP players despite the ±15 level gap check, because the level-to-XP curve compresses at higher levels — two players 15 levels apart at high level can have vastly different actual power. The bracket system uses raw XP thresholds which are linear and intuitive, making protection predictable for players.

  The immunity-lift and war-participant mechanics preserve meaningful PvP for active combatants without letting low-XP players abuse protection by attacking freely without consequence.

  ## Breaking Changes

  - The ±15 level PvP restriction is **removed and replaced**. Players who could previously attack across level gaps (when at war) now must additionally hold active `warParticipantUntil` status to attack cross-bracket during war.
  - Same-village attacks are now **explicitly blocked** in COMBAT outside of the War-Torn Battleground (previously only blocked in protected zones).
  - Allied-village attacks are now **explicitly blocked** in COMBAT outside of the War-Torn Battleground (previously no explicit check existed for non-protected zones).

  ## Files Changed

  | File | Change |
  |------|--------|
  | `app/drizzle/constants.ts` | Added `XP_BRACKETS`, `BRACKET_IMMUNITY_LIFT_SECS`, `WAR_PARTICIPANT_SECS` |
  | `app/drizzle/schema.ts` | Added `bracketImmunityLiftedUntil` and `warParticipantUntil` to `userData` |
  | `app/drizzle/migrations/0007_nosy_kinsey_walden.sql` | Migration for the two new columns |
  | `app/src/libs/profile.ts` | Added `getExpBracket(experience)` utility |
  | `app/src/server/api/routers/combat.ts` | Replaced level restriction with bracket guard system; stamps `bracketImmunityLiftedUntil` on attackers;
   stamps `warParticipantUntil` only after battle creation succeeds |
  | `app/src/libs/combat/database.ts` | Stamps `warParticipantUntil` on war kill (deduplicated per battle) |
  | `app/src/server/api/routers/quests.ts` | Stamps `warParticipantUntil` on war quest acceptance |
  | `app/tests/libs/profile.test.ts` | Added 14 boundary tests for `getExpBracket` |

  ## License

## Summary by CodeRabbit

* **New Features**
  * PvP bracket system based on player experience
  * Two cooldowns: Bracket Immunity Lifted and War Participant status
  * Combat targeting now enforces bracket-based restrictions
  * UI updates: bracket shown on profiles, cooldown rows, and a Bracket Eligibility badge that auto-updates timers

### DIFF
--- a/app/src/layout/MenuBoxProfile.tsx
+++ b/app/src/layout/MenuBoxProfile.tsx
@@ -61,7 +61,7 @@ import { sealCheck } from "@/libs/combat/tags";
 import type { GroundEffect, UserEffect } from "@/libs/combat/types";
 import { getPreventTypeName, isEffectActive } from "@/libs/combat/util";
 import { useGameMenu } from "@/libs/menus";
-import { calcLevelRequirements } from "@/libs/profile";
+import { calcLevelRequirements, getExpBracket } from "@/libs/profile";
 import { cn } from "@/libs/shadui";
 import { calcCovertTrainingFinishAt } from "@/libs/stealth";
 import { trainingSpeedSeconds } from "@/libs/train";
@@ -374,6 +374,23 @@ const MenuBoxProfile: React.FC = () => {
                 {userData.village.name}
               </Link>
             </p>
+          )}
+          {userData && (
+            <TooltipProvider delayDuration={50}>
+              <Tooltip>
+                <TooltipTrigger asChild>
+                  <p className="cursor-default">
+                    <b>Bracket: </b>
+                    {getExpBracket(userData.experience)}/7
+                  </p>
+                </TooltipTrigger>
+                <TooltipContent>
+                  PvP bracket based on experience. You can freely attack players in your
+                  bracket or higher. Players below your bracket are protected unless
+                  their immunity is lifted or both sides are active war participants.
+                </TooltipContent>
+              </Tooltip>
+            </TooltipProvider>
           )}
           <p suppressHydrationWarning>
             <b>Time: </b> {gameTime}
--- a/app/src/layout/MenuBoxProfile.tsx
+++ b/app/src/layout/MenuBoxProfile.tsx
@@ -201,17 +201,33 @@ const MenuBoxProfile: React.FC = () => {
 
   const immunitySecsLeft = immunityData.secsLeft;
 
-  // Recompute on every render (no memo) so the > 0 guard goes false as soon as the
-  // timestamp expires, unmounting the icon/tooltip row without waiting for fresh userData.
-  const timerNow = Date.now();
-  const bracketImmunitySecsLeft = Math.max(
-    0,
-    ((userData?.bracketImmunityLiftedUntil?.getTime() ?? 0) - timerNow) / 1000,
-  );
-  const warParticipantSecsLeft = Math.max(
-    0,
-    ((userData?.warParticipantUntil?.getTime() ?? 0) - timerNow) / 1000,
-  );
+  // Memoized per-timestamp anchors — stable createdAt prevents Cooldown from tearing down
+  // and re-creating its setInterval on every parent render. secsLeft is recomputed only when
+  // the server-side timestamp changes. The > 0 visibility guard unmounts the row on expiry.
+  const bracketImmunityData = useMemo(() => {
+    const now = Date.now();
+    return {
+      createdAt: now,
+      secsLeft: Math.max(
+        0,
+        ((userData?.bracketImmunityLiftedUntil?.getTime() ?? 0) - now) / 1000,
+      ),
+    };
+  }, [userData?.bracketImmunityLiftedUntil]);
+
+  const warParticipantData = useMemo(() => {
+    const now = Date.now();
+    return {
+      createdAt: now,
+      secsLeft: Math.max(
+        0,
+        ((userData?.warParticipantUntil?.getTime() ?? 0) - now) / 1000,
+      ),
+    };
+  }, [userData?.warParticipantUntil]);
+
+  const bracketImmunitySecsLeft = bracketImmunityData.secsLeft;
+  const warParticipantSecsLeft = warParticipantData.secsLeft;
 
   // Battle user state
   const battleUser = battle?.usersState.find((u) => u.userId === userData?.userId);
@@ -418,7 +434,7 @@ const MenuBoxProfile: React.FC = () => {
                   <div className="flex flex-row items-center text-orange-500">
                     <ShieldAlert className="mr-2 h-6 w-6" />
                     <Cooldown
-                      createdAt={timerNow}
+                      createdAt={bracketImmunityData.createdAt}
                       totalSeconds={bracketImmunitySecsLeft}
                       initialSecondsLeft={bracketImmunitySecsLeft}
                       setState={setState}
@@ -437,7 +453,7 @@ const MenuBoxProfile: React.FC = () => {
                   <div className="flex flex-row items-center text-red-500">
                     <Swords className="mr-2 h-6 w-6" />
                     <Cooldown
-                      createdAt={timerNow}
+                      createdAt={warParticipantData.createdAt}
                       totalSeconds={warParticipantSecsLeft}
                       initialSecondsLeft={warParticipantSecsLeft}
                       setState={setState}
--- a/app/src/layout/MenuBoxProfile.tsx
+++ b/app/src/layout/MenuBoxProfile.tsx
@@ -47,6 +47,7 @@ import {
   MEDICAL_MISSIONS_PER_DAY,
   MISSIONS_PER_DAY,
   PVP_MISSIONS_PER_DAY,
+  XP_BRACKETS,
 } from "@/drizzle/constants";
 import { useLocalStorage } from "@/hooks/localstorage";
 import { useEffectivePools } from "@/hooks/useEffectivePools";
@@ -381,7 +382,7 @@ const MenuBoxProfile: React.FC = () => {
                 <TooltipTrigger asChild>
                   <p className="cursor-default">
                     <b>Bracket: </b>
-                    {getExpBracket(userData.experience)}/7
+                    {getExpBracket(userData.experience)}/{XP_BRACKETS.length}
                   </p>
                 </TooltipTrigger>
                 <TooltipContent>
--- a/app/src/layout/MenuBoxProfile.tsx
+++ b/app/src/layout/MenuBoxProfile.tsx
@@ -212,6 +212,17 @@ const MenuBoxProfile: React.FC = () => {
 
   const bracketImmunitySecsLeft = bracketImmunityData.secsLeft;
 
+  const warParticipantData = useMemo(() => {
+    const now = Date.now();
+    const secsLeft =
+      (userData?.warParticipantUntil &&
+        (userData.warParticipantUntil.getTime() - now) / 1000) ||
+      0;
+    return { createdAt: now, secsLeft };
+  }, [userData?.warParticipantUntil]);
+
+  const warParticipantSecsLeft = warParticipantData.secsLeft;
+
   // Battle user state
   const battleUser = battle?.usersState.find((u) => u.userId === userData?.userId);
 
@@ -426,6 +437,25 @@ const MenuBoxProfile: React.FC = () => {
                   </div>
                 </TooltipTrigger>
                 <TooltipContent>Bracket Immunity Lifted</TooltipContent>
+              </Tooltip>
+            </TooltipProvider>
+          )}
+          {userData && !userData.battleId && warParticipantSecsLeft > 0 && (
+            <TooltipProvider delayDuration={50}>
+              <Tooltip>
+                <TooltipTrigger className="w-full">
+                  <div className="flex flex-row items-center text-red-500">
+                    <Swords className="mr-2 h-6 w-6" />
+                    <Cooldown
+                      createdAt={warParticipantData.createdAt}
+                      totalSeconds={warParticipantSecsLeft}
+                      initialSecondsLeft={warParticipantSecsLeft}
+                      setState={setState}
+                      hideWhenDone
+                    />
+                  </div>
+                </TooltipTrigger>
+                <TooltipContent>War Participant</TooltipContent>
               </Tooltip>
             </TooltipProvider>
           )}
--- a/app/src/layout/MenuBoxProfile.tsx
+++ b/app/src/layout/MenuBoxProfile.tsx
@@ -202,33 +202,39 @@ const MenuBoxProfile: React.FC = () => {
 
   const immunitySecsLeft = immunityData.secsLeft;
 
-  // Memoized per-timestamp anchors — stable createdAt prevents Cooldown from tearing down
-  // and re-creating its setInterval on every parent render. secsLeft is recomputed only when
-  // the server-side timestamp changes. The > 0 visibility guard unmounts the row on expiry.
-  const bracketImmunityData = useMemo(() => {
+  // Memoize the (createdAt, totalSeconds) pair together — Cooldown computes its deadline as
+  // createdAt + totalSeconds*1000, so the two must be sampled at the same instant. The fresh
+  // secsLeft below is used only for the > 0 JSX visibility guard, so the parent row collapses
+  // immediately when Cooldown's onTick setState fires (instead of waiting for fresh userData).
+  const bracketImmunityAnchor = useMemo(() => {
     const now = Date.now();
     return {
       createdAt: now,
-      secsLeft: Math.max(
+      totalSeconds: Math.max(
         0,
         ((userData?.bracketImmunityLiftedUntil?.getTime() ?? 0) - now) / 1000,
       ),
     };
   }, [userData?.bracketImmunityLiftedUntil]);
+  const bracketImmunitySecsLeft = Math.max(
+    0,
+    ((userData?.bracketImmunityLiftedUntil?.getTime() ?? 0) - Date.now()) / 1000,
+  );
 
-  const warParticipantData = useMemo(() => {
+  const warParticipantAnchor = useMemo(() => {
     const now = Date.now();
     return {
       createdAt: now,
-      secsLeft: Math.max(
+      totalSeconds: Math.max(
         0,
         ((userData?.warParticipantUntil?.getTime() ?? 0) - now) / 1000,
       ),
     };
   }, [userData?.warParticipantUntil]);
-
-  const bracketImmunitySecsLeft = bracketImmunityData.secsLeft;
-  const warParticipantSecsLeft = warParticipantData.secsLeft;
+  const warParticipantSecsLeft = Math.max(
+    0,
+    ((userData?.warParticipantUntil?.getTime() ?? 0) - Date.now()) / 1000,
+  );
 
   // Battle user state
   const battleUser = battle?.usersState.find((u) => u.userId === userData?.userId);
@@ -452,9 +458,9 @@ const MenuBoxProfile: React.FC = () => {
                   <div className="flex flex-row items-center text-orange-500">
                     <ShieldAlert className="mr-2 h-6 w-6" />
                     <Cooldown
-                      createdAt={bracketImmunityData.createdAt}
-                      totalSeconds={bracketImmunitySecsLeft}
-                      initialSecondsLeft={bracketImmunitySecsLeft}
+                      createdAt={bracketImmunityAnchor.createdAt}
+                      totalSeconds={bracketImmunityAnchor.totalSeconds}
+                      initialSecondsLeft={bracketImmunityAnchor.totalSeconds}
                       setState={setState}
                       hideWhenDone
                     />
@@ -471,9 +477,9 @@ const MenuBoxProfile: React.FC = () => {
                   <div className="flex flex-row items-center text-red-500">
                     <Swords className="mr-2 h-6 w-6" />
                     <Cooldown
-                      createdAt={warParticipantData.createdAt}
-                      totalSeconds={warParticipantSecsLeft}
-                      initialSecondsLeft={warParticipantSecsLeft}
+                      createdAt={warParticipantAnchor.createdAt}
+                      totalSeconds={warParticipantAnchor.totalSeconds}
+                      initialSecondsLeft={warParticipantAnchor.totalSeconds}
                       setState={setState}
                       hideWhenDone
                     />
--- a/app/src/layout/MenuBoxProfile.tsx
+++ b/app/src/layout/MenuBoxProfile.tsx
@@ -410,7 +410,7 @@ const MenuBoxProfile: React.FC = () => {
               </Tooltip>
             </TooltipProvider>
           )}
-          {userData && bracketImmunitySecsLeft > 0 && (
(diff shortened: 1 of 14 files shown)


app/src/server/api/routers/combat.ts
```diff
@@ -1747,47 +1749,78 @@ export const initiateBattle = async (
...
+              ((r.villageIdA === user.villageId && r.villageIdB === t.villageId) ||
+                (r.villageIdB === user.villageId && r.villageIdA === t.villageId)),
+          ),
+        );
+        if (alliedTarget) {
+          return {
+            success: false,
+            message: `Cannot attack ${alliedTarget.username} — their village is allied with yours`,
+          };
+        }
```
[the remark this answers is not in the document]
ifroshty: Per the issue, "War-Torn Sector - Sector #335 - bypasses ALL bracket restrictions"

ifroshty:
@greptileai

MathiasGruber (commented):
Verified: `opponents.every(...)` was correctly changed to `opponents.some(...)` in commit `24d84f30`. The resolved thread for the Greptile P1 at `database.ts` line 1239–1242 reflects the actual state of the code — the fix is applied and the thread resolution stands.

---

MathiasGruber (commented):
**`BracketEligibilityBadge` will show stale eligibility after immunity expires without a page refresh**

`app/src/layout/PublicUser.tsx` — `BracketEligibilityBadge` component

The badge computes `const now = Date.now()` inline on each render, which is correct — but the component has no timer to force a re-render when `targetBracketImmunityLiftedUntil` elapses. The `getPublicUser` query at line 202 has no `refetchInterval`, so once the profile data is cached the badge can stay in the "✓ Attackable (their immunity is lifted)" state indefinitely after the immunity window closes, until the user navigates away or something else triggers a refetch.

Minimal fix: add a `useEffect`-driven `setTimeout` inside `BracketEligibilityBadge` (or its parent) that forces a re-render when the expiry is reached:

```tsx
const BracketEligibilityBadge: React.FC<BracketEligibilityBadgeProps> = ({ ... }) => {
  const [, forceUpdate] = useState(0);
  const now = Date.now();
  const msUntilImmunityExpires = targetBracketImmunityLiftedUntil.getTime() - now;

  useEffect(() => {
    if (msUntilImmunityExpires <= 0) return;
    const id = setTimeout(() => forceUpdate((n) => n + 1), msUntilImmunityExpires);
    return () => clearTimeout(id);
  }, [msUntilImmunityExpires]);

  // rest of render unchanged...
};
```

Alternatively, set `refetchInterval` on the `getPublicUser` call to a value no longer than `BRACKET_IMMUNITY_LIFT_SECS * 1000` so the server timestamp stays fresh.

---

MathiasGruber:

- PR's original migration `0007_nosy_kinsey_walden.sql`: deleted; migration regenerated after rebase as `0008_old_shooting_star.sql` adding `UserData.bracketImmunityLiftedUntil` and `UserData.warParticipantUntil`.
- All other files (combat libs, quest router, combat router, etc.) auto-merged cleanly.

app/src/layout/MenuBoxProfile.tsx
```diff
@@ -200,6 +202,34 @@ const MenuBoxProfile: React.FC = () => {
...
+      createdAt: now,
+      secsLeft: Math.max(
+        0,
+        ((userData?.warParticipantUntil?.getTime() ?? 0) - now) / 1000,
+      ),
+    };
+  }, [userData?.warParticipantUntil]);
+
+  const bracketImmunitySecsLeft = bracketImmunityData.secsLeft;
+  const warParticipantSecsLeft = warParticipantData.secsLeft;
```
MathiasGruber: Re-opening previous feedback — the cooldown row visibility issue is not fully addressed.

`bracketImmunitySecsLeft` and `warParticipantSecsLeft` are still memoized from `Date.now()` with deps on the server timestamp only. When the `Cooldown` child finishes and calls `setState((prev) => prev + 1)`, the parent re-renders but `useMemo` returns the cached positive value (deps haven't changed). That means `bracketImmunitySecsLeft > 0` at lines 448 and 467 stays truthy, leaving the icon + tooltip row visible until fresh `userData` arrives.

`hideWhenDone` only hides the inner counter text, not the parent `<TooltipProvider>` wrapper. The original suggestion still applies: either recompute `secsLeft` on every render (no `useMemo`), or clear the parent row from local state via `setState` when the countdown completes.

app/src/server/api/routers/combat.ts
```diff
@@ -2176,6 +2231,17 @@ export const initiateBattle = async (
...
+  // Stamp war participants only after battle creation succeeds — avoids leaking
+  // the timer on failed initiations (e.g. target moved before attack landed).
+  if (warAggressorIds.length > 0) {
+    await client
+      .update(userData)
+      .set({
+        warParticipantUntil: sql`GREATEST(${userData.warParticipantUntil}, NOW() + INTERVAL ${WAR_PARTICIPANT_SECS} SECOND)`,
+      })
+      .where(inArray(userData.userId, warAggressorIds));
+  }
```
MathiasGruber: Performance: This adds an extra sequential DB roundtrip in `initiateBattle` after the main `Promise.all` (line 2099). Per CLAUDE.md, prefer one big query over multiple small ones — latency compounds.

Suggested fix: merge `warParticipantUntil` into the existing `userData` update inside the main `Promise.all` using a conditional SQL expression, e.g.:

```ts
warParticipantUntil: warAggressorIds.length > 0
  ? sql`CASE WHEN ${inArray(userData.userId, warAggressorIds)} THEN GREATEST(${userData.warParticipantUntil}, NOW() + INTERVAL ${WAR_PARTICIPANT_SECS} SECOND) ELSE ${userData.warParticipantUntil} END`
  : sql`${userData.warParticipantUntil}`,
```

Since the stamp uses `GREATEST(existing, new)`, a battle that fails the rowsAffected check (rollback path) leaves a benign stamp at most ~2h that gets cleared by `handleWarEnd`. The current comment justifies the separation by ‘avoiding leak on failed initiations,’ but the leak is already bounded and the extra roundtrip hits every successful PvP initiation in a war zone.

app/src/libs/war.ts
```diff
@@ -282,6 +282,14 @@ export const handleWarEnd = async (activeWar: FetchActiveWarsReturnType) => {
       .update(war)
       .set({ status, endedAt })
       .where(and(eq(war.id, activeWar.id), isNull(war.endedAt))),
+    // Clear war participant status for all players in involved villages so the timer
+    // disappears immediately and cross-bracket exemption cannot be exploited after war ends.
+    // Use epoch (new Date(0)) rather than now() to be unambiguously in the past regardless
+    // of any JS-to-DB clock skew or concurrent in-flight requests.
+    drizzleDB
+      .update(userData)
+      .set({ warParticipantUntil: new Date(0) })
```
MathiasGruber, line 307: Logic bug: this wipes `warParticipantUntil` to epoch for every member of `involvedVillageIds` whenever any single war they're in ends. If a player is participating in multiple concurrent wars (e.g. their village is at war with two other villages, or they're a war ally in several wars), ending one war strips their participant status for the others as well — they immediately lose cross-bracket exemption while the other war is still active.

Suggested fix: scope this update to users whose only remaining war exposure is the one ending, or skip the clear entirely and let the 2-hour timer expire naturally (the SQL guard `GREATEST(...)` already prevents stale stamps from extending past the timer).

app/src/server/api/routers/quests.ts
```diff
@@ -753,11 +754,19 @@ export const questsRouter = createTRPCRouter({
...
+      // the CAS success and the upsert burns one daily slot without delivering the quest
+      // (acceptable), whereas the reverse order would deliver a quest when the limit is reached
+      // (unacceptable — creates orphaned quest entries the user cannot access).
       if (questData.questType === "war") {
         const result = await ctx.drizzle
           .update(userData)
-          .set({ dailyWarMissions: sql`${userData.dailyWarMissions} + 1` })
+          .set({
+            dailyWarMissions: sql`${userData.dailyWarMissions} + 1`,
+            warParticipantUntil: sql`GREATEST(${userData.warParticipantUntil}, NOW() + INTERVAL ${WAR_PARTICIPANT_SECS} SECOND)`,
```
MathiasGruber: DRY: the SQL fragment `GREATEST(${userData.warParticipantUntil}, NOW() + INTERVAL ${WAR_PARTICIPANT_SECS} SECOND)` is duplicated verbatim in three places — here, `app/src/libs/combat/database.ts:397` (war kill stamp), and `app/src/server/api/routers/combat.ts:2240` (war aggressor stamp on initiate). Extract a small helper like `extendWarParticipantSql()` (and a sibling for `bracketImmunityLiftedUntil`) to keep the three call sites in sync if the formula or constant changes.

app/src/server/api/routers/combat.ts
```diff
@@ -2040,6 +2076,49 @@ export const initiateBattle = async (
...
+          return targetIds.some((tid) => {
+            const target = users.find((u) => u.userId === tid);
+            return (
+              !!target &&
+              !target.isAi &&
+              getExpBracket(target.experience) >= attackerBracket
+            );
+          });
+        })
+      : [];
```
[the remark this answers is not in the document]
PlayNinjara: Keeping the same-or-higher filter here — it's intentional, and victim retaliation isn't actually blocked by it.

`bracketImmunityLiftedUntil` lets *others* cross-bracket attack the stamped player, and protection is one-directional: high brackets can't hit low without an exemption, but low can always hit high. Walking the cases:

- **Victim is lower-bracket than the attacker** → they retaliate by punching *up*, which is always allowed regardless of any lift.
- **Victim is same/higher-bracket** → the `>= attackerBracket` filter already matched, so the attacker *is* stamped, and the victim punches *down* via that lift.

So the victim can always counterattack. In the B5→B1 example, the B1 victim can already hit the B5 freely — the B7 in the scenario isn't retaliating, it's a third party piling on. Lifting immunity only when attacking same-or-higher is the deliberate rule (refined in f5c5b717): you forfeit protection when you punch up or sideways, but punching down at an already-exempted target doesn't open you to your own higher-bracket peers.

(Minor: the suggested `userIds`-only diff also drops the `!isAi` guard, so it would lift immunity for attacking an AI on the map.)

The stale part is the PR-description wording ("any attack / anyone"), which we'll tighten to match the implemented rule.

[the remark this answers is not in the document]
PlayNinjara: Correction — I was wrong, the filter should go. Checked the spec in #724: theeneon clarified the trigger is "initiate an attack on **anyone**" (Q1) and "**anyone from any bracket** can attack those players" (Q3), and the finalized plan stamps the lift unconditionally on successful attack initiation. The `>= attackerBracket` filter contradicts that.

Applied: dropped the bracket comparison so every COMBAT initiator's `bracketImmunityLiftedUntil` is stamped when they attack a real player. Kept a `!isAi` guard so attacking an NPC/AI doesn't lift PvP immunity.

ifroshty:
@greptileai

MathiasGruber:

**Conflicts resolved:**
- **Migrations**: The PR's migration conflicted with new migrations added to `main` (`0008`–`0010`). Per the migration-conflict policy, the PR's own migration file was dropped, the migration metadata (`_journal.json` / snapshots) was reset to match `main`, and a fresh migration (`0011_curvy_lady_bullseye.sql`) was regenerated via `make makemigrations`. It cleanly adds the `UserData.bracketImmunityLiftedUntil` and `UserData.warParticipantUntil` columns.
- **schema.ts**: Auto-merged; the PR's `userData` schema additions are preserved on top of the latest `main`.

MathiasGruber (commented):
Performance review — minimizing DB round-trips in the new war-participant write paths.

app/src/server/api/routers/quests.ts
```diff
@@ -770,6 +780,10 @@ export const questsRouter = createTRPCRouter({
           );
         }
         await upsertQuestEntry(ctx.drizzle, user, questData);
+        await ctx.drizzle
+          .update(userData)
+          .set({ warParticipantUntil: extendWarParticipantSql() })
+          .where(eq(userData.userId, user.userId));
```
MathiasGruber: Sequential third roundtrip on the same `userData` row already updated by the CAS at line 766. Per CLAUDE.md "Minimize DB Round-Trips", this can be merged. Two viable options:

1. Merge `warParticipantUntil: extendWarParticipantSql()` into the CAS update at line 766 alongside `dailyWarMissions`. Removes the roundtrip; the only tradeoff is a bounded ~2h leak if `upsertQuestEntry` then fails — the same fail-safe leak the equivalent merge in `combat.ts` (lines 2210-2223) explicitly accepts.
2. If the stamp must follow `upsertQuestEntry` for correctness, at minimum run them in parallel via `Promise.all`: order between the upsert and the stamp does not matter once both follow the CAS guard.

As-is, every accepted war mission pays ~3 sequential RTT to the DB instead of 1-2.

app/src/libs/combat/database.ts
```diff
@@ -601,6 +605,19 @@ export const updateWars = async (
...
+        // Deduplicate: one DB write per killer per battle.
+        if (!stampedWarParticipantIds.has(user.userId)) {
+          stampedWarParticipantIds.add(user.userId);
+          otherPromises.push(
+            client
+              .update(userData)
+              .set({
+                warParticipantUntil: extendWarParticipantSql(),
+              })
+              .where(eq(userData.userId, user.userId)),
```
MathiasGruber: The killer's `warParticipantUntil` stamp here is a separate `update(userData)` on the same row that `updateUser()` already updates in its main mutation (see `updateUser` around line 1438). Both run in the parent `Promise.all` inside `performAction` (combat.ts line 735), so every kill in a war battle pays an extra DB roundtrip to write a single column on the same row.

Per CLAUDE.md ("Battle Performance Requirements"), the parallel mutation step should batch per-row updates. Recommend moving the stamp into `updateUser`'s `update(userData).set({...})` block — gate it on `result?.didWin` plus whether the user has any matching war (computable from `getWarsArray` on the pre-loaded battle state, no extra fetch needed) and drop the dedicated UPDATE here. `extendWarParticipantSql()` keeps it idempotent.

app/src/server/api/routers/combat.ts
```diff
@@ -1769,47 +1774,78 @@ export const initiateBattle = async (
...
-            }
+        // Guard 2: Cannot attack members of allied villages
+        const alliedTarget = nonAiTargets.find((t) =>
+          relations.some(
+            (r) =>
+              r.status === "ALLY" &&
+              ((r.villageIdA === user.villageId && r.villageIdB === t.villageId) ||
+                (r.villageIdB === user.villageId && r.villageIdA === t.villageId)),
+          ),
+        );
```
MathiasGruber, line 1802: DRY: this inline alliance check re-implements `findRelationship` (already imported and used elsewhere in this file, e.g. L2893). Consider simplifying to:

```ts
const alliedTarget = nonAiTargets.find(
  (t) => findRelationship(relations, user.villageId, t.villageId)?.status === "ALLY",
);
```

Same semantics, less surface area for the symmetric-pair logic to drift away from the helper.

MathiasGruber (commented):
Logic/readability/DRY review of the bracket immunity + war participant changes.

app/src/layout/MenuBoxProfile.tsx
```diff
@@ -200,6 +202,40 @@ const MenuBoxProfile: React.FC = () => {
...
+      totalSeconds: Math.max(
+        0,
+        ((userData?.warParticipantUntil?.getTime() ?? 0) - now) / 1000,
+      ),
+    };
+  }, [userData?.warParticipantUntil]);
+  const warParticipantSecsLeft = Math.max(
+    0,
+    ((userData?.warParticipantUntil?.getTime() ?? 0) - Date.now()) / 1000,
+  );
```
MathiasGruber: DRY: `bracketImmunityAnchor` / `bracketImmunitySecsLeft` and `warParticipantAnchor` / `warParticipantSecsLeft` are structurally identical — same memo shape, same `Math.max(0, (until - now)/1000)` formula. Worth extracting a small helper (e.g. `useExpiryAnchor(until)` returning `{ createdAt, totalSeconds, secsLeft }`) so a future third timer doesn't get copy-pasted again and the two existing blocks stay in lockstep.

app/src/server/api/routers/combat.ts
```diff
@@ -1769,47 +1774,78 @@ export const initiateBattle = async (
...
+          });
+
+          if (blockedTarget) {
+            const targetBracket = getExpBracket(blockedTarget.experience);
+            return {
+              success: false,
+              message: `Cannot attack ${blockedTarget.username} — different combat bracket (${attackerBracket} vs ${targetBracket})`,
+            };
           }
         }
```
MathiasGruber, line 1848: Readability/efficiency: this whole bracket guard sits inside the `for (const i of users.keys())` loop and runs once per *attacker* (only the `userIds.includes(user.userId)` branch). With multiple attackers it re-derives `nonAiTargets`, `crossBracketTargets`, and walks `findWarsWithUser` from scratch each time. Hoisting the per-target setup (nonAiTargets, isInWarTornSector, `now`, the war/atWar lookups) above the loop and only re-deriving `attackerBracket`/`crossBracketTargets` per attacker would be both faster and easier to follow.

app/src/libs/combat/database.ts
```diff
@@ -1183,6 +1183,29 @@ export const updateUser = async (
...
+        (t) =>
+          t.userId !== userId &&
+          !t.isSummon &&
+          findWarsWithUser(
+            getWarsArray(curBattle, t),
+            userWars,
+            t.villageId,
+            user.villageId,
+          ).length > 0,
+      );
```
MathiasGruber, line 1211: Minor logic: the `usersState.some(t => t.userId !== userId && !t.isSummon && findWarsWithUser(...))` walks *every* other participant — including the killer's own teammates. `findWarsWithUser` will only return true when the two villages are on opposing war sides, so same-village teammates are filtered out naturally, but a teammate from a war-ally village (or any allied side-of-the-war village) will be matched and stamp the killer as a war participant even though no war foe was actually defeated. Restricting `t` to the opposing side (e.g. via the existing left/right side tracking, or the `usersState` `controllerId` opposite of `user.controllerId`) makes the rule "won a fight that contained a war foe on the other side" instead of "...anywhere in the battle".

app/src/layout/PublicUser.tsx
```diff
@@ -2193,3 +2206,74 @@ const CombatHistoryTab: React.FC<TabComponentProps> = ({
...
+      Math.max(0, nextExpiry - Date.now()),
+    );
+    return () => clearTimeout(id);
+  }, [nextExpiry]);
+
+  if (attackerBracket <= targetBracket) {
+    return (
+      <p className="font-medium text-green-600 text-sm">
+        ✓ Bracket-eligible (same or higher bracket — server-side checks still apply)
+      </p>
```
MathiasGruber, line 2257: Readability: "same or higher bracket" is ambiguous when paired with the comparison `attackerBracket <= targetBracket` — it really means *the target* is in your bracket or higher. From the viewer's perspective reading the badge on someone else's profile, that's easy to misread as "you're in the same or higher bracket" (the opposite). A small reword like "✓ Bracket-eligible (they are in your bracket or higher — server-side checks still apply)" removes the ambiguity.

MathiasGruber (approved):
LGTM. Migrations have been run

PR #1334, opened 2026-04-16
Feature/townhall changes 1324

ifroshty:
# Pull Request

  ## Townhall Changes — Building Rebalance, Village Renames & Name Protections

  ### What was implemented

  **Building upgrade cost formula**
  Replaced the flat `baseCost` column with a computed formula: `200,000 + 150,000 × current level`. This makes early upgrades accessible while scaling costs meaningfully at higher levels. The `baseCost` column has been fully removed from the schema and replaced with two constants.

  **Per-building bonus rebalance**
  | Building | Change |
  |---|---|
  | Town Hall | `structureDiscountPerLvl` 1% → 2%/lvl |
  | Training Grounds | `trainBoostPerLvl` 1% → 2%/lvl |
  | Home | `sleepRegenPerLvl` 2% → 5%/lvl |
  | Hospital | Base heal time 30 min → 10 min; base cost 40 Ryo/100HP → 400 Ryo/100HP |
  | Mission Hall / Crimes Board | New `missionRewardPerLvl` = 2%/lvl (mission & errand reward boost) |

  **Village renames**
  | Old | New |
  |---|---|
  | Current | Akikaze |
  | Glacier | Hyorin |
  | Shroud | Akasumi |
  | Shine | Shirohana |

  Village chat channels and music theme assignments updated to match.

  **Legacy name protection**
  Old village names (`Shine`, `Glacier`, `Shroud`, `Current`) are now reserved — they cannot be registered as usernames, clan names, or ANBU squad names. New village names and existing protected names (`Freedom State`, `Horizon`, core villages) are equally protected.

  **Validator hardening**
  - `factionEditSchema` (clan rename) now enforces the same alphanumeric-only regex as `clanCreateSchema` — previously a clan could be renamed to
  include spaces or special characters via the edit endpoint.
  - `anbuEditSchema` introduced; the `editSquad` router endpoint was using an inline schema with no length, format, or name validation.
  - `anbuCreateSchema` and `anbuRenameSchema` extended with alphanumeric regex and banned-name refine.

  **Upgrade dialog UX**
  The structure upgrade confirmation dialog now lists what the next level provides (e.g. "+2% bank interest"), so players can make an informed decision before spending tokens.

    ### Migrations
    - `0009_violet_zemo` — ADD `missionRewardPerLvl tinyint DEFAULT 0` to `VillageStructure`, then DROP `baseCost`. The ADD-before-DROP order leaves a short transition window so the updated application code can roll out before the column DROP lands.
    - No DDL migration exists for the village/channel rename. Renames are reflected in the `app/data/villages.sql` seed for fresh installs; the production rename is intended to run as a separate one-off `UPDATE Village SET name = ... WHERE name = ...` (and matching `Conversation` channel updates) coordinated with the code
  deploy, not via `drizzle-kit migrate`.
    - Migration `0008_busy_dreaming_celestial` (already on `main`) is unrelated to this PR — it adds `AuctionListing.bidderMinLevel`/`bidderMaxLevel` and the associated index, and was picked up during the rebase.

  Replaces the existing three bullets that reference 0007 / 0008 / 0009 numbering — those reflected the pre-rebase migration plan. After the rebase, the actual situation is: a single combined 0009_violet_zemo migration on this branch, main's unrelated 0008_busy_dreaming_celestial underneath it, and the village rename
  intentionally living outside the drizzle migration system.

  ### Breaking changes
  - **Village names changed** — any code outside this repo that hardcodes `"Current"`, `"Glacier"`, `"Shroud"`, or `"Shine"` as village name
  strings will need updating.
  - **Hospital costs increased 10×** — intentional rebalance per issue spec (400 Ryo/100HP).
  - **Clan/ANBU names with spaces or special characters** can no longer be set via the edit endpoints (enforcement now matches creation rules).

  ### Non-breaking
  - `baseCost` removal is purely additive in terms of game behaviour; cost is now higher and better-scaled.
  - `missionRewardPerLvl` defaults to 0 for all structures not explicitly set in the data migration.

  ---

  ## License

  By making this pull request, I confirm that I have the right to waive copyright and related rights to my contribution, and agree that all
  copyright and related rights in my contributions are waived, and I acknowledge that the Studie-Tech ApS organization has the copyright to use and
   modify my contribution for perpetuity.

## Summary by CodeRabbit

* **New Features**
  * Village names updated (Current→Akikaze, Glacier→Hyorin, Shroud→Akasumi, Shine→Shirohana)
  * Buildings can grant mission/errand reward bonuses per upgrade level
  * UI shows "Next level provides" rewards for structures

* **Changes**
  * Building upgrade cost scaling revised
  * Hospital healing cost/duration adjusted and cost rounded
  * Quest rewards include village structure boosts
  * Music themes aligned with renamed villages
  * Stronger reserved-name username/squad/clan validation and legacy-name support
  * Safer concurrent-edit handling for squad/clan renames

[Review Change Stack](https://app.coderabbit.ai/change-stack/studie-tech/TheNinjaRPG/pull/1334)


MathiasGruber (commented):
**`hospitalSpeedupPerLvl` now reduces both heal time AND Ryo cost — confirm this is intentional**

`calcHealCost` now accepts a `boost` parameter and the router passes `hospitalSpeedupPerLvl` to it:

```ts
// hospital router line 256-274
const boost = getStrucBoost("hospitalSpeedupPerLvl", structures);
const finishAt = calcHealFinish({ user, boost });   // reduces heal time
...
cost = calcHealCost(user, boost);                   // also reduces Ryo cost
```

The structure field is named `hospitalSpeedupPerLvl` (time reduction), not something like `hospitalDiscountPerLvl`. If the intent is that levelling the hospital only speeds up healing rather than also giving a Ryo discount, `boost` should not be forwarded to `calcHealCost`. If both effects are intentional the field name and tooltip label (`"hospital speedup"` in `STRUCTURE_BONUSES`) should be updated to reflect both benefits, and the `StructureRewardEntries` tooltip should mention the cost reduction too.

---

MathiasGruber (commented):
**Upgrade dialog label says "mission reward" but the bonus applies to errands and medical quests too**

In `STRUCTURE_BONUSES` (the list powering the upgrade confirmation dialog):

```ts
{ key: "missionRewardPerLvl", label: "mission reward", unit: "%" },
```

But in `quest.ts` the boost is applied to `["mission", "crime", "medical"]` quest types **and** `"errand"` — matching the tooltip in `StructureRewardEntries` which correctly says `"Mission/Errand Rewards"`. The upgrade dialog label should be updated to match, e.g. `"mission/errand reward"`, so players see an accurate description before spending tokens.

---

MathiasGruber (commented):
**`bannedAnbuNames` includes multi-word names that can never match due to the preceding regex**

`bannedAnbuNames` contains `"Freedom State"` and `"Horizon"`, plus all `CoreVillages` and `LegacyVillageNames`. However, the `anbuNameField` schema applies `regex(/^[a-zA-Z0-9_]+$/)` before the `refine` check, which already rejects any name containing a space. This means `"Freedom State"` (and any other multi-word village name that might be added later) will never reach the `refine` — the regex error fires first, so those entries in the banned list are dead code.

If the goal is to also block single-word variants (e.g. `"FreedomState"`), those need to be added explicitly. If the intent is only to block exact matches including spaces, the regex guard needs to be relaxed to allow spaces and the banned-list check becomes the effective gate. Worth aligning the two validators so they behave consistently.

---

MathiasGruber (commented):
**Same dead-code issue in `usernameSchema` — `LegacyVillageNames` entries with spaces will never reach the `refine`**

`reservedUsernames` spreads both `LegacyVillageNames` (`["Shine", "Glacier", "Shroud", "Current"]`) and `CoreVillages`. Those four happen to be single words so they will be caught correctly. But the pattern is fragile: if any future village name contains a space it will silently fail to be blocked, because the `regex(/^[a-zA-Z0-9_]+$/)` fires before the `refine`.

Since usernames are already limited to `[a-zA-Z0-9_]+`, multi-word entries in the reserved list are unreachable dead code. Consider adding a note or an assertion that reserved names must be single-token, or normalise the check so it's robust against future additions.

---

MathiasGruber:

- PR's original migrations (`0009_chilly_preak.sql` and replaying as `0007_dizzy_bloodscream.sql` / `0008_orange_sabretooth.sql`): deleted; migration regenerated after rebase as `0008_moaning_gamma_corps.sql` (drops `VillageStructure.baseCost`, adds `VillageStructure.missionRewardPerLvl`).
- Source files (`schema.ts`, hospital/clan/anbu routers and validators, quest/threejs libs, etc.): auto-merged cleanly.

app/src/validators/anbu.ts
```diff
@@ -1,16 +1,44 @@
...
+  .regex(/^[a-zA-Z0-9_]+$/, {
+    error: "Alphanumeric and underscores only",
+  })
+  .refine(
+    (name) =>
+      !bannedAnbuNames.some((banned) => banned.toLowerCase() === name.toLowerCase()),
+    {
+      error: "This squad name is not allowed.",
+    },
+  );
```
MathiasGruber: DRY: this validator (banned-names list + alphanumeric regex + refine) is nearly identical to the one in `app/src/validators/clan.ts`. Consider extracting a shared `createReservedNameField()` helper (or a shared `bannedFactionNames` constant + reusable schema) so the two files don't drift in labels, error messages, or banned entries.

Also note that since the regex `^[a-zA-Z0-9_]+$` already rejects any string containing a space, `"Freedom State"` can never reach the `.refine()` check — it's effectively dead in the banned list (same applies in `clan.ts`).

MathiasGruber (commented):

app/src/validators/clan.ts
```diff
@@ -34,6 +39,9 @@ export const factionEditSchema = z.object({
     .trim()
     .min(3)
     .max(88)
+    .regex(/^[a-zA-Z0-9_]+$/, {
+      error: "Alphanumeric and underscores only",
+    })
```
MathiasGruber: Behavior change worth double-checking: `factionEditSchema` previously had no character-set restriction on `name`, and now requires `^[a-zA-Z0-9_]+$`. Existing clans/factions whose names contain spaces or other characters (created before this rule was added) will fail validation on *any* edit — including edits that only change the image, since `defaultValues.name` in `Clan.tsx` is populated from `clanData.name`. Either restrict the regex to apply only when the name actually changes, or confirm there are no existing clans whose current names would now be invalid.

app/src/layout/Building.tsx
```diff
@@ -146,6 +146,34 @@ const RestoreStructureButton = ({ structureId }: { structureId: string }) => {
...
+  { key: "itemDiscountPerLvl", label: "item shop discount", unit: "%" },
+  { key: "missionRewardPerLvl", label: "mission reward", unit: "%" },
+  { key: "patrolsPerLvl", label: "patrol slots", unit: "" },
+  { key: "ramenDiscountPerLvl", label: "ramen discount", unit: "%" },
+  { key: "regenIncreasePerLvl", label: "regen increase", unit: "%" },
+  { key: "sleepRegenPerLvl", label: "sleep regen", unit: "%" },
+  { key: "structureDiscountPerLvl", label: "structure upgrade discount", unit: "%" },
+  { key: "trainBoostPerLvl", label: "training boost", unit: "%" },
+  { key: "villageDefencePerLvl", label: "village defence", unit: "%" },
+];
```
MathiasGruber, line 207: DRY: this `STRUCTURE_BONUSES` array enumerates the same per-level attributes that `StructureRewardEntries` (below, around line 266) already lists with an if-chain. The two end up describing the same set of fields with subtly *different* labels (`"ANBU squads"` vs `"Anbu Squads"`, `"hospital speedup"` vs `"Hospital Speed"`, `"mission reward"` vs `"Mission/Errand Rewards"`, etc.), which is confusing for players and easy to let drift further. Consider unifying on a single attribute-metadata table (`{ key, label, unit, scale, format? }`) and having both the upgrade tooltip and the reward list render off it.

app/drizzle/migrations/meta/_journal.json
```diff
@@ -64,6 +64,13 @@
...
       "tag": "0008_busy_dreaming_celestial",
       "breakpoints": false
+    },
+    {
+      "idx": 9,
+      "version": "5",
+      "when": 1778608218943,
+      "tag": "0009_violet_zemo",
+      "breakpoints": false
     }
```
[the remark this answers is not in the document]
ifroshty: Is this fixed?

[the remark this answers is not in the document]
ifroshty: check now, latest commit

app/drizzle/migrations/0009_violet_zemo.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE `VillageStructure` ADD `missionRewardPerLvl` tinyint DEFAULT 0 NOT NULL;
```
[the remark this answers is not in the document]
ifroshty: All sibling *PerLvl columns on VillageStructure use signed tinyint (see 0000_omniscient_ben_parker.sql); singling this one out would be inconsistent and would drift from schema.ts (Drizzle's tinyint() has no unsigned modifier). No write path supplies a negative value either — the column is only written via Kage upgrades using positive constants. Beyond the scope of this PR

ifroshty:
@greptileai

app/src/server/api/routers/clan.ts
```diff
@@ -664,11 +666,22 @@ export const clanRouter = createTRPCRouter({
...
+        }
+      }
+      // Mutate (CAS on current name to avoid clobbering a concurrent rename)
+      const result = await ctx.drizzle
         .update(clan)
         .set({ name: input.name, image: image.avatar })
-        .where(eq(clan.id, fetchedClan.id));
+        .where(and(eq(clan.id, fetchedClan.id), eq(clan.name, fetchedClan.name)));
+      if (result.rowsAffected === 0) {
+        return errorResponse(`${groupLabel} was modified, please try again`);
```
MathiasGruber, line 688: Logic: on PlanetScale (default MySQL `affected_rows` semantics, no `CLIENT_FOUND_ROWS`), an UPDATE that matches the row but leaves every column unchanged returns `rowsAffected === 0`. So if a leader opens the edit dialog and submits without changing the name or image (or re-submits identical values), this branch fires the misleading `"<group> was modified, please try again"` toast even though no concurrent rename happened. The CAS predicate itself is fine, but the no-op case needs to be distinguished from a real conflict — e.g. short-circuit when `input.name === fetchedClan.name && image.avatar === fetchedClan.image`, or only treat `rowsAffected === 0` as a conflict when at least one column is actually being changed.

app/src/server/api/routers/anbu.ts
```diff
@@ -335,11 +339,22 @@ export const anbuRouter = createTRPCRouter({
...
+        }
+      }
+      // Mutate (CAS on current name to avoid clobbering a concurrent rename)
+      const result = await ctx.drizzle
         .update(anbuSquad)
         .set({ name: input.name, image: image.avatar })
-        .where(eq(anbuSquad.id, squad.id));
+        .where(and(eq(anbuSquad.id, squad.id), eq(anbuSquad.name, squad.name)));
+      if (result.rowsAffected === 0) {
+        return errorResponse("Squad was modified, please try again");
```
MathiasGruber, line 361: Same issue as `editClan`: PlanetScale (default MySQL `affected_rows` semantics) returns `rowsAffected === 0` for an UPDATE whose new values equal the existing ones. If the squad leader submits the edit form without changing the name or image, this returns `"Squad was modified, please try again"` even though no concurrent rename happened. Either short-circuit when `input.name === squad.name && image.avatar === squad.image`, or only treat `rowsAffected === 0` as a conflict when the request actually changes a column.

MathiasGruber (commented):
Quick code-quality pass: one logic concern around the rename CAS error handling.

app/src/server/api/routers/clan.ts
ifroshty: addressed https://github.com/studie-tech/TheNinjaRPG/pull/1334/commits/7a74d8832701f9adf30d57fe45bac7636fdad70d

app/src/server/api/routers/anbu.ts
ifroshty: addressed https://github.com/studie-tech/TheNinjaRPG/pull/1334/commits/7a74d8832701f9adf30d57fe45bac7636fdad70d

MathiasGruber (approved):
LGTM

PR #1336, opened 2026-04-16
Feature/987 restrict jutsu by item

### LINKED ISSUE #987: Feature: Restrict Jutsu by Item
No. This is a new feature to support restricting jutsu based on equipped bloodline items, similar to the existing bloodline restriction on jutsu.

When set, the jutsu should only be available if the player has the specified bloodline item equipped.

In the admin jutsu editor, this field should only list items that have a bloodline associated with them.
At runtime, jutsu availability and usage should validate that the required item is currently equipped.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

ifroshty:
# Pull Request

  ## What was implemented

  Adds a `requiredBloodlineItemId` field to jutsu, allowing admins to gate a jutsu's combat availability on a specific bloodline-associated item being equipped.

  **Changes:**
  - New nullable `requiredBloodlineItemId` column on the `jutsu` table (+ index + Drizzle relation)
  - `JutsuValidatorRawSchema` updated to include the new field
  - `checkJutsuBloodlineItem()` helper in `train.ts` — returns `false` if the required item is not equipped
  - `getBloodlineItemNames` tRPC endpoint on the item router (returns only items with a bloodline association)
  - Admin jutsu editor (`hooks/jutsu.ts`) wires up a "Required Bloodline Item" dropdown after the bloodline field
  - `initiateBattle` jutsu filter (`combat.ts`) enforces the check unconditionally for all battle types (mirrors bloodline check behavior)
  - `jutsus/page.tsx` surface: displays a warning when the required item is not equipped, and blocks jutsu equip/evolution when the requirement is unmet
  - 6 unit tests for `checkJutsuBloodlineItem` covering null, empty, undefined items, unequipped, equipped, and wrong-item cases

  ## Why

  Closes #987. Enables bloodline-item-gated jutsus — e.g. a jutsu that is only usable while a specific bloodline weapon is equipped.

  ## Screenshots

  N/A — admin dropdown addition; no visual regression on player-facing UI beyond the existing warning/lock pattern already used for bloodline checks.

  ## Breaking changes

  **Yes — requires 2 DB migrations:**

  1. `0007_black_payback.sql` — `ALTER TABLE Jutsu ADD requiredBloodlineItemId varchar(191)`
  2. `0008_aromatic_professor_monster.sql` — `CREATE INDEX Jutsu_requiredBloodlineItemId_idx ON Jutsu (requiredBloodlineItemId)`

  Run `make makemigrations && make migrate` (or apply via PlanetScale deploy request) before deploying.

  No API surface or existing jutsu behavior changes — all existing jutsus have `requiredBloodlineItemId = NULL`, which the helper treats as "no restriction".

  ---

  ## License

  By making this pull request, I confirm that I have the right to waive copyright for the submitted work, and agree that all contributions made by
  me to this project are, and will remain, released under the project's open-source license

## Summary by CodeRabbit

* **New Features**
  * Jutsu can now require a specific bloodline item to be equipped; affected jutsu are unavailable until the item is equipped.
  * Added a "Required Bloodline Item" select when configuring jutsu.

* **Behavior Changes**
  * Action warnings now show "You do not have the required bloodline item equipped."
  * Training ignores the bloodline-item requirement so training lists remain unchanged.
  * Loadouts now prevent re-equipping jutsu if the prerequisite item was removed.

* **Tests & Backend**
  * Server-side validation and tests added to enforce the new requirement.


MathiasGruber (commented):
**Bug: `startTraining` auto-equips a jutsu with `requiredBloodlineItemId` without verifying the item is equipped**

In the `startTraining` mutation (lines 1044–1050), `canAutoEquip` only checks slot caps (total, residual, pierce, event, barrier, stun). It never calls `checkJutsuBloodlineItem`, so a freshly-learned jutsu that requires a specific bloodline item is auto-equipped into the player's active loadout even if that item is not equipped.

The `toggleEquip` mutation (line 1200) correctly guards this with:
```ts
if (!isEquipped && !checkJutsuBloodlineItem(userjutsuObj.jutsu, user.items)) {
  return errorResponse("You don't have the required bloodline item equipped for this jutsu");
}
```

The same guard needs to be added to `canAutoEquip`:

```ts
const canAutoEquip =
  curEquip < maxEquip &&
  checkJutsuBloodlineItem(info, user.items) &&   // <-- add this
  (!jutsuHasResidual || residualJutsus.length < JUTSU_MAX_RESIDUAL_EQUIPPED) &&
  (!jutsuHasPierce || pierceJutsus.length < JUTSU_MAX_PIERCE_EQUIPPED) &&
  (!jutsuIsEvent || eventJutsus.length < JUTSU_MAX_EVENT_EQUIPPED) &&
  (!jutsuHasBarrier || barrierJutsus.length < JUTSU_MAX_BARRIER_EQUIPPED) &&
  (!jutsuHasStun || stunJutsus.length < JUTSU_MAX_STUN_EQUIPPED);
```

`user` here comes from `fetchUpdatedUser`, which already loads `items` (equipped items only, line 2276 of `profile.ts`), so `user.items` is available without any extra query.

---

MathiasGruber (commented):
**`canUseJutsu` does not include the bloodline item check, creating a divergence**

`canUseJutsu` in `app/src/libs/train.ts` (lines 256–270) checks rank, level, jutsu rank, village, bloodline, and elements — but not `requiredBloodlineItemId`. As a result, `canTrainJutsu` (which delegates to `canUseJutsu`) also omits the check.

This means the `startTraining` guard at line 971:
```ts
if (!canTrainJutsu(info, user) && !info.parentJutsuId)
  return errorResponse("Jutsu not for you");
```
does **not** block training initiation for a jutsu whose bloodline item requirement the user doesn't satisfy. Learning is intentionally allowed without the item (the item gates equipping and in-combat use), but this divergence is a maintenance hazard: callers of `canUseJutsu` naturally expect it to be the complete gate for "can this user use this jutsu", yet the bloodline item requirement is silently excluded.

Consider either:
1. Merging `checkJutsuBloodlineItem` into `canUseJutsu` (since `userdata` already carries `items` in `UserWithRelations`), and only skipping the check for training eligibility explicitly, or
2. Renaming the existing function to `canLearnJutsu` / `meetsBaseRequirements` to make the scope unambiguous, and keeping a separate `canEquipJutsu` / `canUseJutsuInCombat` that includes the item check.

Without this, any future caller that relies on `canUseJutsu` to be the complete "can use" predicate will silently bypass the bloodline item gate.

---

MathiasGruber (commented):
**`getBloodlineItemNames` returns all bloodline items when `bloodlineId` is an empty string**

In `app/src/server/api/routers/item.ts` lines 111–113:
```ts
where: (table, { and, eq }) =>
  input.bloodlineId
    ? eq(table.bloodlineId, input.bloodlineId)
    : isNotNull(table.bloodlineId),
```

`input.bloodlineId` is typed as `string | null | undefined`. When the jutsu editor form has no bloodline selected, `useWatch` returns `""` (empty string) rather than `null` or `undefined`. An empty string is falsy in JavaScript, so the condition falls through to the `isNotNull` branch and returns every bloodline item in the database — not an empty list as the admin would expect.

This is confirmed in `app/src/hooks/jutsu.ts` line 52:
```ts
api.item.getBloodlineItemNames.useQuery({ bloodlineId: selectedBloodlineId });
```
where `selectedBloodlineId` comes from `useWatch` and will be `""` when the bloodline field is cleared.

The fix is to treat empty string the same as absent:
```ts
where: (table, { eq }) =>
  input.bloodlineId && input.bloodlineId !== ""
    ? eq(table.bloodlineId, input.bloodlineId)
    : isNotNull(table.bloodlineId),
```
Or validate/transform on the input side with `.transform(v => v || null)` in the Zod schema.

---

MathiasGruber (commented):
**`selectJutsuLoadout` re-equips loadout jutsus without checking `requiredBloodlineItemId`**

`selectJutsuLoadout` in `app/src/server/api/routers/jutsu.ts` (lines 2090–2124) bulk-sets `equipped = 1` for every jutsu ID in the stored loadout:
```ts
client
  .update(userJutsu)
  .set({
    equipped:
      loadout.jutsuIds.length > 0
        ? sql`CASE WHEN ${inArray(userJutsu.jutsuId, loadout.jutsuIds)} THEN 1 ELSE 0 END`
        : false,
  })
  .where(eq(userJutsu.userId, user.userId)),
```

If a loadout was saved while the player had the required bloodline item equipped, then the item is later unequipped (or breaks), switching back to that loadout will re-equip the gated jutsu without any bloodline item validation. This is the same class of bypass as the auto-equip bug in `startTraining`.

The fix is to filter `loadout.jutsuIds` before the bulk update, removing any jutsus whose `requiredBloodlineItemId` isn't currently satisfied. This requires joining jutsu data and user items in `selectJutsuLoadout`, or performing the validation in the caller before invoking `selectJutsuLoadout`.

---

MathiasGruber:

- `app/drizzle/migrations/0007_black_payback.sql` (PR's original migration): deleted; migration regenerated after rebase as `0008_loving_sue_storm.sql` containing the `Jutsu.requiredBloodlineItemId` column add + index.
- `app/drizzle/schema.ts`: auto-merged; both `main`'s changes and PR's `requiredBloodlineItemId` field retained.

app/src/server/api/routers/jutsu.ts
```diff
@@ -1197,6 +1202,11 @@ export const jutsuRouter = createTRPCRouter({
       if (!isEquipped && !canUseJutsu(userjutsuObj.jutsu, user)) {
         return errorResponse("You cannot equip this jutsu due to missing requirements");
       }
+      if (!isEquipped && !checkJutsuBloodlineItem(userjutsuObj.jutsu, user.items)) {
```
MathiasGruber, line 1242: The bloodline-item gate was added here for `toggleEquip` and at the evolution check, but the same gate is missing from `startTraining` (around line 1049-1065). When a user first learns a gated jutsu, `canAutoEquip` is computed purely from slot/category limits, so the new `userJutsu` row is inserted with `equipped: canAutoEquip = true` even if the required bloodline item is not equipped.

Result: a player with no `Kaiken` who learns `Sand Clone` ends up with the jutsu shown as equipped on the jutsu page. Combat and the re-equip path correctly filter it out, but the UI state is inconsistent until the user manually unequips and the equip block kicks in.

Suggested fix: gate `canAutoEquip` on `checkJutsuBloodlineItem(info, user.items)` as well, e.g.

```ts
const canAutoEquip =
  curEquip < maxEquip &&
  checkJutsuBloodlineItem(info, user.items) &&
  (!jutsuHasResidual || residualJutsus.length < JUTSU_MAX_RESIDUAL_EQUIPPED) &&
  // ...
```

`user.items` is already loaded by `fetchUpdatedUser` with `where: ne(userItem.equipped, "NONE")`, and `checkJutsuBloodlineItem` itself filters on `equipped !== "NONE"` and usable durability, so the call is safe here.

MathiasGruber (commented):
Review of automated browser-test findings.

app/src/hooks/jutsu.ts
```diff
@@ -43,6 +43,14 @@ export const useJutsuEditForm = (data: Jutsu, refetch: () => void) => {
   const { data: villages, isPending: l2 } = api.village.getAllNames.useQuery(undefined);
   const { data: jutsus, isPending: l4 } = api.jutsu.getAllNames.useQuery(undefined);
 
+  // Watch bloodlineId to filter bloodline items to only those for the selected bloodline
+  const selectedBloodlineId = useWatch({
+    control: form.control,
+    name: "bloodlineId",
+  });
+  const { data: bloodlineItems, isPending: l5 } =
+    api.item.getBloodlineItemNames.useQuery({ bloodlineId: selectedBloodlineId });
```
MathiasGruber, line 53: When the admin changes `bloodlineId` on a jutsu, the `bloodlineItems` dropdown re-filters to the newly selected bloodline, but the form's current `requiredBloodlineItemId` value is not reset. The admin can therefore save a jutsu whose `requiredBloodlineItemId` points at an item belonging to a different bloodline (or no bloodline at all), and that mismatch will silently slip past the UI.

Consider clearing the field whenever `selectedBloodlineId` changes, e.g.:
```ts
useEffect(() => {
  form.setValue("requiredBloodlineItemId", null, { shouldDirty: true });
}, [selectedBloodlineId, form]);
```
or at minimum validating server-side in `jutsu.update` that the chosen item's `bloodlineId` matches the jutsu's `bloodlineId`.

app/src/server/api/routers/combat.ts
```diff
@@ -2630,6 +2630,12 @@ export const processUsersForBattle = async (
           .forEach((e) => {
             if ("aiId" in e) allSummons.push(e.aiId);
           });
+        // Bloodline item enforcement applies to all battle types including ranked,
+        // mirroring bloodline check behavior (unlike checkJutsuItems which is waived in ranked).
+        // Not if required bloodline item is not equipped
+        if (!checkJutsuBloodlineItem(userjutsu.jutsu, user.items) && !user.isAi) {
```
MathiasGruber: `checkJutsuBloodlineItem` runs *after* the summons-collection block above (lines 2627-2632), so a jutsu the user is about to be filtered out of (because they have no required bloodline item equipped) still contributes its `summon` effect's `aiId` to `allSummons`. That mirrors the pre-existing positioning of the bloodline check below, but it now means a player who unequips their keystone can still cause summon AIs to be loaded into the battle.

Move this check (and ideally the bloodline check on lines 2640-2644) above the summons-collection block, so summons are only collected for jutsus that survive every gate.

app/src/libs/train.ts
```diff
@@ -204,6 +205,19 @@ export const checkJutsuItems = (
...
+  userItems: Pick<UserItemWithItem, "itemId" | "equipped" | "durability">[] | undefined,
+): boolean => {
+  if (!jutsu.requiredBloodlineItemId) return true;
+  return !!userItems?.some(
+    (ui) =>
+      ui.itemId === jutsu.requiredBloodlineItemId &&
+      ui.equipped !== "NONE" &&
+      ui.durability > DURABILITY_USABILITY_THR,
+  );
+};
```
MathiasGruber, line 231: `canUseJutsu` is the canonical "can the player wield this jutsu?" predicate used by `canTrainJutsu`, the equip endpoint, the evolution endpoint, the combat filter, and `traininggrounds/page.tsx`. The new bloodline-item gate is *not* folded into it, so every caller has to remember to also call `checkJutsuBloodlineItem` separately — and several already miss it (e.g. `startTraining`/`canTrainJutsu` path, and `app/src/app/traininggrounds/page.tsx:737` which only checks `canUseJutsu`).

Consider extending `canUseJutsu` to take the user's items (the `UserWithRelations` it already receives carries `items` on most paths) and AND-ing in `checkJutsuBloodlineItem`. The duplicated guards in `jutsu.ts` (evolution + equip) and `combat.ts` could then collapse to a single `canUseJutsu(...)` call, and future callers can't forget.

app/src/server/api/routers/combat.ts
```diff
@@ -2630,6 +2630,12 @@ export const processUsersForBattle = async (
           .forEach((e) => {
             if ("aiId" in e) allSummons.push(e.aiId);
           });
+        // Bloodline item enforcement applies to all battle types including ranked,
+        // mirroring bloodline check behavior (unlike checkJutsuItems which is waived in ranked).
+        // Not if required bloodline item is not equipped
```
MathiasGruber: Minor readability: the comment block here repeats itself — "Bloodline item enforcement applies to all battle types..." and then "Not if required bloodline item is not equipped" describe the same check. Per CLAUDE.md ("Natural Comments Only"), one concise comment is enough; drop the second line.

app/src/server/api/routers/item.ts
```diff
@@ -87,6 +99,24 @@ export const itemRouter = createTRPCRouter({
...
+      mcp: {
+        enabled: true,
+        description: "Get names of items associated with a bloodline",
+      },
+    })
+    .input(z.object({ bloodlineId: z.string().nullish() }))
+    .query(async ({ ctx, input }) => {
+      return await ctx.drizzle.query.item.findMany({
+        columns: { id: true, name: true },
+        where: (table, { and, eq }) =>
```
MathiasGruber: `and` is destructured but never used inside the `where` callback. Drop it: `where: (table, { eq }) => ...`.

MathiasGruber (commented):
Code review: logic, readability, DRY.

app/src/hooks/jutsu.ts
ifroshty: Fixed — canAutoEquip now ANDs in checkJutsuBloodlineItem(info, user.items), so a
  freshly-learned gated jutsu is inserted with equipped: false when the required item
  isn't equipped. user comes from fetchUpdatedUser, so items is already loaded — no
  extra query.

app/src/server/api/routers/combat.ts
ifroshty: Fixed — both the bloodline-item gate and the bloodline-match gate now run before the
  summon-aiId collection block, so summons are only collected for jutsus that survive
  every gate. The item gate stays outside the battleType !== RANKED conditional, so
  ranked still enforces it.

app/src/server/api/routers/combat.ts
ifroshty: Fixed — dropped the duplicate line; one concise comment remains.

app/src/server/api/routers/item.ts
ifroshty: Fixed — removed and from the where destructuring (and the now-unused isNotNull import)
  as part of the empty-string fix in the same endpoint.

app/src/libs/train.ts
ifroshty: Folded it in: canUseJutsu now takes ignoreBloodlineItem = false and ANDs in
  checkJutsuBloodlineItem(jutsu, userdata.items) by default, so it's the complete gate
  and future callers can't silently bypass it. canTrainJutsu passes true (learning/
  leveling is intentionally item-free), and traininggrounds/page.tsx does the same for
  the leveling display.

  I kept the explicit checkJutsuBloodlineItem calls at the equip and evolution endpoints
  (passing ignoreBloodlineItem: true to canUseJutsu there) so they keep the specific
  "you don't have the required bloodline item equipped" message instead of the generic
  "missing requirements". Combat also keeps its explicit check because the item must be
  enforced in ranked, where canUseJutsu isn't called. Net result: no behavior change at
  the current call sites, but the default is now the safe one.

MathiasGruber:

**Conflicts resolved:**
- **Migrations**: The PR's migration files conflicted with new migrations on `main` (`0008`–`0010`). Per the migration-conflict policy, the PR's own migration files were dropped, the migration metadata (`_journal.json` / snapshots) was reset to match `main`, and a fresh migration (`0011_misty_apocalypse.sql`) was regenerated via `make makemigrations`. It cleanly adds the `Jutsu.requiredBloodlineItemId` column and its index.
- **schema.ts**: Auto-merged; the PR's jutsu schema additions are preserved on top of the latest `main`.

app/src/server/api/routers/combat.ts
```diff
@@ -942,7 +942,14 @@ export const combatRouter = createTRPCRouter({
       const [jutsuLoadoutResult, itemLoadoutResult] = await Promise.all([
```
MathiasGruber: **Loadout staleness: `selectJutsuLoadout` validates against pre-switch items.**

When a combat-lobby request supplies both `jutsuLoadoutId` and `itemLoadoutId`, the two `select*Loadout` helpers run in parallel via `Promise.all`. `selectJutsuLoadout` therefore filters its loadout against `useritems` as they existed *before* `selectItemLoadout` swapped equipment.

Two real bypasses:
- Switching to an item loadout that drops the required bloodline item while simultaneously switching to a jutsu loadout containing a gated jutsu: the gated jutsu is re-equipped because the item still appears equipped in the snapshot passed to `selectJutsuLoadout`.
- The opposite: a previously-gated jutsu that becomes valid (new item loadout equips the required item) is wrongly rejected.

Run the two selects sequentially so the jutsu pass sees the updated item state, e.g. await `selectItemLoadout` first, then feed its resulting items into `selectJutsuLoadout`. (Originally raised by CodeRabbit on 2026-06-02 as an "outside diff range" finding; not yet addressed.)

app/src/libs/train.ts
```diff
@@ -204,6 +205,19 @@ export const checkJutsuItems = (
...
+  jutsu: Jutsu,
+  userItems: Pick<UserItemWithItem, "itemId" | "equipped" | "durability">[] | undefined,
+): boolean => {
+  if (!jutsu.requiredBloodlineItemId) return true;
+  return !!userItems?.some(
+    (ui) =>
+      ui.itemId === jutsu.requiredBloodlineItemId &&
+      ui.equipped !== "NONE" &&
+      ui.durability > DURABILITY_USABILITY_THR,
+  );
```
MathiasGruber: **Durability check diverges from combat: low-durability bloodline weapons.**

`checkJutsuBloodlineItem` rejects every equipped required item whose `durability <= DURABILITY_USABILITY_THR`. Combat-side (`processUsersForBattle` in `combat.ts` ~2746-2754, and `process.ts` durability decrement) only forces `equipped = "NONE"` for low-durability `ARMOR` / `ACCESSORY` / `KEYSTONE`. Weapons and other item types stay usable at low durability.

Net effect: a player whose required bloodline item is a low-durability weapon can still use it in combat, but this helper blocks them from equipping or evolving the gated jutsu, and `selectJutsuLoadout` strips it. Fix by widening the parameter to include `item.itemType` and only enforcing the durability floor when the item type is one of the strip-on-low-durability types — or, equivalently, mirror combat's exact behavior. (Originally raised by CodeRabbit on 2026-06-02 as an "outside diff range" finding; not yet addressed.)

app/src/libs/combat/util.ts
```diff
@@ -286,20 +292,33 @@ export const hydrateUserForQuests = (battle: CompleteBattle, user: BattleUserSta
...
+      createdAt: now,
+      updatedAt: now,
+      userId: user.userId,
+      storedAtHome: false,
+      craftingFinishedAt: null,
+      isInAuction: false,
+      item: {
+        id: bi.itemId,
+        itemType: itemData?.itemType ?? "CONSUMABLE",
+        maxDurability: itemData?.maxDurability ?? 0,
```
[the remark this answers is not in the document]
ifroshty: I looked into this and I’m leaving the code as-is.

### 1. `createdAt` / `updatedAt` cannot come from `bi`

`BattleUserItem` is a trimmed battle-reference type. It only carries combat state like `id`, `itemId`, `quantity`, `equipped`, `durability`, `dropChancePerc`, `lastUsedRound`, and `originalCooldown`.

It does not include `createdAt` or `updatedAt`, so `bi.createdAt` / `bi.updatedAt` would be a TS2339 compile error. The `new Date()` values are not overwriting original timestamps; they are filling required `UserItem` fields that the source type does not carry.

### 2. The `getItem` fallback is intentional

`extraState.items` is populated conditionally during `initiateBattle`, so stale/deleted item definitions can legitimately resolve to `undefined`.

`getItem` is the non-throwing lookup. `getItemOrThrow` exists for paths that should hard-fail. Throwing here would crash an in-progress combat round for a player with a stale item reference.

The fallback `itemType`, `maxDurability`, and timestamps are inert here; `getNewTrackers` does not read them. They only satisfy the `UserWithRelations.items` shape.

This is graceful degradation by design, so no change is needed.

MathiasGruber:

**Conflicts resolved:**
- The only conflicts were in the Drizzle migration files (`app/drizzle/migrations/meta/_journal.json` and `meta/0011_snapshot.json`). `main` had advanced to migration `0011`, colliding with this branch's own migration.
- Per the standard procedure, the stale branch migration files were dropped, the migrations directory was reset to match `main`, and a fresh migration (`0012_dashing_storm.sql`) was regenerated from `schema.ts` via `make makemigrations`. It cleanly adds the `Jutsu.requiredBloodlineItemId` column and its index — no other schema changes.

No application/source code conflicts occurred; all commits replayed cleanly. The branch is now up to date with `main`.

MathiasGruber (commented):
One follow-up — see inline.

app/drizzle/migrations/0011_misty_apocalypse.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `Jutsu` ADD `requiredBloodlineItemId` varchar(191);
+CREATE INDEX `Jutsu_requiredBloodlineItemId_idx` ON `Jutsu` (`requiredBloodlineItemId`);
```
MathiasGruber, line 2: Orphaned migration file from the rebase — this file is byte-identical to `0012_dashing_storm.sql` (same `ALTER TABLE Jutsu ADD requiredBloodlineItemId` and index creation) but is not registered in `app/drizzle/migrations/meta/_journal.json` (journal only contains `0011_curvy_lady_bullseye` and `0012_dashing_storm`).

Drizzle's migrator reads the journal so it won't double-apply, but any tooling that scans the migrations directory directly will hit a duplicate column / duplicate index error. Please delete `0011_misty_apocalypse.sql`.

PR #1341, opened 2026-04-18
fix: honor reduced combat AP costs

### LINKED ISSUE #1015: Decreases AP cost (Time Dialation) using original AP to see if usable
**Describe the bug**
When decrease AP usage is in effect, it properly reduces the AP of compatible jutsu when used. The problem is its using the original AP instead of the lower AP to determine if the jutsu can be used. 

This worked properly prior to the recent pvp changes that were made. 

**To Reproduce**

1) user has -10 AP cost
2) user uses a 60 AP jutsu (now 50).
3) the check sees that there is enough AP and the jutsu goes through 
4) user has 50 remaining AP
5) user attempts another 60 AP jutsu (now 50)
6) the check fails because its looking for 60 available AP. 

**Expected behavior**
The AP check should recognize the lower AP values so that the AP reduction can be properly used. 

**Further Information**
It could be you already have an idea of which code is at fault for this issue, but are not confident enough to attempt to fix it yourself; if so, please list any such additional information here.

**Screenshots**

TheNinja-RPG:
## Summary
- reuse the shared combat AP-cost calculation for the selector's usability checks
- keep Time Dilation and Time Compression UI gating aligned with the actual combat execution path
- add a regression test covering a 60 AP jutsu becoming usable at 50 AP under Time Dilation

Fixes #1015

## Test Plan
- `NEXT_PUBLIC_PUSHER_APP_KEY=test NEXT_PUBLIC_PUSHER_APP_CLUSTER=test NEXT_PUBLIC_BASE_URL=http://localhost:3000 NODE_ENV=test npx vitest run tests/libs/combat/actions.test.ts`
- `npx biome check src/libs/combat/actions.ts src/layout/CombatActions.tsx src/app/combat/page.tsx tests/libs/combat/actions.test.ts`
- `npm run check` currently fails on unrelated pre-existing Biome diagnostics elsewhere in the repo

## Summary by CodeRabbit

* **New Features**
  * Combat action costs now reflect the current battle and the acting user, so active temporal effects (e.g., time dilation/compression) can change ability AP costs and selection behavior.
  * UI now prevents or auto-deselects actions when they become unaffordable under current battle conditions.
* **Tests**
  * Added tests covering AP cost calculations and affordability under temporal effects.

### DIFF
--- a/app/tests/libs/combat/actions.test.ts
+++ b/app/tests/libs/combat/actions.test.ts
@@ -1,0 +1,186 @@
+import { describe, expect, it, vi } from "vitest";
+
+/** Keep AP helpers testable without loading the full combat graph (process, hex, Three). */
+vi.mock("@/libs/combat/process", () => ({
+  applyEffects: vi.fn(() => ({
+    newBattle: {},
+    actionEffects: [],
+  })),
+  checkFriendlyFire: vi.fn(() => true),
+}));
+
+vi.mock("@/libs/hexgrid", () => ({
+  getPossibleActionTiles: vi.fn(() => new Set()),
+  PathCalculator: vi.fn(),
+}));
+
+/** Avoid executing real db/env when transitive imports touch `@/server/db`. */
+vi.mock("@/server/db", () => ({
+  drizzleDB: {},
+}));
+
+import {
+  actionPointsAfterAction,
+  getActionPointCost,
+} from "@/libs/combat/actions";
+import type { ElementName } from "@/drizzle/constants";
+import type { CombatAction, ReturnedBattle, ReturnedUserState, UserEffect } from "@/libs/combat/types";
+
+const makeBattle = (effects: UserEffect[]): ReturnedBattle =>
+  ({
+    id: "battle-1",
+    activeUserId: "user-1",
+    createdAt: new Date("2026-01-01T00:00:00Z"),
+    updatedAt: new Date("2026-01-01T00:00:00Z"),
+    roundStartAt: new Date("2026-01-01T00:00:00Z"),
+    background: "default",
+    width: 10,
+    height: 10,
+    battleType: "COMBAT",
+    version: 1,
+    round: 1,
+    rewardScaling: 1,
+    forceKeepPools: false,
+    groundEffects: [],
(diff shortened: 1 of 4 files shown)


MathiasGruber (commented):
Reviewed `getActionPointCost`, `getTemporalApDelta`, `ActionOption`, and the new test file. Three findings below — one logic bug, one dead-code/misleading-signature issue, and one test DRY issue.

---

app/tests/libs/combat/actions.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    cooldown: 0,
+    originalCooldown: 0,
+    effects: [],
+    data: {
+      effects: [{ type: "damage", elements: ["Fire"] }],
+    },
+  }) as unknown as CombatAction;
+
+/** Fire only on normalized top-level `effects` (no embedded `data.effects`). */
+const makeJutsuFireTopLevelOnly = (): CombatAction =>
```
MathiasGruber: `makeJutsuAction` and `makeJutsuFireTopLevelOnly` share ~90% of their structure. The only meaningful difference is where the Fire element is declared (`data.effects` vs top-level `effects`) and whether `data` is `undefined`. A single factory with an options param would eliminate the duplication and make the intent of each test case clearer:

```ts
const makeJutsuAction = (opts?: {
  topLevelElements?: ElementName[];
  dataElements?: ElementName[];
}): CombatAction =>
  ({
    // ...shared fields...
    effects: opts?.topLevelElements
      ? [{ type: "damage", elements: opts.topLevelElements }]
      : [],
    data: opts?.dataElements
      ? { effects: [{ type: "damage", elements: opts.dataElements }] }
      : undefined,
  }) as unknown as CombatAction;
```

Then tests read as:
```ts
makeJutsuAction({ dataElements: ["Fire"] })       // covers data.effects path
makeJutsuAction({ topLevelElements: ["Fire"] })   // covers top-level effects path
```

---

MathiasGruber (commented):
Three findings across the new logic and tests.

---

MathiasGruber (approved):
lgtm

PR #1342, opened 2026-04-21
Item shop qol

Phrosfire:
# Pull Request

Changes:
- Better modernizes the item shop, souvenir shop, and the blackmarket item shop

**Important note**
I checked the tutorial, and you are still able to proceed through the item shop segment with this update.  While it tells you to buy the shuriken, it doesn't not point at it for you like it does other steps, I'm not sure if that was already the case for it or not.

## License

## Summary by CodeRabbit

* **New Features**
  * Modern catalog UI with category tabs, card-based browsing, hero text/badges and per-shop catalog search/list controls
  * Dedicated catalog search and filter trigger for quicker item discovery
  * Affordability badges on item cards and clearer shop naming/description (Seichi → Silver where applicable)

* **Improvements**
  * Improved filtering defaults and option to hide item-type selector
  * Purchase quantity resets when switching or settling items; clearer loading/empty/update catalog messages

### DIFF
--- a/app/src/layout/ItemShopFiltering.tsx
+++ b/app/src/layout/ItemShopFiltering.tsx
@@ -1,3 +1,4 @@
+import { useMemo } from "react";
 import {
   BattleUsageTypes,
   ItemRarities,
@@ -19,19 +20,28 @@ interface ItemShopFilteringProps {
   state: ItemShopFilteringState;
   defaultType: ItemType;
   restrictTypes?: ItemType[];
+  /** When true, item type is controlled elsewhere (e.g. shop tabs); omit Type from the filter popover. */
+  hideItemTypeField?: boolean;
+  /** Popover trigger button id (unique when several shops on one page). */
+  filterTriggerId?: string;
 }
 
-const makeItemShopSchema = (types: readonly string[]) =>
-  defineFilteringSchema({
+const makeItemShopSchema = (types: readonly string[], itemTypeDefault?: ItemType) => {
+  const first = (types[0] ?? "WEAPON") as ItemType;
+  const initialItemType =
+    itemTypeDefault && types.includes(itemTypeDefault) ? itemTypeDefault : first;
+  return defineFilteringSchema({
     fields: [
       {
         id: "itemType",
         label: "Type",
         type: "single-select",
-        defaultValue: types[0] ?? "WEAPON",
+        defaultValue: initialItemType,
         includeNone: false,
         emptyValues: [],
         options: types.map((t) => ({ value: t, label: t })),
+        visibleIf: (ctx) =>
+          !(ctx as { hideItemType?: boolean } | undefined)?.hideItemType,
       },
       { id: "name", label: "Name", type: "text", defaultValue: "" },
       {
@@ -82,17 +92,25 @@ const makeItemShopSchema = (types: readonly string[]) =>
       },
     ] as const,
   });
+};
 
 const ItemShopFiltering: React.FC<ItemShopFilteringProps> = (props) => {
-  let categories = Object.values(ItemTypes) as string[];
-  if (props.restrictTypes)
-    categories = categories.filter((t) => props.restrictTypes?.includes(t as ItemType));
-  const schema = makeItemShopSchema(categories);
+  const restrictKey = props.restrictTypes?.join() ?? "";
+  const schema = useMemo(() => {
+    let categories = Object.values(ItemTypes) as string[];
+    if (props.restrictTypes?.length) {
+      categories = categories.filter((t) =>
+        props.restrictTypes?.includes(t as ItemType),
+      );
+    }
+    return makeItemShopSchema(categories, props.defaultType);
+  }, [restrictKey, props.defaultType]);
   return (
     <ContentFiltering
       schema={schema}
       state={props.state.cf}
-      triggerButtonId="filter-item"
+      triggerButtonId={props.filterTriggerId ?? "filter-item"}
+      context={{ hideItemType: props.hideItemTypeField }}
     />
   );
 };
@@ -103,12 +121,11 @@ export const getShopFilter = (state: ItemShopFilteringState) =>
   buildFilter(state.cf, makeItemShopSchema(Object.values(ItemTypes)));
 
 export const useShopFiltering = (defaultType: ItemType) => {
-  const cf = useContentFiltering(makeItemShopSchema(Object.values(ItemTypes)));
-  if (!cf.values.itemType) {
-    (cf.setters.itemType as React.Dispatch<React.SetStateAction<ItemType>>)(
-      defaultType,
-    );
-  }
+  const schema = useMemo(
+    () => makeItemShopSchema(Object.values(ItemTypes), defaultType),
+    [defaultType],
+  );
+  const cf = useContentFiltering(schema);
   return {
     ...cf.values,
     cf,
--- a/app/src/layout/ItemShopFiltering.tsx
+++ b/app/src/layout/ItemShopFiltering.tsx
@@ -26,6 +26,8 @@ interface ItemShopFilteringProps {
   filterTriggerId?: string;
 }
 
+type ItemShopFilterContext = { hideItemType?: boolean };
+
 const makeItemShopSchema = (types: readonly string[], itemTypeDefault?: ItemType) => {
   const first = (types[0] ?? "WEAPON") as ItemType;
   const initialItemType =
@@ -40,8 +42,7 @@ const makeItemShopSchema = (types: readonly string[], itemTypeDefault?: ItemType
         includeNone: false,
         emptyValues: [],
         options: types.map((t) => ({ value: t, label: t })),
-        visibleIf: (ctx) =>
-          !(ctx as { hideItemType?: boolean } | undefined)?.hideItemType,
+        visibleIf: (ctx) => !(ctx as ItemShopFilterContext).hideItemType,
       },
       { id: "name", label: "Name", type: "text", defaultValue: "" },
       {
--- a/app/src/layout/Shop.tsx
+++ b/app/src/layout/Shop.tsx
@@ -1,3 +1,4 @@
+import type { LucideIcon } from "lucide-react";
 import {
   Check,
   CircleAlert,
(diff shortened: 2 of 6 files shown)


app/src/layout/Shop.tsx
```diff
@@ -36,23 +59,169 @@ interface ShopProps {
...
+  /** Village item shop / souvenirs: auction-style floor layout, category tabs, card grid. */
+  useItemTypeTabs?: boolean;
+  /** When useItemTypeTabs: overrides hero headline (default: village / souvenir copy). */
+  catalogHeroTitle?: string;
+  catalogHeroDescription?: string;
+  catalogHeroBadgeCatalog?: string;
+  /** When useItemTypeTabs: id for search field + label (avoid duplicates if multiple shops on one page). */
+  catalogSearchId?: string;
+  /** Catalog list wrapper id; use `false` to omit (e.g. black market). Default village: tutorial-itemshop. */
+  catalogTutorialId?: string | false;
```
MathiasGruber: I'm a bit unsure, but seems to me this could be breaking the tutorial step for the user buying the shuriken? please confirm that this still works as intended

app/src/layout/Shop.tsx
```diff
@@ -183,122 +480,205 @@ const Shop: React.FC<ShopProps> = (props) => {
...
+                        <Loader explanation="Updating catalog…" />
+                      </div>
+                    )}
+                  </>
+                )}
+              </div>
+            </>
+          )}
+
+          {!useModern && (
```
MathiasGruber: No longer used? I'm confused, do we need to support two display types?

- Remove the !useModern branch and the useItemTypeTabs/useModern flag.
- Remove useItemTypeTabs prop from all three call sites.
- Drop the now-unused ActionSelector import if no longer referenced.

app/src/layout/ItemShopFiltering.tsx
```diff
@@ -19,19 +20,28 @@ interface ItemShopFilteringProps {
...
       {
         id: "itemType",
         label: "Type",
         type: "single-select",
-        defaultValue: types[0] ?? "WEAPON",
+        defaultValue: initialItemType,
         includeNone: false,
         emptyValues: [],
         options: types.map((t) => ({ value: t, label: t })),
+        visibleIf: (ctx) =>
```
MathiasGruber: context is always an object (never undefined), so the | undefined half of the cast is belt‑and‑braces. Declare a named type once and share it:

```typescript
type ItemShopFilterContext = { hideItemType?: boolean };
visibleIf: (ctx) => !(ctx as ItemShopFilterContext).hideItemType,
```

app/src/layout/Shop.tsx
```diff
@@ -183,122 +480,205 @@ const Shop: React.FC<ShopProps> = (props) => {
...
+                        role="tab"
+                        aria-selected={filteringState.itemType === t}
+                        className={cn(
+                          "flex flex-col items-center justify-center gap-0.5 rounded-md px-0.5 py-1.5 font-semibold text-[10px] transition-all sm:flex-row sm:gap-2 sm:rounded-lg sm:px-1 sm:py-3 sm:text-xs",
+                          filteringState.itemType === t
+                            ? "bg-background text-foreground shadow-sm ring-1 ring-border/60"
+                            : "text-muted-foreground hover:bg-background/60 hover:text-foreground",
+                        )}
+                        onClick={() => {
+                          filteringState.setItemType(t);
```
MathiasGruber, line 553: If the user taps the already‑active tab, the selected card/modal is needlessly cleared. Guard with if (filteringState.itemType === t) return;.

app/src/layout/Shop.tsx
```diff
@@ -98,26 +264,24 @@ const Shop: React.FC<ShopProps> = (props) => {
       document.body.style.cursor = "default";
       setIsOpen(false);
       setItem(undefined);
+      setStacksize(1);
     },
   });
 
-  // Discount factors
   const sDiscount = getStrucBoost("itemDiscountPerLvl", userData.village?.structures);
```
MathiasGruber, line 298: …is shadowed later by:


```typescript
                      {allItems.map((row) => {
                        const factor = shopItemDiscountFactor(
                          row,
                          sDiscount,
                          aDiscount,
                        );
                        const ryoDue = row.cost > 0 ? Math.ceil(row.cost * factor) : 0;
```

Rename the inner one to rowFactor and consider reusing shopItemDiscountFactor(item, sDiscount, aDiscount) for the outer one too so the logic lives in one place.

app/src/layout/Shop.tsx
```diff
@@ -36,23 +59,169 @@ interface ShopProps {
...
+  catalogSearchId?: string;
+  /** Catalog list wrapper id; use `false` to omit (e.g. black market). Default village: tutorial-itemshop. */
+  catalogTutorialId?: string | false;
+  /** Id for the filter popover trigger when multiple shops share a page. */
+  filterPopoverTriggerId?: string;
+  /** User-facing label for Seichi silver (`silver` on black market, default village wording otherwise). */
+  seichiSilverUi?: "seichi" | "silver";
+}
+
+function categoryTabIcon(type: ItemType) {
```
MathiasGruber, line 124: Consider replacing with something like:

```typescript
const TYPE_ICON: Partial<Record<ItemType, LucideIcon>> = {
  WEAPON: Sword,
  CONSUMABLE: FlaskConical,
  ARMOR: Shield,
  ACCESSORY: Sparkles,
  MATERIAL: Package,
  KEYSTONE: Gem,
  CRYSTAL: Diamond,
};
const Icon = TYPE_ICON[type] ?? LayoutGrid;
```

See next comment for tabHint as well, which is the same idea

app/src/layout/Shop.tsx
```diff
@@ -143,32 +307,165 @@ const Shop: React.FC<ShopProps> = (props) => {
...
+  const catalogHeroDescription =
+    props.catalogHeroDescription ??
+    (props.eventItems
+      ? "Limited-run goods for ryo. Use filters to narrow rarity and effects, then open a card to review and purchase."
+      : "Open categories to browse village stock. Filters stack with search — your discounts apply at checkout.");
+  const catalogHeroBadgeCatalog =
+    props.catalogHeroBadgeCatalog ??
+    (props.eventItems ? "Event catalog" : "Village pricing");
+
+  const tabHint =
```
MathiasGruber: Hard to read. Like the TYPE_ICON, change to replace with a Record<ItemType, string> map and a single lookup with a fallback.

app/src/layout/Shop.tsx
```diff
@@ -36,23 +59,169 @@ interface ShopProps {
...
 }
 
 const Shop: React.FC<ShopProps> = (props) => {
-  // Destructure
   const { userData, defaultType, minCost, minRepsCost, minSeichiSilverCost } = props;
+  const useModern = !!props.useItemTypeTabs;
+  const seichiUi = props.seichiSilverUi ?? "seichi";
+  const seichiCatalogWord = seichiUi === "silver" ? "silver" : "seichi";
+  const seichiPurchasePhrase = seichiUi === "silver" ? "silver" : "seichi silver";
+  const seichiWalletLabel = seichiUi === "silver" ? "Silver:" : "Seichi Silver:";
```
MathiasGruber: Collapse to single easier to read lookup:

```typescript
const SILVER_COPY = {
  seichi: { catalog: "seichi", purchase: "seichi silver", wallet: "Seichi Silver:" },
  silver: { catalog: "silver", purchase: "silver",         wallet: "Silver:" },
} as const;
const silverCopy = SILVER_COPY[props.seichiSilverUi ?? "seichi"];
```

app/src/layout/Shop.tsx
```diff
@@ -36,23 +59,169 @@ interface ShopProps {
...
+  catalogHeroDescription?: string;
+  catalogHeroBadgeCatalog?: string;
+  /** When useItemTypeTabs: id for search field + label (avoid duplicates if multiple shops on one page). */
+  catalogSearchId?: string;
+  /** Catalog list wrapper id; use `false` to omit (e.g. black market). Default village: tutorial-itemshop. */
+  catalogTutorialId?: string | false;
+  /** Id for the filter popover trigger when multiple shops share a page. */
+  filterPopoverTriggerId?: string;
+  /** User-facing label for Seichi silver (`silver` on black market, default village wording otherwise). */
+  seichiSilverUi?: "seichi" | "silver";
```
MathiasGruber: All these new ones are only used together, so maybe combine, i.e. something like


```typescript
catalog?: {
  heroTitle?: string;
  heroDescription?: string;
  heroBadge?: string;
  searchId?: string;
  listId?: string;
  filterTriggerId?: string;
  silverLabel?: "silver" | "seichi";
};
```

app/src/layout/Shop.tsx
MathiasGruber, line 264: Heavy query, let's add staleTime: Infinity here

app/src/layout/Shop.tsx
```diff
@@ -183,122 +480,205 @@ const Shop: React.FC<ShopProps> = (props) => {
...
+                        onClick={() => {
+                          filteringState.setItemType(t);
+                          setItem(undefined);
+                          setStacksize(1);
+                          setIsOpen(false);
+                        }}
+                      >
+                        {categoryTabIcon(t)}
+                        <span className="text-center leading-tight">
+                          {capitalizeFirstLetter(t.toLowerCase())}
```
MathiasGruber: Maybe lookup icon here as well, if that looks ok?

app/src/layout/Shop.tsx
```diff
@@ -36,23 +59,169 @@ interface ShopProps {
...
   const [isOpen, setIsOpen] = useState<boolean>(false);
   const [item, setItem] = useState<Item | undefined>(undefined);
   const [stacksize, setStacksize] = useState<number>(1);
   const filteringState = useShopFiltering(defaultType);
   const isAwake = useAwake(userData);
+  const itemTypeTabOptions = useMemo(
+    () =>
+      (props.restrictTypes?.length
+        ? props.restrictTypes
+        : [...ItemTypes]) as ItemType[],
```
MathiasGruber: No need to spread?

MathiasGruber (changes requested):

app/src/layout/Shop.tsx
Phrosfire: I left a note in the PR details. Buying the shuriken still works for the tutorial. It just doesn't point at the when you go to the item shop. I dont recall if it did before or not though.

MathiasGruber (approved):

PR #1350, opened 2026-05-15
Apply new damage formula

Phrosfire:
# Pull Request

## License

## Summary by CodeRabbit

* **New Features**
  * Accessories and keystones now count as pre-battle gear and affect damage calculations.
  * Out-of-combat damage tuning added (configurable base increase and reduction).
  * Combat now records detailed per-hit damage breakdowns for visibility/analysis.

* **Bug Fixes**
  * Damage reduction cap raised from 80% to 90% — minimum damage now 10% of base.

* **Tests**
  * Expanded tests covering the consolidated damage pipeline and pre-battle gear modifiers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/studie-tech/TheNinjaRPG/pull/1350?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/libs/combat/process.ts
+++ b/app/src/libs/combat/process.ts
@@ -4,7 +4,11 @@ import {
   DMG_REDUCTION_CAP,
   DURABILITY_USABILITY_THR,
   ID_ANIMATION_SMOKE,
+  isPreBattleGearFromType,
+  isPreBattleKeystoneFromType,
   NO_DURABILITY_LOSS_COMBATS,
+  OUT_OF_COMBAT_BASE_DAMAGE_INCREASE,
+  OUT_OF_COMBAT_BASE_DAMAGE_REDUCTION,
   POST_DAMAGE_MODIFIER_TYPES,
 } from "@/drizzle/constants";
 import type { ShieldTagType } from "@/validators/combat";
@@ -18,6 +22,8 @@ import {
 } from "./constants";
 import {
   absorb,
+  adjustDamageGiven,
+  adjustDamageTaken,
   afterburn,
   buffPrevent,
   calcDmgModifier,
@@ -42,6 +48,9 @@ import {
   finalStand,
   flee,
   fleePrevent,
+  getDamageModifierPreventState,
+  getEfficiencyRatio,
+  getPower,
   heal,
   healPrevent,
   immunity,
@@ -89,7 +98,9 @@ import type {
   CombatAction,
   CompleteBattle,
   Consequence,
+  ExtraState,
   GroundEffect,
+  PreBattleGearModifiers,
   UserEffect,
 } from "./types";
 import {
@@ -352,39 +363,6 @@ export const applyEffects = (
     (e) => e.type === "increaseheal" || e.type === "decreaseheal",
   );
 
-  // Separate damage boosts (increases) from reductions (decreases)
-  // We'll apply boosts first, then capture a snapshot, then apply reductions
-  const stage1DamageBoosts = usersEffects
-    .filter((e) => damageBoostTypes.includes(e.type))
-    .filter((e) => getEffectStage(e) === 1);
-
-  // Bloodline damage boosts/decreases run last (Phase 4, after non-bloodline reductions + cap); multiplicative; exclude bloodline boosts from Stage 2
-  const bloodlineDamageBoosts = usersEffects.filter(
-    (e) =>
-      "fromType" in e &&
-      e.fromType === "bloodline" &&
-      (e.type === "increasedamagetaken" || e.type === "increasedamagegiven"),
-  );
-
-  const bloodlineDamageReductions = usersEffects.filter(
-    (e) =>
-      "fromType" in e &&
-      e.fromType === "bloodline" &&
-      (e.type === "decreasedamagetaken" || e.type === "decreasedamagegiven"),
-  );
-
-  const stage2DamageBoosts = usersEffects
-    .filter((e) => damageBoostTypes.includes(e.type))
-    .filter((e) => getEffectStage(e) === 2)
-    .filter(
-      (e) =>
-        !(
-          "fromType" in e &&
-          e.fromType === "bloodline" &&
-          (e.type === "increasedamagetaken" || e.type === "increasedamagegiven")
-        ),
-    );
-
   // Apply non-damage-modifier effects first (maintains existing ordering)
   nonDamageModifierEffects.sort(sortEffects).forEach((effect) => {
     applySingleEffect(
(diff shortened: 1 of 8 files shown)


app/src/libs/combat/process.ts
```diff
@@ -401,125 +376,76 @@ export const applyEffects = (
...
-  // Enforce damage reduction cap: damage cannot be reduced below (1 - DMG_REDUCTION_CAP) of boosted pre-reduction damage
-  consequences.forEach((consequence) => {
-    if (
-      consequence.damage !== undefined &&
-      consequence.baseDamageAfterBoosts !== undefined
-    ) {
-      const minDamage = consequence.baseDamageAfterBoosts * (1 - DMG_REDUCTION_CAP);
-      consequence.damage = Math.max(consequence.damage, minDamage);
-    }
+  // Damage modifier tags are excluded from applySingleEffect; carry them forward so they
```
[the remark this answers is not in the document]
Phrosfire: buff/debuff prevent should not kill active effects.  The prevent only stops new incoming buffs/debuffs while the prevent is active.

app/src/libs/combat/process.ts
```diff
@@ -1237,3 +1167,525 @@ export const applySingleEffect = (
...
+    perTagContributions.push({
+      effectId: effect.id,
+      type: effect.type,
+      fromType: "fromType" in effect ? effect.fromType : undefined,
+      delta: increase,
+    });
+  }
+
+  if (totalStaticIncrease !== 0) {
+    damage += totalStaticIncrease;
```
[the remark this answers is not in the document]
Phrosfire: We will not be using static bl boosts

Phrosfire:
@MathiasGruber The formula has been tested by multiple users and appears to be working properly.  While we would like for this to be approved, please do not merge it yet.  We want to make sure everything is lined up and ready to go for when this is merged.

app/src/libs/combat/process.ts
```diff
@@ -401,125 +377,79 @@ export const applyEffects = (
...
 
-  // Capture baseDamageAfterStage1 for all damage consequences, including DOT
-  // This becomes the base for Stage 2 percentage calculations
-  consequences.forEach((consequence) => {
-    const stagedDamage = consequence.damage ?? consequence.residual;
-    if (stagedDamage !== undefined) {
-      consequence.baseDamageAfterStage1 = stagedDamage;
-    }
+  // Consolidated damage pipeline: multiplicative inc buckets, shared sequential % DR, static DR, BL inc
+  applyDamageModifierPipelineToConsequences({
```
MathiasGruber, line 391: iterates over every damage consequence and, for each one, calls computeDamagePacket which runs 7 separate .filter().sort() passes over the full
 usersEffects array:

 - stage1PreBattleIncreases
 - inBattleIncreases
 - stage1PreBattleDrEffects
 - inCombatDrEffects
 - staticIncEffects
 - staticDrEffects
 - bloodlineIncreases
 - bloodlineDrEffects

 Plus a usersEffects.find() per consequence in the wrapper.

 In a big battle with, say, 20 damage consequences and 100+ effects, that's ~160 filter+sort passes per round. The old code did the same kind of work but iterated effects once and applied to all
 consequences — the new code inverts that (iterates consequences, scans all effects each time).

 Mitigation: For typical battles this is probably fine (effects list is usually 20-50, consequences 1-5 per action). But for AOE jutsus hitting many targets with many active buffs/debuffs, this
 could become noticeably slower. Consider pre-computing the filtered/sorted arrays once before the consequence loop and passing them in, rather than re-filtering inside computeDamagePacket for every   damage packet.

app/src/libs/combat/process.ts
```diff
@@ -1237,3 +1167,525 @@ export const applySingleEffect = (
...
+  for (const [effectId, consequence] of consequences) {
+    const damageKey =
+      consequence.damage !== undefined
+        ? "damage"
+        : consequence.residual !== undefined
+          ? "residual"
+          : undefined;
+    if (!damageKey) continue;
+
+    const dmgEffect = usersEffects.find((e) => e.id === effectId);
```
MathiasGruber: the find call is O(n) per consequence. A pre-built Map<string, UserEffect> would make this O(1)

app/src/libs/combat/process.ts
```diff
@@ -401,125 +377,79 @@ export const applyEffects = (
...
-      newGroundEffects,
-      actionEffects,
-      appliedEffects,
-      battle,
-      actorId,
-      effect,
-      action,
-    );
-  });
+  const sealEffects = getActiveSealEffects(usersEffects);
```
MathiasGruber, line 382: getActiveSealEffects(usersEffects) is called once in applyEffects AND again inside every computeDamagePacket call. Since the seal list doesn't change during the pipeline, it should be computed once and passed in.

app/src/libs/combat/process.ts
```diff
@@ -1237,3 +1167,525 @@ export const applySingleEffect = (
...
+      defenderId: consequence.targetId,
+      preBattleGearModifiers,
+      battleRound,
+    });
+
+    consequence[damageKey] = damage;
+    consequence.baseDamageForModifiers ??= rawDamage;
+    consequence.baseDamageAfterStage1 = breakdown.afterPreBattleInc;
+    consequence.baseDamageAfterBoosts = breakdown.baseDamageAfterBoosts;
+    consequence.damageBreakdown = breakdown;
```
MathiasGruber: Every damage consequence now stores a full DamageBreakdown object with arrays (sequentialDrSteps, perTagContributions). If the battle state is serialized to the database each round, this adds non-trivial JSON payload size. Consider whether this should only be computed in debug/analysis mode, or stripped before persistence, cleaned when no longer relevant, or other?

app/src/libs/combat/process.ts
```diff
@@ -1237,3 +1167,525 @@ export const applySingleEffect = (
...
+  effect: UserEffect,
+  consequence: Consequence,
+  kind: "increase" | "decrease",
+  battleRound: number,
+): boolean => {
+  // Same condition as adjustDamageGiven / adjustDamageTaken after applySingleEffect
+  // assigns `effect.castThisRound = (startRound === curRound)`.
+  const { startRound, curRound } = calcEffectRoundInfo(effect, {
+    round: battleRound,
+  } as ReturnedBattle);
```
MathiasGruber: This casts a partial object as ReturnedBattle. If calcEffectRoundInfo ever accesses other battle fields, this will silently produce wrong results rather than failing at compile time. A safer pattern would be to extract what calcEffectRoundInfo actually needs into a narrower type. Goes for all other places this is called as well (I see at least 3 more)

MathiasGruber (changes requested):

theeneon:
@Phrosfire @MathiasGruber Retested back tags and calculations all are working, Setting as complete for review.

MathiasGruber (approved):
Looks good, thanks for addressing all the concerns :)

app/src/libs/combat/process.ts
```diff
@@ -1237,3 +1166,457 @@ export const applySingleEffect = (
...
+): DamagePacketModifierLists => {
+  const packetConsequence = {
+    userId: attackerId,
+    targetId: defenderId,
+  } as Consequence;
+  const applies = (effect: UserEffect, kind: "increase" | "decrease") =>
+    modifierAppliesThisDamagePacket(effect, packetConsequence, kind, battleRound);
+
+  const filterSort = (predicate: (effect: UserEffect) => boolean) =>
+    usersEffects.filter(predicate).sort(sortEffects);
```
MathiasGruber: Performance: `buildDamagePacketModifierLists` makes 8 independent `usersEffects.filter().sort()` passes per attacker-defender pair. Even though results are memoized per pair, each AOE / multi-target action with several distinct pairs pays `8 x O(N) filter + 8 sorts` per pair, per round. Consider a single pass that categorizes each effect into the appropriate bucket (switch on `e.type` + `getEffectStage` + `isBloodlineDamageMod` + `calculation`) and then `.sort()` each bucket once. This is the hot path called every `performAction`.

app/src/libs/combat/process.ts
```diff
@@ -1237,3 +1166,457 @@ export const applySingleEffect = (
...
+
+const modifierAppliesThisDamagePacket = (
+  effect: UserEffect,
+  consequence: Consequence,
+  kind: "increase" | "decrease",
+  battleRound: number,
+): boolean => {
+  // Same condition as adjustDamageGiven / adjustDamageTaken after applySingleEffect
+  // assigns `effect.castThisRound = (startRound === curRound)`.
+  const { startRound, curRound } = calcEffectRoundInfo(effect, { round: battleRound });
```
MathiasGruber: Performance: `modifierAppliesThisDamagePacket` calls `calcEffectRoundInfo(effect, ...)` (and computes `castThisRound` / checks `effect.isNew` / `effect.type`) for every (effect, pair) combination during `buildDamagePacketModifierLists`. These checks are pair-independent - only the `targetId === userId/targetId` test varies per pair. Precompute a `Map<effectId, { eligible: bool, type }>` once over `usersEffects`, then the per-pair filters reduce to a single id lookup + side check. This compounds with the 8 filter passes above.

app/src/libs/combat/process.ts
```diff
@@ -401,125 +375,80 @@ export const applyEffects = (
...
-      action,
-    );
-  });
+  const modifierInfoConsequences = new Map<string, Consequence>();
+  usersEffects
+    .filter((e) => damageModifierTypes.includes(e.type))
+    .filter((e) => !sealCheck(e, sealEffects))
+    .sort(sortEffects)
+    .forEach((effect) => {
+      const targetUser = newUsersState.find((u) => u.userId === effect.targetId);
```
MathiasGruber: Minor perf: second damage-modifier pass does `newUsersState.find((u) => u.userId === effect.targetId)` per matching effect (O(users) lookup). Since this loop runs every `applyEffects` call, build a `Map<userId, BattleUserState>` from `newUsersState` once before the loop and use `.get()`.

app/src/libs/combat/process.ts
```diff
@@ -401,125 +375,80 @@ export const applyEffects = (
...
 
-  // Capture baseDamageAfterStage1 for all damage consequences, including DOT
-  // This becomes the base for Stage 2 percentage calculations
-  consequences.forEach((consequence) => {
-    const stagedDamage = consequence.damage ?? consequence.residual;
-    if (stagedDamage !== undefined) {
-      consequence.baseDamageAfterStage1 = stagedDamage;
-    }
+  // Consolidated damage pipeline: multiplicative inc buckets, shared sequential % DR, static DR, BL inc
+  applyDamageModifierPipelineToConsequences({
```
MathiasGruber, line 391: Buff/debuff prevent check is bypassed by the consolidated pipeline. `applyDamageModifierPipelineToConsequences` mutates the real `consequences` here using all stage1/2/static/bloodline `increasedamagegiven`/`decreasedamagegiven`/`increasedamagetaken`/`decreasedamagetaken` effects from `usersEffects`, with no `preventCheck` filtering (only `sealCheck`). The `increaseDamageGiven` / `decreaseDamageGiven` / `increaseDamageTaken` / `decreaseDamageTaken` calls below (lines 400-429) DO run `preventCheck`, but they target the dummy `modifierInfoConsequences` map — so a `buffprevent` on the attacker or `debuffprevent` on the defender now only suppresses the combat-log message; the damage adjustment has already been applied. Either filter the inputs to `buildDamagePacketModifierLists` / the gear consolidation by `preventCheck`, or apply the equivalent guard inside `computeDamagePacket` before each pass.

app/src/libs/combat/process.ts
```diff
@@ -1237,3 +1166,457 @@ export const applySingleEffect = (
...
+  usersEffects: UserEffect[];
+  attackerId: string;
+  defenderId: string;
+  preBattleGearModifiers: Record<string, PreBattleGearModifiers>;
+  battleRound: number;
+  modifierLists?: DamagePacketModifierLists;
+  sealEffects?: UserEffect[];
+};
+
+export const computeDamagePacket = (
```
MathiasGruber, line 1510: `computeDamagePacket` skips the `buffprevent` / `debuffprevent` check that `increaseDamageGiven` / `decreaseDamageGiven` / `increaseDamageTaken` / `decreaseDamageTaken` perform via `preventCheck`. Previously, when a target had an applicable prevent effect whose `createdRound <= effect.createdRound`, the prevent's RNG roll could short-circuit the damage adjustment entirely. With the consolidated pipeline, the modifier is now always applied to damage; the second-pass info loop at lines 389-435 still emits the "cannot be buffed"/"cannot be debuffed" actionEffect text, so users see a prevent message while the damage was still modified. The prevent check needs to be honored in the pipeline (e.g. precompute prevented-effect IDs per (effect, target) and skip them in `buildDamagePacketModifierLists` / `computeDamagePacket`).

PR #1362, opened 2026-06-15
Fix/barrier aoe absorb scaling

ifroshty:
# Pull Request

## What & why

Fixes a scaling bug where **barriers took ~100× the intended damage from AOE attacks and were destroyed in a single hit**, while single-target attacks behaved correctly.

### Root cause

`damageBarrier` (`app/src/libs/combat/tags.ts`) consumes `barrierAbsorb` as a **0–1 fraction**:

```ts
const damage = damageCalc(effect, origin, target, config) * effect.barrierAbsorb;
```

This is confirmed by `damageUser`, which uses `(1 - barrierAbsorb)` — only valid for a value ≤ 1.

- The **single-target** path supplies the fraction correctly: `getBarriersBetween` normalizes the barrier's `absorbPercentage` to `value / 100` before it reaches the effect.
- The **AOE** path instead read the original barrier straight from `groundEffects` via `findBarrier` and assigned the **raw 1–100 integer** (`absorbPercentage`, default `50`) directly to `barrierAbsorb`.

Result: AOE hits computed `damageCalc(...) × 50` instead of `× 0.5` — exactly 100× the intended damage. Against a barrier's default `maxHealth` of `100`, one AOE tick vaporized it. (The per-barrier `barrierAttacks` dedup means a barrier is hit only once per AOE, so the overkill was purely the bad multiplier, not repeated tile hits.)

### The fix

One line in the AOE branch of `app/src/libs/combat/actions.ts`, matching the single-target path's units:

```diff
- barrierEffect.barrierAbsorb = barrierOnTile.absorbPercentage;
+ barrierEffect.barrierAbsorb = barrierOnTile.absorbPercentage / 100;
```

Since exactly one barrier occupies a tile, there is no path-stacking remainder for AOE, so `/100` reproduces `getBarriersBetween`'s `1 × (50/100) = 0.5` correctly.

## What determines barrier durability (for reference)

Three fields on `BarrierTag` (`app/src/validators/combat.ts`):

| Field | Role | Default | Cap |
|---|---|---|---|
| `curHealth` / `maxHealth` | HP pool (in raw-damage units) | 100 | 100,000 |
| `power` | Barrier "level" — scales its defensive stats, reducing per-hit damage | — | — |
| `absorbPercentage` | Fraction of an incoming hit it soaks (and takes as HP loss) | 50 | 100 |

## Changes

- `app/src/libs/combat/actions.ts` — normalize AOE barrier `barrierAbsorb` to a 0–1 fraction.
- `app/tests/libs/combat/barrier_aoe.test.ts` — **new** regression test.

## Testing

- **New regression test** runs the real `insertAction` AOE path against a real battle grid and asserts the barrier-targeted effect's `barrierAbsorb` equals `absorbPercentage / 100`.
  - ✅ Passes with the fix.
  - ✅ **Fails without the fix** (temporarily reverting `/100` produced `expected 50 to be close to 0.5`), proving it pins the bug.
- ✅ Full combat suite green — **69 tests across 5 files** (`bun vitest run tests/libs/combat/`).
- ✅ `make typecheck` clean.
- biome ignores test files (consistent with the existing suite), so the new test is not linted.

## Breaking changes

No API/schema breaking changes.

**Gameplay impact:** barriers are now meaningfully durable against AOE instead of breaking instantly. This restores the intended behavior already in effect for single-target attacks; existing barrier `maxHealth` / `power` / `absorbPercentage` values continue to govern durability and may be retuned if desired.

## Screenshots

N/A — backend combat-logic change.

## License

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed barrier damage absorption for area-of-effect attacks. Barriers now correctly reduce damage from AOE abilities based on their absorption configuration. Previously, barrier damage reduction was not calculated properly, making barriers less effective than intended against AOE attacks.

### DIFF
--- a/app/src/libs/combat/actions.ts
+++ b/app/src/libs/combat/actions.ts
@@ -859,7 +859,8 @@ export const insertAction = (info: {
                     barrierEffect.targetId = barrierOnTile.id;
                     barrierEffect.id = nanoid();
                     if ("absorbPercentage" in barrierOnTile) {
-                      barrierEffect.barrierAbsorb = barrierOnTile.absorbPercentage;
+                      barrierEffect.barrierAbsorb =
+                        barrierOnTile.absorbPercentage / 100;
                     }
                     usersEffects.push(barrierEffect);
                   }
--- a/app/tests/libs/combat/barrier_aoe.test.ts
+++ b/app/tests/libs/combat/barrier_aoe.test.ts
@@ -1,0 +1,155 @@
+import { describe, expect, it, vi } from "vitest";
+
+/**
+ * Regression test for the AOE-vs-barrier damage scaling bug.
+ *
+ * `damageBarrier` consumes `barrierAbsorb` as a 0-1 fraction. The single-target
+ * path supplies that fraction (getBarriersBetween normalizes absorbPercentage to
+ * `value / 100`), but the AOE path used to assign the raw 1-100 `absorbPercentage`
+ * integer directly. That made AOE hits deal ~100x the intended damage to barriers,
+ * instantly destroying them. This test pins the AOE path to the fractional contract.
+ */
+
+/** Avoid executing real db/env when transitive imports touch `@/server/db`. */
+vi.mock("@/server/db", () => ({ drizzleDB: {} }));
+
+/**
+ * insertAction only needs `checkFriendlyFire` (force pass) from process; it never
+ * calls applyEffects. Mocking the module also avoids loading the Three.js graph.
+ */
+vi.mock("@/libs/combat/process", () => ({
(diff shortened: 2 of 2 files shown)


MathiasGruber (approved):

MathiasGruber (approved):
Reviewed: no prior feedback to address, no logic/DRY/performance concerns found in the 1-line fix, and /tnr-review-now found no gameplay regression. The fix correctly normalizes `absorbPercentage` to a 0-1 fraction in the AOE branch — matching the contract already enforced by `getBarriersBetween` for the SINGLE path. Regression test pins the fix.

PR #1364, opened 2026-06-17
Fix loadout switching and inventory bugs, add renaming feature

ifroshty:
# Pull Request

## What was implemented

This PR fixes three reported loadout problems, adds a loadout-renaming feature, and fixes an adjacent duplicate-row issue.

**Bug fixes**

- **"Grab it from your home" when equipping from inventory.** The equip picker was fed the raw item list and filtered only by slot, so home-stored (and in-auction / mid-crafting) items appeared as equip candidates — which the backend then rejected with "Fetch at home first." A single shared `isEquippableUserItem` predicate now gates both the backpack list and the equip picker, so those items no longer appear as candidates.
- **Switching a loadout dropped gear.** `selectItemLoadout` dropped items via three separate defects (admin-hidden items, duplicate slots, and duplicated itemIds) and applied changes by wiping all gear then re-equipping in parallel — which, on PlanetScale (no transactions), could leave a user fully stripped on a partial failure. It is rewritten around a pure, unit-tested `computeLoadoutAssignments` that fixes all three drops and enforces the same equip limits as normal equipping (`maxEquips`, hand-armor, one bloodline item). The result is applied as a **single atomic `UPDATE ... CASE ... ELSE 'NONE'`** statement (mirroring the existing jutsu switch), with the active-loadout pointer updated **after** the equip succeeds. Items belonging to a now-hidden entry are skipped with a clear warning instead of silently disappearing.

**Feature**

- **Rename jutsu and item loadout slots.** Adds a `name` column to both loadout tables (migration `0012`), a `renameLoadoutSchema` validator, `renameLoadout` mutations on both routers, and a pencil inline-edit control in `LoadoutSelector` (wired through both selector wrappers). Unnamed slots fall back to "Loadout N".

**Adjacent fix**

- **Duplicate loadout rows on concurrent reads.** The lazy auto-create loops could insert more than the allowed number of rows when two reads raced (e.g. two tabs / React StrictMode). They now use deterministic per-slot ids + `onDuplicateKeyUpdate`, so concurrent inserts collide on the primary key instead of duplicating. Slot order is also changed to `asc(createdAt)` so positions stay stable.

## Why

These were user-reported: adding items from the normal inventory dead-ended on "grab it from your home," and switching loadouts sometimes failed to re-apply all saved gear. Loadouts were also only distinguishable by an unstable positional number, so renaming was requested alongside the fixes.

## Breaking changes

**None.** All changes are backward-compatible:

- Migration `0012` is **additive only** — `ALTER TABLE ... ADD name varchar(24) DEFAULT '' NOT NULL` on `JutsuLoadout` and `ItemLoadout`. Safe on existing rows, no backfill, no drops.
- `fetchUserItems` gained an optional `{ includeHidden }` argument that defaults to the previous behavior; all existing callers are unchanged.
- `selectItemLoadout`'s signature and return shape are unchanged (verified against the combat caller).

**One-time, non-breaking UX change:** loadout slot ordering switches from newest-first to oldest-first (`asc(createdAt)`), so existing users will see their slots reorder once. Custom names make this a non-issue going forward.

## Out of scope

The jutsu-loadout switch bypassing equip-limit caps (an over-equip risk, not a drop) was intentionally left out of this PR.

## Screenshots

_To add: the new pencil rename control on a loadout slot, and the inline edit input. (The control only renders for accounts with more than one loadout slot.)_

## Testing

- New unit tests: `tests/libs/item.test.ts` (19 cases covering the predicate, equip constraints, and all three loadout-switch drop fixes) and `tests/validators/loadout.test.ts` (4 cases). All pass.
- `make typecheck` clean; all changed source files are biome-clean.
- Manual verification recommended in a running app: home items no longer appear in the equip picker; switching loadouts preserves all saved gear; renaming persists per loadout type.
- Note: two unrelated full-suite failures exist on this branch's base — `barrier_aoe` (pre-existing combat test) and `daily-link-cleaner` (a real-network test that passes in isolation). No files they touch were modified here.

## License

## Summary by CodeRabbit

* **New Features**
  * Added loadout rename functionality for item and jutsu loadouts.
  * Automatic loadout placeholder generation for missing slots.

* **Enhancements**
  * Improved item equipping validation to prevent unavailable items (home storage, auction, crafting-in-progress).
  * Enhanced loadout selection with safer default handling and inventory-aware restrictions.

* **Tests**
  * Added comprehensive test suites for loadout operations, renaming, and item assignment logic.

### DIFF
--- a/app/src/libs/item.ts
+++ b/app/src/libs/item.ts
@@ -354,39 +354,35 @@ export const computeLoadoutAssignments = (
       invalidItems.push(`${item.name} is being imbued`);
       continue;
     }
-    // Resolve the slot (handles legacy values like ITEM_7). When the saved slot
-    // is no longer valid, fall back to a real equip slot for the item's type —
-    // never the NONE sentinel and never for an item with no real slot type.
+    // Resolve the slot (handles legacy values like ITEM_7). Prefer the saved
+    // slot when it is a real, compatible ItemSlots member and still free;
+    // otherwise — or when an earlier same-type assignment already took it — fall
+    // back to any free compatible slot for the item's type. Never the NONE
+    // sentinel and never for an item with no real slot type.
     const validSlots = ItemSlots as readonly string[];
-    let resolvedSlot: ItemSlot | undefined;
-    // Accept the saved slot only when it is a real ItemSlots member AND actually
-    // compatible with this item's slot type. isCompatibleEquipSlot excludes the
-    // NONE sentinel (a saved NONE would otherwise consume a row/slot while
-    // equipping nothing) and rejects a stale/corrupt entry that points at an
-    // incompatible slot (e.g. a HEAD item saved into CHEST). Anything that fails
-    // falls through to the fallback resolver, which recovers a real slot for the
-    // item's type or reports it invalid.
-    if (
-      validSlots.includes(entry.slot) &&
-      isCompatibleEquipSlot(entry.slot, item.slot)
-    ) {
-      resolvedSlot = entry.slot;
-    } else {
-      const slotType = item.slot;
-      resolvedSlot =
+    const slotType = item.slot;
+    const findFreeSlot = (): ItemSlot | undefined =>
+      slotType && slotType !== "NONE"
+        ? ItemSlots.find((s) => isCompatibleEquipSlot(s, slotType) && !usedSlots.has(s))
+        : undefined;
+    // isCompatibleEquipSlot excludes the NONE sentinel (a saved NONE would
+    // otherwise consume a row/slot while equipping nothing) and rejects a
+    // stale/corrupt entry that points at an incompatible slot (e.g. a HEAD item
+    // saved into CHEST). When the saved slot is usable but already taken (two
+    // same-type items carrying the same stale slot), fall back rather than drop
+    // the second item while another compatible slot is still free.
+    const savedSlotUsable =
+      validSlots.includes(entry.slot) && isCompatibleEquipSlot(entry.slot, item.slot);
+    const resolvedSlot =
+      savedSlotUsable && !usedSlots.has(entry.slot) ? entry.slot : findFreeSlot();
+    if (!resolvedSlot) {
+      // A real slot type with every compatible slot already taken is an
+      // exhausted-slot case; no real slot type at all is a corrupt/invalid entry.
+      invalidItems.push(
         slotType && slotType !== "NONE"
-          ? ItemSlots.find(
-              (s) => isCompatibleEquipSlot(s, slotType) && !usedSlots.has(s),
-            )
-          : undefined;
-      if (!resolvedSlot) {
-        invalidItems.push(`${item.name} has invalid slot`);
-        continue;
-      }
-    }
-    // Never reuse a slot already taken by an earlier assignment.
-    if (usedSlots.has(resolvedSlot)) {
-      invalidItems.push(`${item.name} slot already in use`);
+          ? `${item.name} slot already in use`
+          : `${item.name} has invalid slot`,
+      );
       continue;
     }
     // Equip limits shared with toggleEquipItem.
--- a/app/src/libs/item.ts
+++ b/app/src/libs/item.ts
@@ -11,6 +11,8 @@ import {
   FED_MATERIALS_SILVER_SLOTS,
   FED_NORMAL_INVENTORY_SLOTS,
   FED_SILVER_INVENTORY_SLOTS,
+  type ItemSlot,
+  ItemSlots,
   MATERIALS_BASE_SLOTS,
   MEDNIN_HEAL_ITEM_DISCOUNT_PERC,
 } from "@/drizzle/constants";
@@ -18,6 +20,7 @@ import type {
   Item,
   UserData,
   UserItemWithItem,
+  UserItemWithRelations,
   VillageStructure,
 } from "@/drizzle/schema";
 import { getUserFederalStatus } from "@/utils/paypal";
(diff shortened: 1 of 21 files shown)


app/src/server/api/routers/item.ts
```diff
@@ -2140,112 +2187,71 @@ export const selectItemLoadout = async (
...
+      .where(eq(userItem.userId, user.userId));
+  } else {
+    const sqlChunks: SQL[] = [sql`(case`];
+    for (const a of assignments) {
+      // Re-check availability inside the atomic update (mirrors
+      // getEquipBlockReason): if a row was moved home/auction/crafting between
+      // the snapshot and now, its arm no longer matches and it falls to 'NONE'
+      // instead of being equipped while ineligible.
+      sqlChunks.push(
+        sql`when ${userItem.id} = ${a.userItemId} and ${userItem.storedAtHome} = false and ${userItem.isInAuction} = false and (${userItem.craftingFinishedAt} is null or ${userItem.craftingFinishedAt} <= ${now}) then ${a.slot}`,
```
MathiasGruber: **🟠 Two gaps in the atomic loadout apply.**

1. The CASE re-checks `storedAtHome`/`isInAuction`/`craftingFinishedAt` but imbuement is tracked on the separate `userItemImbuement` table, which the CASE never joins. An item that *starts imbuing* between the snapshot and this UPDATE still matches its arm and gets equipped mid-imbue, contradicting `computeLoadoutAssignments`/`isImbuing`. Consider a `NOT EXISTS` arm against `userItemImbuement` (future `craftingFinishedAt`).
2. The in-memory reflection (~2250) and the returned `items` (~2263) use `bySlot` — the *intended* assignments — so when the CASE legitimately drops a row to `NONE` (a home/auction/crafting race the guard correctly catches), the response still reports it as equipped and emits no warning, and combat operates on the wrong in-memory state. Re-read the committed `equipped` state after the UPDATE and reconcile the response/warnings against it.

app/src/libs/item.ts
```diff
@@ -182,3 +185,228 @@ export const calcItemRepairCost = (useritem: UserItemWithItem) => {
...
+              (s) => isCompatibleEquipSlot(s, slotType) && !usedSlots.has(s),
+            )
+          : undefined;
+      if (!resolvedSlot) {
+        invalidItems.push(`${item.name} has invalid slot`);
+        continue;
+      }
+    }
+    // Never reuse a slot already taken by an earlier assignment.
+    if (usedSlots.has(resolvedSlot)) {
```
MathiasGruber: **🟠 Direct-accept path doesn't fall back when the saved slot is already taken.**

When the saved `entry.slot` is valid + compatible it's accepted at ~373 without consulting `usedSlots`; only the *fallback* branch (~376–381) skips used slots. So if two same-type items (e.g. two `HAND` items) both carry a stale `HAND_1`, the first takes `HAND_1` and the second hits this `usedSlots.has(resolvedSlot)` drop ("slot already in use") even though `HAND_2` is free.

Suggest: after the direct-accept slot is found occupied, retry the same-type fallback resolver before rejecting.

app/src/server/utils/loadout.ts
```diff
@@ -0,0 +1,72 @@
...
+ * callback, and treats rowsAffected===0 (row vanished or not owned) as an error.
+ */
+export const applyLoadoutRename = async (
+  decision: RenameDecision,
+  performUpdate: (name: string) => PromiseLike<{ rowsAffected: number }>,
+): Promise<BaseServerResponse> => {
+  if (!decision.ok) return errorResponse(decision.message);
+  if (!decision.changed) return { success: true, message: decision.message };
+  const result = await performUpdate(decision.name);
+  if (result.rowsAffected === 0) return errorResponse("Loadout not found");
```
MathiasGruber: **🟡 `rowsAffected === 0` is treated as "not found", but a no-op UPDATE also returns 0.**

The PlanetScale driver reports *changed* rows. In a narrow race where `decideRename` computed `changed:true` on a stale snapshot but a concurrent request already set the same name, the UPDATE matches the row yet changes nothing → `rowsAffected 0` → spurious "Loadout not found" on a row that exists.

Consider a guarded re-read (or `WHERE name != :name` plus a separate existence check) before reporting not-found.

app/src/libs/jutsu.ts
```diff
@@ -25,6 +33,102 @@ export const getFreeTransfers = (federalStatus: FederalStatus) => {
...
+    if (jutsu.hidden && !canChangeContent(user.role)) {
+      invalidJutsus.push(`${jutsu.name} is hidden`);
+      continue;
+    }
+    if (!canUseJutsu(jutsu, user)) {
+      invalidJutsus.push(`${jutsu.name}: missing requirements`);
+      continue;
+    }
+    const isResidual = jutsu.effects.some(
+      (e) => "residualModifier" in e && e.residualModifier,
```
MathiasGruber: **🧹 The residual/pierce/event/barrier/stun detection here is copy-pasted from `toggleEquip` and the auto-equip path.**

They can drift (a new capped category would need editing in three places). Consider extracting a `getJutsuCapFlags(jutsu)` helper used by all three.

app/src/server/api/routers/combat.ts
```diff
@@ -965,7 +957,13 @@ export const combatRouter = createTRPCRouter({
               jutsuLoadouts,
               userjutsus,
               user,
-              useritems,
+              (jutsuIds) => {
+                const equipIds = jutsuIds.filter((jutsuId) => {
```
MathiasGruber, line 962: **🟠 Jutsu re-validation is skipped entirely on an item-only loadout switch.**

This new callback validates jutsu against the post-switch items, but `selectJutsuLoadout` (and thus this callback) only runs when `jId !== user.jutsuLoadout` (guard at line ~952). If the user keeps their jutsu loadout but switches *only* the item loadout from one equipping bloodline item X to one that doesn't, the callback never runs and a jutsu gated on X stays equipped without its prerequisite.

Consider re-running jutsu validation whenever the item loadout changes too. (Separately, this callback only checks `checkJutsuBloodlineItem` — equip caps / `canUseJutsu` aren't enforced on this path, as noted in the prior bulk-equip thread.)

ifroshty: fixed: updateCombatLoadout now, when itemChanged && !jutsuChanged, surgically unequips only the currently-equipped jutsus that fail checkJutsuBloodlineItem against post-switch items (DB + battle state), without re-applying the loadout or resetting manual toggles.

PR #1371, opened 2026-06-19
Gotta grow fast

Phrosfire:
# Pull Request

- Adds farming occupation (Not a selected occuptation lole gathering/hunting/craft)
- Adds farming collection log
- Adds five new quest trackers (Crops planted, watered, fertilized, and harvested. Also collection log progression tracker)
- Adds a farm specific currency

@MathiasGruber the uploaded images will need to be put into upload things and the constants be updated to use the uploads. 

## License

## Summary by CodeRabbit

* **New Features**
  * Added a complete farming system with a dedicated farm page, interactive plots, planting, watering, fertilizing, harvesting, seed extraction, crop sales, and bulk actions.
  * Added a farm market for seeds, supplies, plot expansions, and extractors.
  * Added farming progression, experience, currency, level requirements, and quest objectives.
  * Added dynamic day/night visuals and indicators across the world and farm.
  * Added farming-related notifications, rewards, and timer reductions from eligible quests and PvP outcomes.

* **Bug Fixes**
  * Improved inventory transaction safety and handling of empty or stale item stacks.

* **Tests**
  * Added coverage for farming gameplay, validation, visuals, quests, and day/night behavior.


app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,1029 @@
...
+      if (!cropStack) return errorResponse("Item not found");
+      if (
+        cropStack.storedAtHome ||
+        cropStack.isInAuction ||
+        (cropStack.craftingFinishedAt && cropStack.craftingFinishedAt >= new Date())
+      ) {
+        return errorResponse("Crop is not available in inventory");
+      }
+      const sellValue = cropStack.item.farmSellValue;
+      if (sellValue <= 0) return errorResponse("This crop cannot be sold here");
```
MathiasGruber, line 964: This only checks `farmSellValue > 0` — the seed/fertilizer exclusion exists solely client-side in `sellableCrops`. Since a seed's `farmSellValue` is its farm-shop purchase price (see `resolveShopEntry`), a direct tRPC call can sell seeds back at full price, and combined with `extractSeeds` this mints coins whenever `farmExtractSeedCount * seedPrice > cropPrice`. The `!isFarmSeed && !isFarmFertilizer` filter needs to be enforced here too.

app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,1029 @@
...
+        return errorResponse("Not enough farm coins");
+      }
+
+      if (input.type === "PLOT") {
+        const maxPurchased = getMaxPurchasablePlots(farmingLevel);
+        if (user.farmPlotsPurchased >= maxPurchased) {
+          return errorResponse("Plot purchase limit reached for your level");
+        }
+        const newTotal = getTotalFarmPlots(user.farmPlotsPurchased + 1);
+        if (newTotal >= FARM_MAX_PLOTS) {
```
MathiasGruber: Off-by-one: with `FARM_STARTING_PLOTS=5` and `maxPurchased=15` at level 50, the 15th purchase gives `newTotal = min(5+15, 20) = 20`, which trips `20 >= 20` — so the last plot can never be bought and farms cap at 19/20. The SQL guard on line 503 (`+ 1 < FARM_MAX_PLOTS`) has the same problem. Both should be `>` / `<=`. Also `buildShopEntries` doesn't apply this check at all, so the shop shows the plot as purchasable and the mutation then always errors.

app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,1029 @@
...
+      );
+
+      return {
+        success: true,
+        message: `Extracted ${seedsToGrant} seeds from ${cropStack.item.name}`,
+      };
+    }),
+});
+
+const guardFarmingMutation = (user: Awaited<ReturnType<typeof fetchUser>>) => {
```
MathiasGruber, line 1298: `guardFarmingMutation` is missing the `isBanned` check every other economy router has (item, bank, blackmarket all return "You are banned"). `protectedProcedure` doesn't block banned users globally, so a banned player keeps a fully working economy through the farm — buy, sell, plant, harvest, extract, XP and all.

app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,1029 @@
...
+        label: shopItem.name,
+        cost: shopItem.farmSellValue,
+        minLevel: shopItem.farmMinLevel,
+        itemId: shopItem.id,
+      };
+    }
+    return {
+      type,
+      label: shopItem.name,
+      cost: shopItem.farmSellValue > 0 ? shopItem.farmSellValue : shopItem.cost,
```
MathiasGruber: Two gaps vs the listing rules in `buildShopEntries`: (1) neither the SEED nor FERTILIZER branch checks `hidden`, so hidden/staged items are purchasable via direct call even though the regular shop's `buyItem` rejects them; (2) this fallback to `shopItem.cost` charges a ryo-denominated price in farm coins for fertilizers with `farmSellValue=0` — and if `cost` is also 0 (legal when repsCost is set), the item is free. The buy path should enforce the same hidden/`farmSellValue > 0` rules as the listing.

app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,1029 @@
...
+          and(
+            eq(userData.userId, ctx.userId),
+            gte(userData.farmCurrency, shopEntry.cost),
+          ),
+        );
+      if (currencyResult.rowsAffected !== 1) {
+        return errorResponse("Not enough farm coins");
+      }
+
+      await awardItemToUser(ctx.drizzle, ctx.userId, input.itemId, 1);
```
MathiasGruber: This grant path (and `extractSeeds` below) skips the inventory caps — `harvestPlot` enforces `calcMaxMaterials` and the regular `buyItem` enforces calcMaxItems/calcMaxMaterials/calcMaxEventItems, but here a player already at the cap can keep buying seeds/fertilizers or extracting (up to `farmExtractSeedCount * 10` seeds per call) without limit.

app/src/layout/FarmCanvas.tsx
```diff
@@ -0,0 +1,252 @@
...
+      const delta = clock.getDelta();
+      animatePlotEntries(plotEntriesRef.current, delta);
+
+      if (dayNightOverlayRef.current) {
+        updateDayNightOverlay(dayNightOverlayRef.current);
+      }
+
+      // Refresh growth stages periodically without full React re-render
+      refreshElapsed += delta;
+      if (refreshElapsed >= REFRESH_INTERVAL) {
```
MathiasGruber, line 270: This 2s refresh doesn't do what the comment says: `syncPlotVisual` only reads `plot.growthStage`/`plot.isReady`, which are server-computed snapshots from `mapPlotToState` — nothing client-side recomputes them from `plantedAt`/`finishAt`. So stages/ready-glow only advance on the 30s refetch, and `handlePlotClick` sees the stale `isReady`, meaning a crop that ripens right after a fetch can't be click-harvested for up to 30s (the panel's Countdown also hits zero without an `onFinish` invalidation and keeps showing "Growing"). Derive stage/readiness client-side from the timestamps via `getPlotGrowthStage`/`isPlotReady` — the helpers already exist.

app/src/libs/dayNight.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export const getWorldCycleBrightness = (now = new Date()) => {
+  const { phase, phaseOffset, phaseDuration } = getWorldCyclePosition(now);
+  const transitionMs = Math.min(
+    WORLD_CYCLE_TRANSITION_SECONDS * 1000,
+    phaseDuration / 4,
+  );
+
+  if (phase === "day") {
+    if (phaseOffset < transitionMs) {
```
MathiasGruber: The dawn ramp is applied twice: the night branch (lines 64-70) already fades night→day over the last 45s of night, and then this day-start branch re-lerps from `WORLD_NIGHT_BRIGHTNESS` again. So brightness reaches ~1.0 at the end of night, snaps to ~0.38 at the boundary, and re-fades — a visible flash every 30 minutes. Dusk is continuous (day-end ramp into flat night), so one of the two dawn ramps has to go. The test only compares mid-day vs mid-night, so it can't catch this.

app/src/libs/threejs/farming.ts
```diff
@@ -0,0 +1,212 @@
...
+export const syncPlotVisual = (
+  entry: PlotMeshEntry,
+  plot: FarmPlotState | undefined,
+  groups: { crops: Group; effects: Group },
+) => {
+  const position = entry.soil.position;
+
+  if (entry.crop) {
+    groups.crops.remove(entry.crop);
+    entry.crop.material.dispose();
```
MathiasGruber: These materials come from `createSpriteMaterial`'s module-level cache and are marked `userData.shared = true` — the whole point of that flag is that `disposeGroupChildren` skips them. Disposing here (and line 128 for the glow) kills a shared material still attached to every other plot using the same crop texture and to any other canvas using the cache, and the next sync fetches the same disposed instance right back. With the 2s refresh loop that's constant GPU dealloc/recompile churn. Keep the sprite and only mutate scale/texture when the stage actually changes — and never dispose cache-owned materials.

app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,1029 @@
...
+              eq(userData.farmPlotsPurchased, user.farmPlotsPurchased),
+              lt(userData.farmPlotsPurchased, maxPurchased),
+              sql`${FARM_STARTING_PLOTS} + ${userData.farmPlotsPurchased} + 1 < ${FARM_MAX_PLOTS}`,
+            ),
+          );
+        if (currencyResult.rowsAffected !== 1) {
+          return errorResponse("Could not purchase plot — please try again");
+        }
+
+        await ctx.drizzle.insert(farmPlot).values({
```
MathiasGruber: This plain insert races `ensureFarmPlots`: getFarmState (polled every 30s on this same page) computes totals from the already-incremented `farmPlotsPurchased` and backfills the missing slot under the same unique `(userId, slotIndex)` key. If a poll lands between the currency UPDATE and this INSERT, this throws ER_DUP_ENTRY and the user gets a 500 for a purchase that was charged and actually succeeded. Same `onDuplicateKeyUpdate` treatment as `ensureFarmPlots` fixes it (or drop this insert and let the backfill create the row).

app/src/layout/PlotActionPanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+          Click a plot on the farm to plant, water, or fertilize.
+        </CardContent>
+      </Card>
+    );
+  }
+
+  const isEmpty = !plot.seedItemId;
+  const isGrowing = !!plot.seedItemId && !plot.isReady;
+  const now = new Date();
+  const canWaterNow = isGrowing && canWaterPlot(plot.lastWateredAt, now);
```
MathiasGruber: This gates the Water button on the raw browser clock. `canWaterPlot` compares 30-minute epoch buckets, so a client clock a few minutes behind keeps the button disabled that long after the server rolls into a new cycle, and a clock ahead enables it early so every click bounces off the server check. We already have the `timeDiff` server-clock sync in UserContext (Countdown takes it everywhere else) — recompute with a timeDiff-adjusted now instead.

app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,1029 @@
...
+  itemId: string,
+  quantity: number,
+) => {
+  if (quantity <= 0) return;
+
+  const [itemInfo, userItems] = await Promise.all([
+    client.query.item.findFirst({ where: eq(item.id, itemId) }),
+    fetchUserItems(client, userId),
+  ]);
+  if (!itemInfo) return;
```
MathiasGruber: If the item row doesn't resolve, this silently returns after the caller has already done its irreversible write — `extractSeeds` consumes the crops and still reports "Extracted N seeds", `buyShopItem` deducts the coins with no item granted. Related: `harvestPlot` requires both `seedItem` and `yieldItem` to still exist and there's no clear-plot endpoint, so if a referenced item ever gets deleted/reworked, planted plots are stuck forever. Worth returning a success flag from here and giving plots an escape hatch.

MathiasGruber (commented):
Did a detailed pass on this. The farming loop itself is nice, but there are several server-side guard gaps the UI only papers over (economy exploitable via direct API calls), an off-by-one that makes the 20-plot max unreachable, and some client staleness around the canvas/day-night code. Details inline.

app/src/validators/combat.ts
```diff
@@ -1486,6 +1486,18 @@ export const ItemValidatorRawSchema = z.object({
   canBeHunted: z.coerce.boolean().prefault(false),
   canBeGathered: z.coerce.boolean().prefault(false),
   canBeTraded: z.coerce.boolean().prefault(false),
+  isFarmSeed: z.coerce.boolean().prefault(false),
+  farmGrowTimeSeconds: z.coerce.number().int().min(0).prefault(0),
```
MathiasGruber, line 1498: `min(0)` with `prefault(0)` means zero grow time is both legal and the default, and `SuperRefineItem` has no farm cross-checks — a seed saved with the untouched default is instantly harvest-ready, i.e. an instant plant→harvest loop for crops/coins/XP. Should be `min(1)`, or a superRefine rule requiring a positive grow time when `isFarmSeed` is set.

app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,1029 @@
...
+  const farmingLevel = levelInfo.level;
+
+  const eligibleItems = userItems.filter(
+    (ui) =>
+      !ui.storedAtHome &&
+      !ui.isInAuction &&
+      (!ui.craftingFinishedAt || ui.craftingFinishedAt < now),
+  );
+
+  const availableSeeds = eligibleItems
```
MathiasGruber: This emits one entry per userItem stack with no aggregation, and the plant picker keys on `seed.itemId` — two stacks of the same seed (which `awardItemToUser`'s overflow loop creates by design once a stack fills) means duplicate React keys and two buttons for the same seed, each showing only its own stack's count. Aggregate by itemId here; the mutation only takes `seedItemId` anyway.

app/src/layout/ItemWithEffects.tsx
```diff
@@ -465,6 +466,48 @@ const ItemWithEffects: React.FC<ItemWithEffectsProps> = (props) => {
...
                 </p>
               )}
+              {"isFarmSeed" in item && item.isFarmSeed && (
+                <p>
+                  <b>Farm Seed</b>: grow {item.farmGrowTimeSeconds}s · min grow lvl{" "}
+                  {item.farmMinLevel}
+                  {getFarmPlantExperience(item) > 0
+                    ? ` · plant ${getFarmPlantExperience(item)} XP`
+                    : ""}
+                  {item.farmYieldItemId ? ` · yields ${item.farmYieldItemId}` : ""}
```
MathiasGruber: This prints the raw `farmYieldItemId` nanoid to players (same for `farmExtractSeedItemId` on line 498), and the "⚠ also in item shop (should be off)" hint below is admin-speak in a player-facing card. These need name resolution, or to be gated behind a content-staff role check.

app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,1029 @@
...
+  return [...staticEntries, ...dynamicEntries];
+};
+
+export const buildFarmState = async (
+  client: DrizzleClient,
+  userId: string,
+): Promise<FarmStateResponse | { success: false; message: string }> => {
+  const user = await fetchUser(client, userId);
+  const totalPlots = getTotalFarmPlots(user.farmPlotsPurchased);
+  await ensureFarmPlots(client, userId, totalPlots);
```
MathiasGruber: `buildFarmState` runs 5-6 serial query layers on a 30s-polled endpoint: fetchUser → ensureFarmPlots (which SELECTs all plots) → a Promise.all that re-SELECTs the same plots → seedItems → yieldItems. The seedItems layer can go entirely via the existing `with: { seedItem: true }` relation (already used in `fetchFarmPlotWithItems`), and the plot rows shouldn't be read twice. Same pattern in `buyShopItem`: `fetchUser` then `resolveShopEntry` sequentially even though `resolveShopEntry` never reads its `user` param. Per the round-trip rule in CLAUDE.md these should be parallelized/collapsed.

Phrosfire:
Moved back to draft. I meant to move back to a draft yesterday. I'll fix up the requested changes

Phrosfire:
@coderabbitai review

MathiasGruber (approved):

MathiasGruber (commented):
Validated against the current PR head. Six server-side issues reproduce against the PR schema on disposable MySQL (6/6 focused tests, 16 assertions); the client clock issue reproduces with a deterministic skew trace. I have left the actionable details inline.

app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,2214 @@
...
+
+      const guard = guardFarmingMutation(user);
+      if (guard) return guard;
+      if (!plot) return errorResponse("Plot not found");
+      if (!plot.seedItemId || !plot.finishAt)
+        return errorResponse("Nothing growing here");
+      if (plot.fertilizerApplied) return errorResponse("Fertilizer already applied");
+      if (isPlotReady(plot.finishAt)) return errorResponse("Crop is ready to harvest");
+
+      const fertilizerStack = userItems.find((ui) => ui.id === input.userItemId);
```
MathiasGruber: [P1] Reject unavailable fertilizer stacks here. I reproduced this against the PR schema: a stack with `storedAtHome = true` is accepted, consumed, and applied to the plot. This lookup needs the same `!storedAtHome`, `!isInAuction`, and completed-crafting availability predicate used by `fertilizeAll`.

app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,2214 @@
...
+      const now = new Date();
+      const newFinishAt = applyWaterReduction(plot.finishAt, now);
+      const waterResult = await ctx.drizzle
+        .update(farmPlot)
+        .set({ finishAt: newFinishAt, lastWateredAt: now, updatedAt: now })
+        .where(
+          and(
+            eq(farmPlot.id, plot.id),
+            eq(farmPlot.userId, ctx.userId),
+            plot.lastWateredAt
```
MathiasGruber, line 259: [P1] Include the expected crop/timer state in this compare-and-swap. I paused this update after its read, changed `finishAt` through the real database, then released it; the request still succeeded and overwrote the newer timer because it compares only `lastWateredAt`. The same stale request can also match after harvest resets `lastWateredAt` to null and mutate an empty plot.

app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,2214 @@
...
+  }
+
+  if (xpGain <= 0 && !advancesObjective) return true;
+  const result = await client
+    .update(userData)
+    .set({
+      ...(xpGain > 0
+        ? { farmingExperience: sql`${userData.farmingExperience} + ${xpGain}` }
+        : {}),
+      ...questDataUpdate,
```
MathiasGruber: [P1] Protect the quest-data write with a CAS/retry or an atomic representation. In a real-MySQL reproduction, concurrent water actions on two different plots both succeeded and farming XP increased by 30, but both derived quest progress from the same empty snapshot; the last JSON write left `plants_watered` at 1 instead of 2.

app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,2214 @@
...
+          itemId: extraction.seedItemId,
+          quantity,
+          equipped: "NONE" as const,
+        });
+        remaining -= quantity;
+        stackIndex += 1;
+      }
+
+      await client
+        .insert(userItem)
```
MathiasGruber: [P1] Settle rewards with the same capacity-aware stack merge used for admission. I reproduced this at the material-inventory cap with an existing 49/50 seed stack: admission correctly requires zero new slots, but settlement inserts a new one-seed row and leaves inventory at cap + 1.

app/src/server/api/routers/farming.ts
```diff
@@ -0,0 +1,2214 @@
...
+    orderBy: (extraction, { asc }) => [asc(extraction.extractorSlot)],
+    with: { cropItem: true, seedItem: true },
+  });
+
+  const completedExtractions = extractions.filter(
+    (extraction) => extraction.finishAt <= now,
+  );
+  await Promise.all(
+    completedExtractions.map(async (extraction) => {
+      if (!extraction.seedItem || extraction.seedQuantity < 1) return;
```
MathiasGruber: [P2] Do not return while retaining this completed extraction row. With the referenced seed item missing, repeated state settlement leaves the row in `FarmExtraction`; the response filters it out, but `extractSeeds` still counts the occupied extractor slot, permanently locking that slot.

app/drizzle/schema.ts
```diff
@@ -5337,3 +5364,129 @@ export const userStreakProgressRelations = relations(userStreakProgress, ({ one
     references: [activityStreakConfig.id],
   }),
 }));
+
+export const farmPlot = mysqlTable(
+  "FarmPlot",
+  {
+    id: varchar("id", { length: 191 }).primaryKey().notNull(),
+    userId: varchar("userId", { length: 191 }).notNull(),
```
MathiasGruber, line 5372: [P1] Integrate all three new user-owned farm tables into `deleteUserInternal` (or add actual FK cascades). A real-database deletion reproduction removed `UserData` but left one orphan each in `FarmPlot`, `FarmCollectionLog`, and `FarmExtraction`; these columns currently have indexes/Drizzle relations, not database foreign keys with cascades.

app/src/layout/FarmCanvas.tsx
```diff
@@ -0,0 +1,315 @@
...
+    const REFRESH_INTERVAL = 2;
+
+    const animate = () => {
+      performanceMonitor.begin();
+      const delta = clock.getDelta();
+      animatePlotEntries(plotEntriesRef.current, delta, reducedMotionRef.current);
+      animateFarmEffects(groups.effects, delta);
+
+      if (dayNightOverlayRef.current) {
+        updateDayNightOverlay(dayNightOverlayRef.current);
```
MathiasGruber: [P2] Drive the visual cycle from the server-synchronized clock. This call defaults to `getWorldCycleBrightness(new Date())`, even though farming already has `timeDiff`; `Sector` does the same, and `DayNightIndicator` omits `timeDiff` when rendering `Countdown`. With a deterministic 10-minute client skew, the hook reported day with 20 seconds remaining while this raw-clock path reported night and the countdown computed -580 seconds.

app/src/server/api/routers/farming.ts
MathiasGruber: Confirmed and fixed in 0a48c1629. `applyFertilizer` looked the stack up by id alone; it now applies the same availability rule `fertilizeAll` already used (`!storedAtHome`, `!isInAuction`, crafting finished). Covered by a database-backed regression test that fails if the predicate is removed.

app/src/server/api/routers/farming.ts
MathiasGruber: Confirmed and fixed in 0a48c1629. The predicate now also pins `seedItemId` and `finishAt`, matching `applyFertilizer` — which had this right and made watering the odd one out. That closes both halves: the stale reduction can no longer overwrite a newer timer, and the null-`lastWateredAt` match after harvest no longer lets it touch an emptied plot. The regression test stages the stale read through a client whose reads answer with the pre-move row, and fails without the added predicates.

app/src/server/api/routers/farming.ts
MathiasGruber: Confirmed and fixed in 0a48c1629. XP was already conflict-free as a SQL increment, but the trackers are one JSON document rewritten wholesale, so both writers produced `1`. `awardFarmingProgress` now compare-and-swaps on the row it derived the trackers from (via the existing `claimUserSnapshot`) and re-reads and recomputes when it loses, up to three passes. The XP-only path keeps the plain increment, since it needs no guard.

app/src/server/api/routers/farming.ts
MathiasGruber: Confirmed and fixed in 0a48c1629. Settlement now goes through `awardItemToUser`, which fills the existing partial stack before opening a new one — the same merge admission assumed when it counted zero new slots.

app/src/server/api/routers/farming.ts
MathiasGruber: Confirmed and fixed in 0a48c1629. The early return left the completed row behind, and `extractSeeds` only asks whether a row exists for the slot, so it stayed locked. The delete now runs whether or not the seed item still resolves.

app/drizzle/schema.ts
MathiasGruber: Confirmed and fixed in 0a48c1629. All three tables are now deleted in `deleteUserInternal`, alongside `userItem` in the game-progress batch. I went with explicit deletes rather than FK cascades to match how every other user-owned table there is handled.

app/src/layout/FarmCanvas.tsx
MathiasGruber: Confirmed and fixed in 0a48c1629. `FarmCanvas` already kept a `timeDiffRef` for plot state and only the overlay call was on the raw clock; it now passes `getWorldCycleBrightness` a corrected `now`. Worth noting your wider point stands: `Sector` and `DayNightIndicator` have the same gap, and I have left those alone as outside this PR.

MathiasGruber: Correcting myself: I was wrong to defer these as outside the PR. `DayNightIndicator` is added by this branch, and so are the `applyCanvasDayNightBrightness` call and the indicator mount in `Sector` — both are this PR's own code. Fixed in 24dc387d1.

`Sector` now keeps a `timeDiff` ref like `FarmCanvas` and feeds the corrected instant in. `DayNightIndicator`'s phase was already right (`useDayNightCycle` corrects internally), but its `Countdown` was handed the raw `nextPhaseAt` — it now passes `timeDiff`, which is what produced the negative countdown you measured. Sweeping the rest turned up one more: `FarmCanvas` painted the overlay once at setup on the raw clock before the animate loop took over. Every client-side day/night call now derives its instant from `timeDiff`.

app/src/server/api/routers/farming.ts
MathiasGruber: Fixed in 0a48c1629. The availability predicate now matches the one the rest of the inventory uses: the stack must be unequipped, not stored at home, not in an auction, and past any crafting timer. `farming.router.test.ts` covers the stored-at-home case against a real database.

app/src/server/api/routers/farming.ts
MathiasGruber: Fixed in 0a48c1629. The update now pins `seedItemId` and `finishAt` alongside the id, so a reduction derived from a stale read cannot resurrect the longer timer, and a plot harvested in between no longer matches. Regression test stages the interleaving with a client whose reads answer with the pre-move row.

app/src/server/api/routers/farming.ts
MathiasGruber: Fixed, in two steps. 0a48c1629 put the tracker write behind a CAS with a bounded re-read on loss. That was not enough on its own: fe703f716 corrects the guard itself. The document came from `fetchFarmingQuestState` but the timestamp came from `fetchUser`, and those are separate queries, so a write landing between them paired a current timestamp with a stale document and the CAS accepted it. The timestamp now comes from the same read as the trackers. Reverting it reproduces exactly what you measured: expected 2, persisted 1.

app/src/server/api/routers/farming.ts
MathiasGruber: Fixed in 49d34b7c4, after two incomplete attempts worth recording.

Chasing this properly showed the cause sat earlier than settlement. The cap is on material rows, not seeds, and `extractSeeds` sized each job against the inventory as it stood, ignoring seeds the other extractors had already been promised. Two jobs could each fit alone and overflow together. Admission now counts the pending grants, and `guardBulkItemAwardInventoryCapacity` folds repeat awards of one item instead of handing each of them the same partial stack’’s free space — it had the same double-count internally.

At settlement: grants are aggregated per item, `awardItemToUser` re-reads a stack whose CAS it lost rather than falling through to an insert (that staleness also arrives from concurrent requests, which aggregation cannot fix), and a payout that still would not fit is held rather than opening the extra row.

Worth noting one thing your repro implies that is not actually a defect: 49 + 2 seeds across two rows is correct when `stackSize` is 50. The violation is the 25-row cap, not the seed count.

app/src/server/api/routers/farming.ts
MathiasGruber: Fixed. The row is cleared even when the seed item has gone missing — with the seed item absent there is nothing to pay out, so the extraction is settled and deleted rather than returned. Leaving it behind would occupy the extractor slot forever, since `extractSeeds` only asks whether a row exists for that slot.

app/drizzle/schema.ts
MathiasGruber: Fixed in 0a48c1629. `deleteUserInternal` now deletes `FarmPlot`, `FarmExtraction` and `FarmCollectionLog` alongside `UserItem`, in the same parallel batch.

app/src/layout/FarmCanvas.tsx
MathiasGruber: Fixed in 24dc387d1. You were right that this was in scope — I had initially read these as pre-existing, but `git log --diff-filter=A` shows `DayNightIndicator` and the `Sector` brightness call are both added by this PR.

Rather than fix only the call you flagged I swept every caller, which turned up a third: `FarmCanvas` painted the overlay once at setup on the raw clock before the animate loop took over. All three now derive their instant from `timeDiff`. `DayNightIndicator`’’s phase was already correct — `useDayNightCycle` corrects internally — but its `Countdown` was handed the raw `nextPhaseAt`, which is what produced the negative countdown.

app/src/server/api/routers/farming.ts
MathiasGruber: You are right on both counts, and my "needs a killed process" claim was wrong — the delete is an ordinary write, so a transient failure on it was enough to replay the payout. Fixed in 0542f5bf2 rather than documented.

The completion signal was the delete. It is now a `settledAt` mark written straight after the seeds land: that write completes the payout, and the delete became cleanup that may fail freely, because a settled row is never claimed again, only removed. Your sequence — grant commits, delete fails, claim goes stale, settlement retries — is now a regression test, and reverting the guard reproduces it exactly (3 seeds became 6).

What is left is the `settledAt` write itself failing after a successful grant. That is one statement on a row this request already holds, and it is irreducible without transactions: any two-write protocol has a window somewhere, and a payout ledger keyed on the extraction id just moves it between "insert before grant" (loses the reward) and "insert after grant" (duplicates it). So I agree with your framing — this should not be called exactly-once. It is at-most-once except for a single-statement window, which is narrower than what it replaced, and I would put it to the team as accepted risk rather than claim it is closed.

PR #1374, opened 2026-06-20
fix: multi-session clerk bug

ifroshty:
# Pull Request

## Summary

Fixes a Clerk multi-session bug where a user signed into two accounts in the same browser would intermittently see their profile/character switch to the **other** account ("I'm on my main, and it sometimes changes my profile to my alt").

## Why / Root Cause

Under Clerk multi-session, the **active session is per-tab in memory**, but the `__session` cookie the server reads is **shared across all browser tabs and reflects whichever tab is currently focused**.

Our tRPC client authenticated with that ambient cookie only (no token), so a **background request from a non-focused tab** (the 5-minute `profile.getUser` refetch, or a Pusher-triggered `getUser.invalidate()`) authenticated server-side as **whichever account last wrote the cookie** (the alt). Because `profile.getUser` was queried with no account in its React Query key and a global `staleTime: Infinity`, that wrong-account response was written into the tab's cache and pinned there. The same ambient-cookie path also affected **mutations**, not just the displayed profile.

Verified against the installed `@clerk/backend` source: `authenticateRequest` is **header-first**. When an `Authorization` Bearer token is present it is used and the cookie/handshake path is skipped. `getToken()` returns the calling tab's own in-memory session token, independent of the shared cookie.

### 1. Per-tab token on every tRPC request (the fix)

- New `app/src/app/_trpc/authHeaders.ts` with `buildClerkRequestHeaders(token, isMultiSession)`: returns `{ Authorization: Bearer <token> }` when the tab has its own token (and never `Bearer null`); otherwise a fail-closed marker (see #3) or `{}`.
- `app/src/app/_trpc/Provider.tsx`: `TrpcClientProvider` reads `useAuth().getToken` and `useClerk()` (kept in refs so the once-created client always uses the latest values) and attaches headers via `httpBatchLink`'s `async headers()`. Because Clerk authenticates header-first, the server now resolves **this tab's** session instead of the shared cookie — for queries **and** mutations.

### 2. Per-account scoping + server guard (defense-in-depth)

- New `app/src/server/api/viewerGuard.ts` with `assertViewerMatchesSession(sessionUserId, viewerId?)`, which throws `FORBIDDEN` when the client-asserted account does not match the server-authenticated session. Identity for the actual work is **always** taken from `ctx.userId`; `viewerId` is only checked, never trusted.
- `profile.getUser` gains an **optional** `{ viewerId }` input and calls the guard. The React Query cache is now scoped per account.
- Client call sites pass the tab's Clerk id via a small `getUserQueryInput(userId)` helper: `UserContext.tsx` (`useQuery` plus two optimistic `setData`) and `PusherHandler.tsx` (`setData`). The ~120 argless `invalidate()` / `cancel()` calls match by key-prefix and are unchanged.
- `Provider.tsx` `handleTrpcError` silently ignores the transient "Viewer/session mismatch" guard error (matched by message **and** `FORBIDDEN` code), since it self-resolves on the next refetch and is never user-facing.

### 3. Fail-closed when the per-tab token cannot be fetched (defense-in-depth)

`getToken()` yields no usable token in two cases: it **returns `null`** when the tab has no active session, and it **throws** (e.g. `ClerkOfflineError`) on an offline / token-refresh blip. Either way, falling back to the shared cookie could authenticate as another tab's account. Now:

- `Provider.tsx` `headers()` routes **both** the null and the thrown path through the same no-token branch of `buildClerkRequestHeaders`. It sends the `x-tnr-auth-required` marker **only when the browser has more than one signed-in session** (`client.signedInSessions`, which excludes ended/expired/replaced sessions); a single-session browser falls back to the cookie (its own account), so ordinary network blips do not cause needless auth failures for the common single-account case.
- New `app/src/server/api/authContext.ts` with `resolveAuthedUserId(sessionUserId, { hasAuthorizationHeader, tabAuthFailed })`: returns `null` (treat as unauthenticated) when the marker is present and no `Authorization` header arrived, instead of authenticating via the shared cookie. Wired into the tRPC context in `trpc.ts` (the marker is detected by header **presence**).

### 4. Tests

- `app/tests/app/_trpc/authHeaders.test.ts`: `buildClerkRequestHeaders` — Bearer format, never `Bearer null`, fail-closed marker when no token + multi-session, and no marker for a single-session browser.
- `app/tests/server/api/viewerGuard.test.ts`: match passes, mismatch throws `FORBIDDEN`, no assertion when no `viewerId`.
- `app/tests/server/api/authContext.test.ts`: normal request passes through, Authorization-header-present passes through, marker-without-header fails closed to `null`, no-session is `null`.

## Files Changed

| File | Change |
| --- | --- |
| `app/src/app/_trpc/authHeaders.ts` | new |
| `app/src/app/_trpc/Provider.tsx` | modified |
| `app/src/server/api/viewerGuard.ts` | new |
| `app/src/server/api/authContext.ts` | new |
| `app/src/server/api/trpc.ts` | modified |
| `app/src/server/api/routers/profile.ts` | modified |
| `app/src/utils/UserContext.tsx` | modified |
| `app/src/layout/PusherHandler.tsx` | modified |
| `app/tests/app/_trpc/authHeaders.test.ts` | new |
| `app/tests/server/api/viewerGuard.test.ts` | new |
| `app/tests/server/api/authContext.test.ts` | new |
| `app/package.json` / `app/bun.lock` | modified (test-only devDependency pin, see Testing Steps) |

## Breaking Changes

**None.**

- `profile.getUser`'s new input is **optional**, so all existing callers (including server-side `createCaller` / MCP usages that pass no input) keep working unchanged. The guard is a no-op when no `viewerId` is supplied.
- No database / schema migration. No public API or runtime-dependency change.
- The only dependency change is a **test-only** devDependency pin (`@vitejs/plugin-react` `^6` → `^5`) described under Testing Steps.

## Conformance

The approach was validated against current official docs (Clerk v7, tRPC v11, TanStack Query v5, React 19): the `getToken()` + `Authorization: Bearer` pattern with a fail-closed marker is exactly what Clerk prescribes for multi-tab background fetches; async `headers()` on the terminating `httpBatchLink`, optional `.input` for cache-keying with `ctx`-derived identity, and per-account query keys are all idiomatic for their respective libraries.

### Automated (from repo root)

```bash
make typecheck   # passes, 0 errors
make lint        # clean
cd app && bunx vitest run tests/app/_trpc/authHeaders.test.ts tests/server/api/viewerGuard.test.ts tests/server/api/authContext.test.ts
```

> Note: this branch also pins `@vitejs/plugin-react` from `^6` to `^5`. `@vitejs/plugin-react@6` requires vite 8, but `vitest@4` uses vite 7, which broke the test runner at config load (`vite/internal` not exported). `^5` peers vite 7, so `vitest` / `make test` run normally again. This is a test-only devDependency change.

### Manual: confirm the fix is wired (single account, any instance)

1. Sign in, open DevTools and go to the **Network** tab.
2. Inspect any `/api/trpc` request and check Request Headers; it now carries `Authorization: Bearer ...`. (Previously there was no Authorization header.)

### Manual: behavioral repro (requires a Clerk instance with multi-session enabled)

1. Tab A: sign in as account **A**, sit on `/profile`.
2. Tab B: sign in as account **B** (same browser, Clerk multi-session).
3. Focus **Tab B**, leave Tab A idle and do not interact with it.
4. Expected (fixed): Tab A stays account A. Before this change it would flip to B on the next background `getUser`.
5. To trigger quickly instead of waiting on the 5-minute timer, temporarily set `refetchInterval` from `300000` to `3000` in `app/src/utils/UserContext.tsx`, reproduce, then revert.

> Note: multi-session is a per-instance Clerk Dashboard setting and differs between development and production instances. In single-session mode the second sign-in is redirected away, so the bug (and this repro) only occur where multi-session is enabled.

## Screenshots

N/A. No visual/UI changes; the fix is in request authentication and query caching.

## Evaluated and intentionally not included

- **SSR**: `auth()` / `currentUser()` in server components reads the shared cookie. Verified this is not an exposure here: `initialIsSignedIn` is an account-agnostic boolean used only for the pre-hydration layout shell, and the only server components calling `currentUser()` are forum pages that render on foreground navigation (focused tab = correct cookie), use `force-dynamic` (so prefetch does not run their data fetch), and surface only public forum content. There is also no clean way to pass a per-tab bearer token to a server component during document render, so no change was made.
- **Uploadthing**: `api/uploadthing/core.ts` is still cookie-authenticated and does not traverse the tRPC link. Uploads are foreground/click-driven (the focused tab's cookie is the correct account), so the cross-account precondition is essentially never met. Pre-existing and out of scope for this PR.

## License

## Summary by CodeRabbit

* **New Features**
  * Added per-tab Clerk session pinning to keep authentication and user data scoped to the pinned viewer/account.
  * Updated profile user lookup to accept an optional viewer id and enforce viewer/session match.
* **Bug Fixes**
  * Improved fail-closed handling for multi-session per-tab auth failures to avoid incorrect session fallback.
  * Reduced user-facing noise by suppressing transient viewer/session mismatch errors.
  * Scoped battle-related profile cache updates to the current pinned viewer/account.
* **Tests**
  * Added coverage for header composition, multi-session detection, fail-closed auth resolution, viewer/session enforcement, and session pinning helpers.
* **Chores**
  * Updated the Vite React plugin version.

### DIFF
--- a/app/src/app/_trpc/SessionPinProvider.tsx
+++ b/app/src/app/_trpc/SessionPinProvider.tsx
@@ -13,6 +13,7 @@ import {
 } from "react";
 import {
   clearPinnedSessionId,
+  decidePinnedSession,
   readPinnedSessionId,
   resolvePinnedSession,
   writePinnedSessionId,
@@ -69,23 +70,21 @@ export function SessionPinProvider(props: { children: React.ReactNode }) {
   const pinnedIdRef = useRef(pinnedSessionId);
   pinnedIdRef.current = pinnedSessionId;
 
-  // Adopt the active session as this tab's pin on first load, and re-adopt only if
-  // the stored pin is no longer a signed-in session (e.g. signed out elsewhere). A
-  // still-valid stored pin is NEVER overridden by a background active-session
-  // change — that is what keeps the tab on its account.
+  // Adopt the active session as this tab's pin on first load. A still-signed-in pin
+  // (in memory, the source of truth) is NEVER overridden by a background
+  // active-session change — that is what keeps the tab on its account. See
+  // decidePinnedSession for the full ordering and the transient-empty guard.
   useEffect(() => {
     if (!isLoaded) return;
-    const stored = readPinnedSessionId();
-    if (resolvePinnedSession(sessions, stored)) {
-      if (pinnedIdRef.current !== stored) setPinnedSessionId(stored);
-      return;
-    }
-    if (activeSessionId) {
-      writePinnedSessionId(activeSessionId);
-      setPinnedSessionId(activeSessionId);
-    } else if (pinnedIdRef.current !== null) {
-      clearPinnedSessionId();
-      setPinnedSessionId(null);
+    const decision = decidePinnedSession({
+      sessions,
+      inMemoryPinnedId: pinnedIdRef.current,
+      storedPinnedId: readPinnedSessionId(),
+      activeSessionId,
+    });
+    if (decision.type === "set") {
+      writePinnedSessionId(decision.id);
+      setPinnedSessionId(decision.id);
     }
   }, [isLoaded, sessions, activeSessionId]);
 
--- a/app/src/app/_trpc/SessionPinProvider.tsx
+++ b/app/src/app/_trpc/SessionPinProvider.tsx
@@ -1,0 +1,156 @@
+"use client";
+
+import { useSession, useSessionList } from "@clerk/nextjs";
+import {
+  createContext,
+  Fragment,
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
+import {
+  clearPinnedSessionId,
+  readPinnedSessionId,
+  resolvePinnedSession,
+  writePinnedSessionId,
+} from "./sessionPin";
+
+interface SessionPinContextValue {
+  /** Clerk session id this tab is pinned to, or null before it resolves. */
+  pinnedSessionId: string | null;
+  /** User id of the pinned session, or null when no pinned session is signed in. */
+  pinnedUserId: string | null;
+  /** Count of signed-in Clerk sessions in the browser (undefined while loading). */
+  signedInSessionCount: number | undefined;
+  /** Mints a token for THIS tab's pinned session (not the browser-active one). */
+  getPinnedToken: () => Promise<string | null>;
+  /** Pins this tab to a different signed-in session (an intentional in-tab switch). */
+  switchPinnedAccount: (sessionId: string) => void;
+  /** Clears the pin (used on sign-out). */
+  clearPin: () => void;
+}
+
+const SessionPinContext = createContext<SessionPinContextValue | null>(null);
+
+/**
+ * Pins each browser tab to a specific Clerk session.
+ *
+ * Clerk's active session is browser-global: with multiple accounts signed in, one
+ * `lastActiveSessionId` is shared across tabs and re-resolved on reload/refocus,
+ * so a tab can silently flip to another signed-in account. This provider records
+ * the tab's session id in `sessionStorage` (per-tab, survives reload, not shared
+ * across tabs) and exposes it so the app's identity (`UserContext`) and tRPC auth
+ * (`TrpcClientProvider`) follow the PINNED session instead of the active one.
+ *
+ * It also replaces Clerk's `MultisessionAppSupport`: that component remounts the
+ * app keyed on the ACTIVE session, which would re-flip the tab on every global
+ * active-session change. Here we key the subtree on the PINNED session, so an
+ * intentional in-tab account switch cleanly remounts (fresh tRPC client + React
+ * Query cache) while a background active-session change does not.
+ */
+export function SessionPinProvider(props: { children: React.ReactNode }) {
+  const { isLoaded, sessions, setActive } = useSessionList();
+  const { session: activeSession } = useSession();
+  const activeSessionId = activeSession?.id ?? null;
+
+  // The tRPC headers() callback is created once and reads the latest values via
+  // refs (it runs per request batch, outside React's render).
+  const sessionsRef = useRef(sessions);
+  sessionsRef.current = sessions;
+  const setActiveRef = useRef(setActive);
+  setActiveRef.current = setActive;
+
+  const [pinnedSessionId, setPinnedSessionId] = useState<string | null>(() =>
+    readPinnedSessionId(),
+  );
+  const pinnedIdRef = useRef(pinnedSessionId);
+  pinnedIdRef.current = pinnedSessionId;
+
+  // Adopt the active session as this tab's pin on first load, and re-adopt only if
+  // the stored pin is no longer a signed-in session (e.g. signed out elsewhere). A
+  // still-valid stored pin is NEVER overridden by a background active-session
+  // change — that is what keeps the tab on its account.
+  useEffect(() => {
+    if (!isLoaded) return;
+    const stored = readPinnedSessionId();
+    if (resolvePinnedSession(sessions, stored)) {
+      if (pinnedIdRef.current !== stored) setPinnedSessionId(stored);
+      return;
+    }
+    if (activeSessionId) {
+      writePinnedSessionId(activeSessionId);
+      setPinnedSessionId(activeSessionId);
+    } else if (pinnedIdRef.current !== null) {
+      clearPinnedSessionId();
+      setPinnedSessionId(null);
+    }
+  }, [isLoaded, sessions, activeSessionId]);
+
+  const pinnedSession = useMemo(
+    () => resolvePinnedSession(sessions, pinnedSessionId),
+    [sessions, pinnedSessionId],
+  );
+
+  const getPinnedToken = useCallback(async () => {
+    const pinned = resolvePinnedSession(sessionsRef.current, pinnedIdRef.current);
+    if (!pinned) return null;
+    return pinned.getToken();
+  }, []);
+
+  const switchPinnedAccount = useCallback((sessionId: string) => {
+    writePinnedSessionId(sessionId);
+    setPinnedSessionId(sessionId);
+    // Align Clerk's active session for UI consistency (e.g. UserButton); the app's
+    // data/auth use the pin directly and do not depend on this succeeding.
+    void setActiveRef.current?.({ session: sessionId });
+  }, []);
+
+  const clearPin = useCallback(() => {
+    clearPinnedSessionId();
+    setPinnedSessionId(null);
+  }, []);
+
+  // Mirror Clerk's `client.signedInSessions` (active + pending) so the fail-closed
+  // multi-session detection never under-counts and falls back to the shared cookie.
+  const signedInSessionCount = useMemo(
+    () =>
+      sessions?.filter((s) => s.status === "active" || s.status === "pending").length,
+    [sessions],
+  );
+
+  const value = useMemo<SessionPinContextValue>(
+    () => ({
+      pinnedSessionId,
+      pinnedUserId: pinnedSession?.user?.id ?? null,
+      signedInSessionCount,
+      getPinnedToken,
+      switchPinnedAccount,
+      clearPin,
+    }),
+    [
+      pinnedSessionId,
+      pinnedSession,
+      signedInSessionCount,
+      getPinnedToken,
+      switchPinnedAccount,
+      clearPin,
+    ],
+  );
+
+  return (
+    <SessionPinContext.Provider value={value}>
+      <Fragment key={pinnedSessionId ?? "no-pin"}>{props.children}</Fragment>
+    </SessionPinContext.Provider>
+  );
+}
+
+export function useSessionPin(): SessionPinContextValue {
+  const ctx = useContext(SessionPinContext);
+  if (!ctx) {
+    throw new Error("useSessionPin must be used within a SessionPinProvider");
+  }
+  return ctx;
+}
--- a/app/src/app/_trpc/SessionPinProvider.tsx
+++ b/app/src/app/_trpc/SessionPinProvider.tsx
@@ -13,7 +13,6 @@ import {
   useState,
 } from "react";
 import {
-  clearPinnedSessionId,
   decidePinnedSession,
   isAuthRoute,
   pinChangeRequiresRemount,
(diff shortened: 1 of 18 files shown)


app/src/layout/PusherHandler.tsx
```diff
@@ -71,8 +71,8 @@ export const usePusherHandler = (
           if (data?.battleId) {
             // NOTE: for some reason using updateUser does not work from this hook
             await utils.profile.getUser.cancel();
-            // Update user data with battle status
-            utils.profile.getUser.setData(undefined, (old) => {
+            // Update user data with battle status (scoped to this tab's account)
+            utils.profile.getUser.setData({ viewerId: userId }, (old) => {
```
MathiasGruber: The cache-key shape `{ viewerId: userId }` is duplicated here. `UserContext.tsx` already has a `getUserQueryInput` helper for this exact purpose. Since both sites must agree on the input shape for the per-account cache scoping to work, consider exporting `getUserQueryInput` (or moving it to a shared module) and reusing it here. If the two ever drift (e.g. someone adds another scoping field to the helper), the cache scoping will silently break and could serve cross-account data.

MathiasGruber: Issue: The cache-key shape `{ viewerId: userId }` is still duplicated here. `UserContext.tsx` (lines 69-71) defines `getUserQueryInput` for this exact purpose but it is not exported, so this call site can drift if the helper ever gains another scoping field — silently breaking the per-account cache scoping and risking cross-account data.

Fix: Export `getUserQueryInput` from `UserContext.tsx` (or move it to a shared module like `@/app/_trpc/` alongside `sessionPin.ts`) and reuse it here, e.g.
```ts
utils.profile.getUser.setData(getUserQueryInput(userId), (old) => { ... });
```
Note `getUserQueryInput` returns `undefined` for falsy `userId`, so guard the call (the surrounding `if (userId)` on line 44 already ensures it's defined here).

MathiasGruber (commented):
Verified the prior round of feedback against the current branch. One item remains unaddressed.

app/tests/app/_trpc/sessionPin.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    sessionStorage: {
+      getItem: (k: string) => (store.has(k) ? (store.get(k) as string) : null),
+      setItem: (k: string, v: string) => void store.set(k, v),
+      removeItem: (k: string) => void store.delete(k),
+    },
+  };
+};
+
+afterEach(() => {
+  vi.unstubAllGlobals();
```
MathiasGruber: Issue: `vi.unstubAllGlobals()` (L22) and `vi.stubGlobal(...)` (L56, L65) are vitest-only APIs. CI runs `bun test` (see `.github/workflows/test_build.yml`), and bun's built-in `vi` shim does not implement `stubGlobal` / `unstubAllGlobals`, so all 13 tests in this file fail with `TypeError: vi.stubGlobal is not a function`. This is what's currently blocking the `test_build` check.

Fix: replace the stubs with direct assignment + restore, e.g.
```ts
let originalWindow: typeof globalThis.window | undefined;
beforeEach(() => { originalWindow = (globalThis as any).window; });
afterEach(() => { (globalThis as any).window = originalWindow; });
// then in the test:
(globalThis as any).window = makeFakeWindow();
```
The other new test files in this PR (authHeaders.test.ts, authContext.test.ts, viewerGuard.test.ts) already run cleanly under bun.

MathiasGruber (commented):
Build / test runtime

app/src/app/_trpc/SessionPinProvider.tsx
```diff
@@ -0,0 +1,155 @@
...
+      sessions,
+      inMemoryPinnedId: pinnedIdRef.current,
+      storedPinnedId: readPinnedSessionId(),
+      activeSessionId,
+    });
+    if (decision.type === "set") {
+      writePinnedSessionId(decision.id);
+      setPinnedSessionId(decision.id);
+    }
+  }, [isLoaded, sessions, activeSessionId]);
```
MathiasGruber: Critical: the multi-session fix does not hold when a user signs in a *second* account inside an already-loaded tab. Repro: tab 1 signed in as user A; open tab 2 fresh and navigate to `/login`. Tab 2's first effect run sees `inMemoryPinnedId=null, storedPinnedId=null, sessions=[sess_A], activeSessionId=sess_A` (Clerk's session list is browser-global) and the fallback at `sessionPin.ts:93-94` pins tab 2 to `sess_A`. After the user signs in user B via `<SignIn>`, Clerk adds `sess_B` and switches active to `sess_B`, but in this effect `inMemoryPinnedId` is `sess_A` and `resolvePinnedSession` at `sessionPin.ts:79` still returns a signed-in session, so the decision is `keep`. Tab 2 keeps rendering user A's profile.

`switchPinnedAccount` would handle this, but `grep -rn switchPinnedAccount app/src` shows zero call sites outside this file — nothing in the Clerk sign-in flow ever invokes it. The latest commit (`107cbd71`) only extracted `getUserQueryInput` and fixed vitest globals; this code path is unchanged since the prior review.

Needs a fix that detects an in-tab account switch — e.g. on observing a *newly* signed-in session that becomes active while the prior pin is also still signed in, re-pin to the new active id; or wire `switchPinnedAccount` into the `<SignIn>` post-sign-in callback; or defer the first-load adoption until something proves the tab is acting as that account. Please add a regression test that exercises `sessions=[A] → [A,B]` with `active A → B` and asserts the pin moves to B.

app/src/server/api/viewerGuard.ts
```diff
@@ -0,0 +1,25 @@
...
+ * `viewerId` — this only rejects mismatches, it never trusts the client id.
+ */
+export function assertViewerMatchesSession(
+  sessionUserId: string,
+  viewerId?: string,
+): void {
+  if (viewerId !== undefined && viewerId !== sessionUserId) {
+    throw new TRPCError({
+      code: "FORBIDDEN",
+      message: "Viewer/session mismatch",
```
MathiasGruber: The literal `"Viewer/session mismatch"` is a contract between this guard and the client's silent-suppression logic in `Provider.tsx` (line 184 does `error.message.includes("Viewer/session mismatch")`, and the comments on lines 155/178/194 repeat it). If anyone tweaks the wording here, the client filter silently breaks and these transient errors start showing destructive toasts to users. Suggest extracting a shared constant (e.g. `VIEWER_SESSION_MISMATCH_MESSAGE`) that both sides import, so the linkage is explicit and grep-able.

MathiasGruber (commented):
Audit of prior feedback against the rebased branch: all 10 previously-raised review threads are addressed. However, the /tnr-review-now finding from 2026-06-28 still reproduces, and there is one DRY contract worth tightening.

app/src/app/_trpc/SessionPinProvider.tsx
```diff
@@ -0,0 +1,163 @@
...
+      signedInSessionCount,
+      getPinnedToken,
+      switchPinnedAccount,
+      clearPin,
+    ],
+  );
+
+  return (
+    <SessionPinContext.Provider value={value}>
+      <Fragment key={pinnedSessionId ?? "no-pin"}>{props.children}</Fragment>
```
MathiasGruber: **Perf**: `<Fragment key={pinnedSessionId ?? "no-pin"}>` here wraps `TrpcClientProvider` + `UserContextProvider` + the entire page tree (see `app/layout.tsx`: `SessionPinProvider` is the outermost of these). On any first load where `readPinnedSessionId()` returns `null` (first-ever visit, sessionStorage cleared, incognito, or freshly redirected from `/login` to an app route), `pinnedSessionId` starts as `null` and the effect at L84 then flips it to the resolved id; the key changes from `"no-pin"` to that id and the entire subtree unmounts and remounts.

That remount throws away the React Query cache, recreates the `httpBatchLink` tRPC client, and causes every initial `useQuery` (including `profile.getUser`) to refire from scratch. For unauthenticated visitors there is no pin to adopt so it does not bite, but every signed-in first load and every post-sign-in redirect pays this cost once. Consider initializing `pinnedSessionId` from the active session synchronously when possible, or keying only on transitions between non-null pins (e.g. start the key at the resolved id and only remount on intentional `switchPinnedAccount` / `clearPin`), so the normal first-load adoption is not a tree-wide remount.

MathiasGruber (commented):
Performance review of the Clerk multi-session changes. One real concern flagged inline; the server-side `resolveAuthedUserId` and `assertViewerMatchesSession` are pure CPU and add no DB roundtrips, and the new `viewerId` input on `profile.getUser` does not change the query plan.

app/src/app/_trpc/SessionPinProvider.tsx
ifroshty: Fixed in latest commit

app/src/app/_trpc/SessionPinProvider.tsx
```diff
@@ -0,0 +1,182 @@
...
+    [sessions, pinnedSessionId],
+  );
+
+  const getPinnedToken = useCallback(async () => {
+    const pinned = resolvePinnedSession(sessionsRef.current, pinnedIdRef.current);
+    if (!pinned) return null;
+    return pinned.getToken();
+  }, []);
+
+  const switchPinnedAccount = useCallback((sessionId: string) => {
```
MathiasGruber: `switchPinnedAccount` is exposed on the context but has no consumers (grep confirms only self-references and the JSDoc in `sessionPin.ts`). Clerk's `UserButton` account-switch calls `setActive` directly, bypassing this helper, so in-tab account switching from the standard UI does not run through here. Either wire a consumer (or a listener on Clerk's `setActive` that re-pins the tab) so account switching actually retargets the pinned session, or drop the unused export.

app/src/app/_trpc/SessionPinProvider.tsx
```diff
@@ -0,0 +1,182 @@
...
+    writePinnedSessionId(sessionId);
+    setPinnedSessionId(sessionId);
+    // Align Clerk's active session for UI consistency (e.g. UserButton); the app's
+    // data/auth use the pin directly and do not depend on this succeeding.
+    void setActiveRef.current?.({ session: sessionId });
+  }, []);
+
+  const clearPin = useCallback(() => {
```
MathiasGruber: `clearPin` is exposed on the context but never invoked anywhere in the app. The JSDoc in `sessionPin.ts` describes it as "used on sign-out", but sign-out goes through Clerk's own UI/`signOut` and never calls this. It happens to work because `resolvePinnedSession` returns null for signed-out ids, but the `sessionStorage` entry lingers until a new pin is adopted or the tab closes. Either call `clearPin` from a sign-out handler, or delete it (and `clearPinnedSessionId`) as unused.

MathiasGruber:
Completed label removed due to failing checks

MathiasGruber (approved):
Approved.

PR #1375, opened 2026-06-21
Feature/player piloted summons

ifroshty:
# Pull Request

## Summary

Adds **player-piloted summons**: when a human player summons a creature in an interactive battle, they now **drive that summon themselves** on its turn (its own jutsu, items, action points and targeting) instead of handing it to the rule-based AI. Along the way the summon system was rewritten to fix long-standing issues (re-cast, one-per-controller, masterless summons), two regressions found during review were fixed, and the summon lifecycle logic was **consolidated into a single tested module** so it stops breaking on each change. Summons are now also correctly **disabled in auto-resolved battle types** (`KAGE_AI` / `CLAN_CHALLENGE`).

## What was implemented

**Player-piloted summons**
- New transient `isPiloted` flag on the battle user state. A summon is piloted when its controller is a live human in a non-auto battle type (`shouldPilotSummon`). `isAi` deliberately stays `true`, so reward attribution, drop tables and the AI accounting path are unchanged — `isPiloted` only affects turn routing and UI.
- Server turn routing via `getTurnControl` / `wantsHumanActionSet`: on a piloted summon's turn the action is taken by the **human controller**, not the AI. The AI dispatch loop never auto-drives a piloted summon, and a defense-in-depth guard prevents the AI exhaustion path from self-killing a piloted summon.
- Per-user masking exposes `isPiloted` to all viewers (added to the public allow-list) while the summon's private stats remain visible only to its controller (unchanged `controllerId`-based masking).
- Client: a shared `resolveControlledActorId` helper makes the combat UI follow the controlled actor — normally the player, transparently the summon on its turn — so the action menu, timer, AP, targeting, camera and keyboard end-turn all act for the summon without any extra UI. AI auto-play is suppressed for the player's own piloted summon.

**Summon system rewrite / hardening**
- **One-summon-per-controller cap** (`hasLiveSummon`) with a clear "already has a summon!" message; clears once the prior summon is dead/fled/left so re-summon works.
- **Orphan cleanup** (`isOrphanedSummon` / `spliceOrphanedSummons`): removes a *live* masterless summon when its controller dies/flees/leaves, before actor selection, so the turn order never lands on or names a controllerless summon.
- AI self-kill guard for piloted summons.

**Bug fixes (found in review)**
- **Summoning was completely broken on the branch** — the summon-template lookup had been changed to match on `userId`, but battle-loaded AI templates carry their DB id in `controllerId` (their `userId` is a fresh nanoid). Restored the `controllerId` match so `summon()` finds the template again. Fixes *"Failed to create summon!"* for all summons.
- **Re-summon failed after the first** — orphan cleanup was splicing out the hidden clone-source template (parked at `curHealth: 0`), so subsequent casts had no template. `isOrphanedSummon` now only treats *live* summons as orphans, leaving templates (and dead summons) in place. Both first-cast and re-cast now work.

**Hardening — single source of truth**
- Consolidated the scattered summon helpers into a new `app/src/libs/combat/summon.ts` and replaced the hand-re-derived "template vs active summon" checks with one tested predicate set (`isSummonTemplate` / `isActiveSummon` / `isLiveSummon`), routed through the cap, orphan cleanup and AI gate. The summon-template **lookup and teardown stay raw** (`controllerId`-based) by design — routing them through the predicate would regress legacy lookups and dead-summon removal.
- Added an explicit, **server-internal** `isSummonTemplate` marker (set at load, cleared on spawn/clone), and enforced the template `curHealth:0` invariant in **all** battle types by guarding the AutoBattle pool reset with `!user.isSummon` — so a hidden off-grid template can never be selected as the active actor.
- `spliceOrphanedSummons` now classifies against a pre-mutation snapshot (order-independent).
- **No summons in `KAGE_AI` / `CLAN_CHALLENGE`** — `summon()` now refuses to spawn in auto-resolved battle types, where there are no human turns. Previously an AI participant there could spawn an AI-driven summon.
- Comprehensive tests pin every category (template / spawned / clone), the legacy fallback, masking-absence, splice determinism, turn-order exclusion and the auto-battle block, so future edits can't silently reintroduce the old bugs.

## Why

The summon feature previously ran entirely under AI, giving players no control over creatures they summoned. Piloting makes summons a real tactical tool. The rewrite also resolves re-cast and masterless-summon edge cases that were awkward in the old implementation.

## Screenshots

_TODO: add a combat screenshot showing the action panel switching to the summon's jutsu/AP on the summon's turn._

## Breaking changes

**No schema/API breaks; one intentional behavior change.**
- No database schema or migration changes. `isPiloted` (public) and `isSummonTemplate` (server-internal) are transient fields inside the battle-state JSON, not new columns.
- **Behavior change:** summons no longer occur in auto-resolved battle types (`KAGE_AI`, `CLAN_CHALLENGE`). Previously an AI participant there could spawn an AI-driven summon; `summon()` now refuses to spawn in those types, matching the design that those battles have no human turns. (Interactive battle types are unaffected.)

## Testing

- `bun test tests/libs/combat/` → **140 passed / 0 failed** (14 files). Coverage spans the piloting suites (turn control, action-set selection, controlled-actor resolution, AI piloted-guard, masking, one-per-controller cap, orphan cleanup), the `summon()` template-lookup regression tests (first-cast, re-cast-through-cleanup, template-kept, live-orphan-removed, turn-order placement, missing-AI message, auto-battle block), and the new `summon.ts` predicate module (full template/spawned/clone matrix, legacy fallback, masking-absence, splice determinism, `summonsAllowedInBattle`).
- `bun run typecheck` → exit 0; import cycle (`summon ↔ actions ↔ tags`) verified runtime-safe (combat route serves 200).
- New test files added; no schema/migration changes.

## Notes / follow-ups

- Summon jutsus reference an AI creature by `aiId`, which must exist as a `UserData` row to be loaded as a template. Only **jutsu** and **item** summon effects are scanned at battle init (not bloodline/skill effects). If a referenced AI is missing from the database, the summon now fails with a clearer message (*"…'s summon creature could not be found."*) rather than a generic error — this is a content/data concern, not a code issue (e.g. most summon AIs are unseeded in local dev DBs; production has them).

## License

## Summary by CodeRabbit

* **New Features**
  * Combat now recognizes a controlled actor, including piloted summons, for turn order, actions, targeting, and highlighting.
  * Summons can be marked as player-controlled, with battle state preserving this status.
  * Added summon lifecycle handling for orphaned summons and improved summon placement in turn order.

* **Bug Fixes**
  * Fixed turn handling so piloted summons use their own action set and aren’t treated like the session user.
  * Improved AI and wait-action behavior to respect piloted summons and avoid clearing selections unnecessarily.
  * Added safeguards so summon templates, clones, and orphaned summons are handled more consistently.


app/src/libs/combat/util.ts
```diff
@@ -2198,6 +2270,9 @@ export const alignBattle = (
   // 4. update all updatedAt fields on items & jutsus
   if (progressRound) {
     refillActionPoints(battle);
+    // Orphan cleanup: drop any summon whose controller has died/fled/left
+    // before the next round's actors are picked.
+    spliceOrphanedSummons(battle.usersState, battle.usersEffects);
     battle.round = actionRound;
```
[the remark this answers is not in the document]
ifroshty: Keeping both calls — they are not redundant.

`alignBattle` is called from three sites:

* `combat.ts:240`
* `combat.ts:577`
* `combat.ts:671`

Only the `671` path has a preceding router-level `spliceOrphanedSummons` call at `combat.ts:668`.

The `spliceOrphanedSummons` call inside `alignBattle` at `util.ts:2228` is still needed for the `240` and `577` paths. Without it, those paths could select a controllerless summon.

The two calls also serve different timing purposes:

* The router-level call runs **before** `calcActiveUser` selects the next actor, so the immediate pick / “now X’s turn” message never names a removed summon.
* The internal `alignBattle` call runs **after selection** during round progression, cleaning up for the next round or caller.

The only overlap is one `O(n)` no-op rescan at the `668` / `671` site when the round advances. Since `spliceOrphanedSummons` early-returns when there are no orphans, this is negligible over a small combatant list.

I added a comment at the router call site documenting why the pairing is intentional.

app/src/libs/combat/tags.ts
```diff
@@ -3050,25 +3105,36 @@ export const summon = (
...
-    if (ai && idx > -1) {
+    // The spawned summon is tracked by controllerId (creatorId), since aiId now
+    // stays pointed at the template. Remove this controller's summon.
+    const idx = usersState.findIndex(
+      (u) => u.isSummon && u.controllerId === effect.creatorId,
+    );
+    if (idx > -1) {
+      const removed = usersState[idx];
       usersState.splice(idx, 1);
       return {
```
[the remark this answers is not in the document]
ifroshty: Already handled in the current code.

In `tags.ts` around `3106-3121`, the expiry path prefers a non-live matching summon first:

```ts
findIndex(
  u =>
    u.isSummon &&
    u.controllerId === effect.creatorId &&
    !isLiveSummon(...)
)
```

It only falls back to any matching summon if no non-live match is found.

So in the dead `S1` + live re-summoned `S2` case, it removes dead `S1` and leaves live `S2` intact.

This is also scoped by:

```ts
controllerId === effect.creatorId
```

So it will not touch a template summon, since templates keep `controllerId = aiId`, and it will not affect another controller’s summon.

MathiasGruber:
Rebased this branch onto the latest `main` to clear the merge conflicts.

**Conflict resolved:**
- `app/tests/libs/combat/actions.test.ts` — the `@/libs/hexgrid` mock block conflicted. `main` had already landed an equivalent fix for leaked hexgrid test mocks (commit *Fix leaked hexgrid test mocks*) that serves the same purpose as this branch's "bun-test hardening" change. Kept `main`'s version: `getPossibleActionTiles` returns the passed grid and `PathCalculator` exposes `getShortestPath`. Both mocked functions are import-time guards only (not referenced in any test body), so the two variants are functionally equivalent for this file; the branch's now-redundant explanatory comment was dropped.

No migration conflicts were involved. The 16 branch commits replayed cleanly otherwise, and the PR is now mergeable.

app/src/layout/Combat.tsx
```diff
@@ -277,6 +311,11 @@ const Combat: React.FC<CombatProps> = (props) => {
       setBattleState({ battle: battleRef.current, result: null, isPending: true });
     },
     onSuccess: async (data) => {
+      // Clear the selected action when an action resolves. Control is
+      // sequential, so resetting here also clears the highlight when the
+      // turn hands off to (or away from) a piloted summon. Reset here in
+      // onSuccess rather than a useEffect (React Compiler lints the effect form).
+      setCombatActionId(undefined);
```
MathiasGruber: `setCombatActionId(undefined)` runs on every successful `performAction` mutation, not just when control changes hands. Pre-PR the selected action persisted across actions, letting a player chain the same jutsu/item across multiple AP spends in one turn. Now the action highlight clears after each successful action (including no-op responses like `{ notification: "Not your turn" }`, since onSuccess fires for any returned payload), forcing a re-click for every action. Consider gating this on an actual controlled-actor handoff (e.g. only clear when the response indicates the active actor changed, or compare controlledActorId before/after) rather than unconditionally.

app/src/libs/combat/tags.ts
```diff
@@ -2975,18 +2983,42 @@ export const summon = (
...
-      const obj = usersState.find(
-        (u) =>
-          u.username === ai?.username && u.curHealth && u.controllerId === user.userId,
-      );
-      if (ai && !obj) {
+      // One-summon-per-controller cap, keyed consistently on
+      // (controllerId === user.userId && isSummon). Allows re-summon once the
+      // prior summon is no longer live (so the AI does_not_have_summon gate
+      // and the player re-cast both work).
+      const alreadyHasSummon = hasLiveSummon(usersState, user.userId, userEffects);
```
MathiasGruber: Behavioral regression in the summon cap: `hasLiveSummon` matches any live `isSummon` with `controllerId === user.userId`, but `clone()` at line 1187 also sets `newAi.isSummon = true` on clones (with the same `controllerId`). So a live clone now blocks the player from summoning, whereas the pre-PR check matched on the AI template's username and allowed summon+clone to coexist. The same widened match also affects the AI `does_not_have_summon` check in `ai_v2.ts` (now routed through `isLiveSummon`). If that's intentional, please add a test/comment; otherwise key the cap on the template identity (e.g. `controllerId === effect.aiId` or an explicit non-clone flag) rather than on `isSummon` alone.

app/src/libs/combat/util.ts
```diff
@@ -2224,6 +2234,21 @@ export const hasNoAvailableActions = (
...
+ * Determine whose turn the dispatch loop should treat this actor as.
+ * A piloted summon (isPiloted) is human-driven on its turn even though it
+ * keeps isAi=true for accounting; only its controller may act for it.
+ */
+export const getTurnControl = (
+  actor: Pick<BattleUserState, "isAi" | "isPiloted" | "controllerId">,
+  sessionUserId: string,
+): { isUserTurn: boolean; isAITurn: boolean } => {
+  const isMyActor = actor.controllerId === sessionUserId;
+  const isUserTurn = isMyActor && (!actor.isAi || !!actor.isPiloted);
```
MathiasGruber: DRY: `isUserTurn` re-derives the human-action-set predicate (`!actor.isAi || !!actor.isPiloted`) that `wantsHumanActionSet` (defined just above at line 2202-2204) already encapsulates. Reuse it for a single source of truth: `const isUserTurn = isMyActor && wantsHumanActionSet(actor);`.

MathiasGruber (commented):

app/src/libs/combat/tags.ts
ifroshty: Confirmed regression. A live clone tripped both the cap and the AI does_not_have_summon gate. Added an explicit isClone predicate (isSummon && isOriginal === false, matching how process.ts already treats !isOriginal as a clone) and excluded clones from both hasLiveSummon and the ai_v2 gate, so a summon and clone coexist again. Left isLiveSummon untouched so orphan cleanup still reaps clones. Also set newAi.isOriginal = true explicitly in the summon spawn block so a future caller can't fool the discriminator, and added a regression test that a lone live clone doesn't trip the cap.

app/src/libs/combat/tags.ts
```diff
@@ -2975,18 +2983,45 @@ export const summon = (
...
+      if (alreadyHasSummon) {
+        effect.rounds = 0;
+        return {
+          txt: `${user.username} already has a summon!`,
+          color: "red",
+        } as ActionEffect;
+      }
```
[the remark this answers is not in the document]
ifroshty: The blocked-cast teardown doesn't reproduce. On the casting action realizeTag stamps createdRound = battle.round, so applyEffects recomputes castThisRound = true; summon() takes the isNew && castThisRound block branch and sets rounds = 0, and applyEffects then drops the effect in that same pass at process.ts:293 (isEffectActive(0) is false). 

It never persists into a later alignBattle, so the rounds === 0 teardown is never reached for a blocked cast — the existing summon is untouched. The rounds ≤ 0 retention in util.ts only applies to effects decremented to 0 by alignBattle in a prior round (normal expiry), not a freshly-blocked one. 

Added summon_teardown.test.ts that drives the real applyEffects pipeline and confirms the live summon survives a blocked re-cast across two actions (with a positive control that does trigger the teardown).

Separately, the teardown's findIndex could match a clone (clones share isSummon + controller), so I added !isClone(u) to both predicates — un-summon now only targets real summons; clones keep their own clone() teardown. Covered by a test.

app/src/libs/combat/tags.ts
```diff
@@ -8,6 +8,12 @@ import type {
   StatType,
 } from "@/drizzle/constants";
 import type { Battle } from "@/drizzle/schema";
+import {
+  hasLiveSummon,
+  isLiveSummon,
+  shouldPilotSummon,
+  summonsAllowedInBattle,
+} from "@/libs/combat/summon";
```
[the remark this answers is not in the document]
ifroshty: fixed latest commit

app/src/libs/combat/ai_v2.ts
```diff
@@ -261,11 +262,14 @@ export const performAIaction = (
...
+            // isClone excludes the AI's clones, which are not its summon.
+            return !nextBattle.usersState.some(
               (u) =>
-                u.controllerId === user.userId &&
-                u.isSummon &&
-                stillInBattle(u, effects),
+                isLiveSummon(u, nextBattle.usersState, effects) &&
+                !isClone(u) &&
+                u.controllerId === user.userId,
             );
```
MathiasGruber: DRY: this predicate is exactly the negation of `hasLiveSummon` from `@/libs/combat/summon` (same `isLiveSummon && !isClone && controllerId === X` triple). The whole block could be replaced with `!hasLiveSummon(nextBattle.usersState, user.userId, effects)`, keeping the AI `does_not_have_summon` gate and the summon() one-per-controller cap in lock-step by construction (any future tweak to what "has a summon" means only needs to change one predicate).

app/src/server/api/routers/combat.ts
```diff
@@ -684,6 +698,7 @@ export const combatRouter = createTRPCRouter({
           // Check if we should let the inner-loop continue
           if (
             newActor.isAi && // Continue new loop if it's an AI
+            !newActor.isPiloted && // but never auto-drive a human-piloted summon
```
MathiasGruber: DRY: `newActor.isAi && !newActor.isPiloted` re-derives what `getTurnControl` (already imported and used a few lines above) computes as `isAITurn`. Since this is the same actor whose control status was just resolved, doing `const { isAITurn: newIsAITurn } = getTurnControl(newActor, suid);` and gating on `newIsAITurn` keeps the "who acts autonomously" definition in a single place, so a future change (e.g., a third control mode) doesn't have to touch both the top-of-block guard and this continue-gate.

MathiasGruber (commented):
Automated code review — logic/DRY/readability

MathiasGruber (commented):
## Review — player-piloted summons

I read the full diff, traced the summon/clone lifecycle end to end against `main`, and ran the suite, typecheck and lint locally on `2f92788`.

**Overall:** the core piloting design is sound. Keeping `isAi = true` and adding a separate `isPiloted` turn-routing flag is the right call — it leaves reward attribution, drop tables and the AI accounting path untouched — and `getTurnControl` correctly denies another player's piloted summon (`isUserTurn` false, `isAITurn` false). Masking is right too: `controllerId` and `isPiloted` are public so every client can route turns, while the summon's private stats stay with its controller via the existing `controllerId`-based `maskUsersState` rule, so no new masking concept was needed. `playerControlled` defaults to `false` and `z.boolean()` renders as a real toggle in the tag editor, so existing content is unaffected and the flag is actually reachable for content editors.

What I'd want addressed before merge:

**1. CI is red.** `bun test` fails (2 failures in `summon_teardown.test.ts`), and the `test_build` check on this PR is failing for exactly those. The PR body's "140 passed / 0 failed" doesn't reproduce — the file only passes in isolation. Details inline.

**2. Battle-bricking crash.** `isOrphanedSummon` doesn't exclude clones, so a live clone whose creator dies or flees gets spliced, and `clone()`'s `throw new Error("Summoner not found")` then fires on every subsequent `applyEffects` pass, permanently. I reproduced it as a unit test on this branch. Every remaining participant gets `{ updateClient: false, notification: "Summoner not found" }` on every action, and the `getBattle` timeout path 500s. Details inline.

**3. The orphan-cleanup ordering isn't what the comments claim.** `alignBattle` splices *after* `calcActiveUser` has already chosen the actor, so it can pin `activeUserId` to a just-removed summon and name it in "It is now X's turn." Moving the splice to the top of `alignBattle` fixes it and lets the duplicate pre-splice in `combat.ts` be deleted.

### Simplification / reuse

* `resolveControlledActorId` is a third encoding of the rule `getTurnControl` already owns; it can be written as a one-line projection of it, which is a stronger guarantee of the client/server agreement its own doc comment asks for.
* It's also derived twice from the same object (`page.tsx` **and** `Combat.tsx`). Passing the value already computed in `page.tsx` as the `userId` prop removes the second derivation, the new `userIdRef` sync effect, and the "do NOT reset this ref" comment.
* The `AutoBattleTypes` summon block should live in the action-availability filter — `ai_v2.ts:80-87` already strips `summon` effects from AI candidates — rather than no-op'ing after the cost is paid.
* The un-summon lookup replaced an exact id with a two-pass heuristic. Adding a `summonedUserId` field to `SummonTag` (instead of overloading `aiId`) makes it exact and deletes the heuristic; the current version also picks the wrong summon when a dead and a live summon coexist.
* `shouldPilotSummon`'s `AutoBattleTypes` clause is unreachable, and two tests assert it.

### Undocumented behaviour changes

Three gameplay changes affect *all existing* summon/clone content, not just piloted summons, and aren't in the "Breaking changes" section:

* one-summon-per-**player** (was one-per-**creature** — two different summon jutsus could previously coexist);
* the summon is inserted right after its summoner instead of appended, so it now acts before all enemies on the cast round;
* summons and clones are despawned when their master dies or flees. In PvP this means killing the summoner instantly clears the board and can end the battle a turn earlier.

### One note on the changelog

Both headline "bugs found in review" were introduced by earlier commits in this branch, not by `main`. The merge-base (`1b91988`) already looks up the template with `u.controllerId === effect.aiId`, and re-cast already worked there (`realizeTag` builds a fresh ground effect per cast, so the old `effect.aiId` rebinding was scoped to one instance). Not a code problem — but the description reads as if the branch fixes pre-existing production bugs, which will mislead whoever reads it later.

### Verified locally on `2f92788`

* `bun tsc --noEmit` → clean (after `rm tsconfig.tsbuildinfo`; incremental builds mask errors here).
* `bun run lint` → warnings only, no new ones from this PR.
* `bun test` → 346 pass / **2 fail**; `bun test tests/libs/combat/` → 169 pass / **2 fail**.

app/tests/libs/combat/summon_teardown.test.ts
```diff
@@ -0,0 +1,209 @@
+import { describe, it, expect } from "vitest";
+import { applyEffects } from "@/libs/combat/process";
```
MathiasGruber, line 2: **Blocker — this file makes `bun test` (and CI `test_build`) red.**

The PR description claims *"bun test tests/libs/combat/ → 140 passed / 0 failed"*, but that is not reproducible. On the head commit:

```
bun test tests/libs/combat/   ->  169 pass,  2 fail
bun test                      ->  346 pass,  2 fail
```

and the `test_build` job on this PR is failing for exactly these two:

```
(fail) summon teardown — positive control > an EXPIRED summon effect (rounds=0, not new) tears down the live summon
(fail) blocked summon re-cast does NOT remove the existing live summon
TypeError: undefined is not an object (evaluating 'state.filter')
```

Cause: `tests/libs/combat/actions.test.ts` (and `barrier_aoe.test.ts`) do

```ts
vi.mock("@/libs/combat/process", () => ({
  applyEffects: vi.fn(() => ({ newBattle: {}, actionEffects: [] })),
  ...
}));
```

Bun shares one module registry across files in a run, so this file's `import { applyEffects }` resolves to that stub, `newBattle.usersState` is `undefined`, and `liveSummonsOf` throws. It passes only when run in isolation. You already knew about this hazard — `ai_v2_piloted_guard.test.ts` carries a long comment about Bun's shared registry — the same care just wasn't applied here.

Two of the three tests here don't actually need the real pipeline: the third (`summon expiry teardown targets the summon, not a clone`) already calls `summon()` directly and passes. Suggest driving the other two through `summon()` the same way (as `summon_template_lookup.test.ts` does), or re-stubbing `@/libs/combat/process` locally with a passthrough.

app/src/libs/combat/summon.ts
```diff
@@ -0,0 +1,128 @@
...
+/** An ACTIVE summon whose controller can no longer act (dead/fled/left/absent).
+ *  Templates are excluded by isActiveSummon; the stillInBattle(summon) guard
+ *  excludes dead summons; the tri-condition keeps the absent-controller branch
+ *  that shipped tests depend on. */
+export const isOrphanedSummon = (
+  summon: BattleUserState,
+  usersState: BattleUserState[],
+  effects: UserEffect[],
+): boolean => {
+  if (!isActiveSummon(summon, usersState)) return false;
```
MathiasGruber, line 187: **High — `isOrphanedSummon` also matches *clones*, and splicing a live clone permanently bricks the battle.**

Every other new predicate deliberately excludes clones (`hasLiveSummon`, the un-summon search in `tags.ts`, the `does_not_have_summon` gate). This one doesn't, and clones set `isSummon = true`, so a live clone whose creator dies/flees/leaves is spliced out of `usersState`.

That is fatal because `clone()` rebinds the ground effect to the spawned clone (`effect.creatorId = newAi.userId`) and then hard-throws when it can't find it:

```ts
const user = usersState.find((u) => u.userId === effect.creatorId);
if (!user) { throw new Error("Summoner not found"); }   // tags.ts:1178
```

Once the clone is gone, the clone ground effect can never be cleaned up — `alignBattle` explicitly retains `type === "clone"` effects at `rounds <= 0` so the tag function can run, but the tag function now throws before reaching the `rounds === 0` teardown branch. So it throws forever:
* `performAction` → `performBattleAction` → `applyEffects` → caught → `{ updateClient: false, notification: "Summoner not found" }` for **every remaining participant**, every action.
* `getBattle`'s timeout path (`applyEffects(userBattle, actId)`, combat.ts:265) has no local catch — after 3 retries the outer handler rethrows, so the combat page 500s.

Reproduced as a unit test against this branch (passes, i.e. the throw happens):

```ts
const usersState = [player, ally];
const effect = mkCloneEffect();               // creatorId: "p1"
clone(usersState, effect, { jutsus: {} });    // effect.creatorId now = clone's id
player.fledBattle = true;                     // creator leaves
spliceOrphanedSummons(usersState, []);        // -> clone removed
clone(usersState, effect, { jutsus: {} });    // throws "Summoner not found"
```

Note a dead clone is safe (the `stillInBattle(summon)` guard skips it) — it's specifically *alive clone + gone creator*, i.e. any 2v2 / raid where a cloner flees or dies while their clone stands.

Fix: `if (isClone(summon)) return false;` here, which also makes this predicate consistent with the other three. Hardening `clone()` to tolerate a missing creator instead of throwing would be worth doing regardless.

app/src/libs/combat/util.ts
```diff
@@ -2262,6 +2287,9 @@ export const alignBattle = (
   // 4. update all updatedAt fields on items & jutsus
   if (progressRound) {
     refillActionPoints(battle);
+    // Orphan cleanup: drop any summon whose controller has died/fled/left
+    // before the next round's actors are picked.
+    spliceOrphanedSummons(battle.usersState, battle.usersEffects);
```
MathiasGruber: **The splice runs *after* the actor has already been selected, so the invariant the router comment claims isn't actually held.**

Order in `alignBattle` is: `calcActiveUser` picks `actor` → (if `progressRound`) `spliceOrphanedSummons` → `battle.activeUserId = actor.userId`. So when the actor that was just picked *is* the orphan, this call removes it and then pins `activeUserId` to a user that no longer exists in `usersState`, and returns a dangling `actor` to the caller. `performAction` then uses that `newActor` for `"It is now X's turn."` and for the `newActor.isAi && !newActor.isPiloted` loop-continue — exactly the two things combat.ts:657-664 says this can't happen. It self-heals on the next `calcActiveUser` (check 3), but only after a wasted round-trip and a log line naming a ghost.

Simplest fix, which also removes the duplication: move this call to the **top of `alignBattle`, before `calcActiveUser`, unconditional** (not gated on `progressRound`). Then the pre-splice added at combat.ts:665 can be deleted, all three `alignBattle` call sites get the guarantee, and there's one place that owns orphan cleanup instead of two with a comment explaining why they overlap.

app/src/server/api/routers/combat.ts
```diff
@@ -650,6 +654,16 @@ export const combatRouter = createTRPCRouter({
...
 
+          // Mid-round orphan cleanup BEFORE picking the next actor: a controller
+          // may have died/fled/left on an earlier actor's turn (e.g. the enemy's).
+          // Removing its now-masterless summon here guarantees alignBattle selects
+          // a valid actor (never a summon about to be spliced) and that the "It is
+          // now X's turn" line below never names a removed summon. alignBattle also
+          // splices on round-progress, but only AFTER actor selection and only for
+          // callers without this pre-splice; the two cover the immediate pick vs.
+          // future rounds, so the brief overlap (a no-op rescan) is intentional.
+          spliceOrphanedSummons(newBattle.usersState, newBattle.usersEffects);
```
MathiasGruber: This duplicate call (and the 8-line comment justifying the overlap) goes away if `spliceOrphanedSummons` is moved to the top of `alignBattle`, before `calcActiveUser` — see the comment on `util.ts`. As written the guarantee is also incomplete: the `alignBattle` at line 571 (top of the inner loop, first iteration straight off the DB) has no pre-splice at all.

app/src/libs/combat/tags.ts
```diff
@@ -2975,18 +2984,45 @@ export const summon = (
 
   if (effect.isNew && effect.castThisRound) {
     effect.isNew = false;
+    // No summons in auto-resolved battle types (KAGE_AI / CLAN_CHALLENGE), where
+    // there are no human turns — not even AI-cast ones.
+    if (!summonsAllowedInBattle(battle)) {
+      effect.rounds = 0;
```
MathiasGruber, line 3114: **Silent failure at the wrong layer.** By the time `summon()` runs, the caster has already paid chakra/stamina/AP and burned the turn. This returns `undefined`, so there is no `ActionEffect` either — nothing in the battle log explains where the turn went.

Worse for `KAGE_AI`: the AI's `does_not_have_summon` condition stays true forever, so an AI kage with a summon jutsu will re-cast it *every round* for the whole fight and accomplish nothing.

There is already a filter for exactly this shape in `ai_v2.ts:80-87`, which strips `move`/`summon` actions from the AI's candidate list for user-backed AIs:

```ts
if (isUser && action.id !== "move") {
  if (action.effects.some((e) => "type" in e && e.type === "summon")) return false;
}
```

Extending that condition (or `availableUserActions`, next to the existing `isSummonPrevented` handling) to cover `AutoBattleTypes` keeps the action from being *chosen* rather than no-op'ing it after payment, and reuses machinery that already exists.

Separately: is this restriction wanted at all? `shouldPilotSummon` already returns `false` for `AutoBattleTypes`, so an AI-driven summon in a kage battle was never at risk of needing a human turn. Blocking them outright is an unrelated gameplay nerf riding along with the piloting feature — it is listed under "Breaking changes", but it isn't required by it.

app/src/libs/combat/tags.ts
```diff
@@ -3055,25 +3091,52 @@ export const summon = (
...
-    const ai = usersState.find((u) => u.userId === effect.aiId);
-    const idx = usersState.findIndex((u) => u.userId === effect.aiId);
-    if (ai && idx > -1) {
+    // The spawned summon is tracked by controllerId (creatorId), since aiId now
+    // stays pointed at the template. Prefer removing a non-live (dead/expired)
+    // summon: a re-cast inserts the new live summon ahead of an older dead one,
+    // so a plain first-match could remove the live one instead.
+    // Exclude clones: they share isSummon + the same controllerId, so an
+    // un-summon must never target the controller's clone instead of the summon.
+    let idx = usersState.findIndex(
```
MathiasGruber: **This replaces an exact identity with a two-pass heuristic; the exact version is available and simpler.**

Dropping `effect.aiId = newAi.userId` was right (it clobbered the template pointer), but the answer doesn't have to be a guess. Store the spawned id in its own field instead of overloading `aiId` — e.g. add `summonedUserId` alongside `aiId` in `SummonTag`, set it at spawn, and the whole block collapses to:

```ts
const idx = usersState.findIndex((u) => u.userId === effect.summonedUserId);
```

That is exact, needs no `isClone` exclusion, no live/non-live preference, and no fallback pass.

The heuristic is also observably wrong in one case: "prefer a non-live summon" assumes the only non-live candidate is a stale one, but a controller can legitimately have a dead summon *and* a live summon at once (summon A rounds=10 → A dies → re-summon B rounds=3, allowed because the cap only counts *live* summons). When B's effect expires it removes **A**, and B then survives until A's effect expires seven rounds later.

app/src/libs/combat/summon.ts
```diff
@@ -0,0 +1,128 @@
...
+ *  coexist (as they did pre-pilot). It is the only live isSummon entity created
+ *  with isOriginal === false (clone() sets it; summons keep the template's
+ *  isOriginal === true), mirroring how process.ts already treats !isOriginal as
+ *  a clone. */
+export const isClone = (u: Pick<BattleUserState, "isSummon" | "isOriginal">): boolean =>
+  !!u.isSummon && u.isOriginal === false;
+
+/** One-summon-per-controller cap predicate. Clones are excluded (see isClone)
+ *  so they neither block summoning nor are blocked by a summon. */
+export const hasLiveSummon = (
```
MathiasGruber: **Undocumented gameplay change: the cap went from per-creature to per-player.**

Before (merge-base `tags.ts`):

```ts
const obj = usersState.find(
  (u) => u.username === ai?.username && u.curHealth && u.controllerId === user.userId,
);
```

i.e. "you can't have two of *the same* creature out" — two different summon jutsus could be active simultaneously. `hasLiveSummon` keys only on `controllerId`, so now it's one summon total per player, for every existing summon jutsu in the game. That may well be the intent, but it isn't in the "Breaking changes" section and it silently rebalances existing content.

Same for the "already has a summon!" red message: it costs the player the full chakra/AP for the cast. Worth deciding whether a blocked cast should refund, or whether the jutsu should just be unavailable (`availableUserActions` already has `isSummonPrevented` for exactly this kind of gate).

app/src/libs/combat/tags.ts
```diff
@@ -3055,25 +3091,52 @@ export const summon = (
           userEffects.push(realizedEffect);
           return realizedEffect;
         });
-        // Push to userState
-        usersState.push(newAi);
+        // Insert the summon immediately after its summoner so it takes its turn
+        // right after them (turn order follows usersState array order). Falls
+        // back to appending if the summoner can't be located.
+        const summonerIdx = usersState.findIndex((u) => u.userId === user.userId);
+        if (summonerIdx >= 0) {
```
MathiasGruber: Another unlisted balance change: `usersState` order **is** turn order (`calcActiveUser` walks it as a ring, and `initiateBattle` establishes it via `usersState.sort((a, b) => b.initiative - a.initiative)`). Previously the summon was appended, so it acted at the end of the round; now it acts immediately after its summoner, ahead of every enemy — a straight tempo buff on the round it's cast, for all existing summon jutsus.

It's a reasonable choice for piloting (the player gets both turns back to back), but it should be called out, and it does break the "array is initiative-sorted" invariant. If the intent is initiative-based placement, setting `newAi.initiative` from the summoner and inserting by initiative would be more consistent with how the array is built.

app/src/libs/combat/actions.ts
```diff
@@ -1416,6 +1416,31 @@ export const stillInBattle = (
...
 
+/**
+ * The actor `myUserId` controls right now: themselves on their turn, or their
+ * piloted summon on the summon's turn (control is sequential, so exactly one).
+ * Shared by Combat.tsx (reads battleRef) and combat/page.tsx (reads the reactive
+ * battle) so the action menu and the action submit/validate can never disagree
+ * about whose turn it is. Returns myUserId when no piloted summon of theirs is
+ * the active actor.
+ */
+export const resolveControlledActorId = (
```
MathiasGruber: **This is a third encoding of a rule that `getTurnControl` already owns — and it can be expressed in terms of it.**

The doc comment says the point is that "the action menu and the action submit/validate can never disagree about whose turn it is". The strongest way to get that is to not restate the predicate:

```ts
export const resolveControlledActorId = (
  battle: ReturnedBattle | null | undefined,
  myUserId: string | undefined,
): string | undefined => {
  const actor = battle?.usersState.find((u) => u.userId === battle.activeUserId);
  return actor && myUserId && getTurnControl(actor, myUserId).isUserTurn
    ? actor.userId
    : myUserId;
};
```

I checked the truth table against the current implementation — identical for all five cases the tests cover (own turn, own piloted summon, own non-piloted summon, another player's piloted summon, no battle), plus `AutoBattleTypes` where the human is `isAi = true`. Now there is literally one predicate, and `resolveControlledActorId` is a projection of it.

Two smaller points:
* **Placement** — every other summon/piloting predicate went into the new `libs/combat/summon.ts`; this one landed in `actions.ts` (already ~1500 lines). It belongs with its siblings, or in `util.ts` next to `getTurnControl`.
* **Typing** — the ad-hoc structural parameter (`{ activeUserId; usersState: { userId; isPiloted?; controllerId }[] }`) is the only place in this area that doesn't use `ReturnedBattle`. `data.battleUpdate` (`ReturnedBattleDynamic`) and `battleState.battle` both satisfy `ReturnedBattle`'s `usersState`, so the real type works and stays consistent with the rest of the file.

app/src/app/combat/page.tsx
```diff
@@ -126,7 +132,7 @@ export default function CombatPage() {
         />
       )
     );
-  }, [versionId, actionId, userId, results, config.showGridNumbers]);
+  }, [versionId, actionId, userId, results, config.showGridNumbers, actions]);
```
MathiasGruber: `controlledActorId` is derived here (line 51) and then derived **again** inside `Combat` from the same `battleState.battle` object that is passed down as a prop. That duplication is what forces the rest of the churn in `Combat.tsx`: a second `resolveControlledActorId` call, a new `useEffect` to push it into `userIdRef`, and the removal of `userIdRef.current = props.userId` from the props-sync effect (with a comment explaining that the ref must *not* be reset).

Passing the already-computed value would collapse all of that:

```tsx
<Combat ... userId={controlledActorId} />
```

`Combat` uses `props.userId` only to seed `userIdRef`; everything else reads `suid` from `useRequiredUserData()` directly. So the original `userIdRef.current = props.userId` sync in the props effect just works again, and both the extra derivation and the extra effect can go. (`controlledActorId` should then also be added to this dep array explicitly rather than riding on `actions`.)

app/src/libs/combat/summon.ts
```diff
@@ -0,0 +1,128 @@
...
+ *  has human turns. With the flag off (the default) the summon stays AI-driven. */
+export const shouldPilotSummon = (
+  controller: Pick<BattleUserState, "isAi"> | undefined,
+  battle: Pick<Battle, "battleType">,
+  playerControlled: boolean | undefined,
+): boolean =>
+  !!playerControlled &&
+  !!controller &&
+  !controller.isAi &&
+  !AutoBattleTypes.includes(battle.battleType);
```
MathiasGruber: Dead condition: `summon()` returns early via `summonsAllowedInBattle(battle)` before `shouldPilotSummon` is ever reached, so this `AutoBattleTypes` check can never fire. The two `piloted_summon.test.ts` cases that assert it ("does NOT pilot ... in KAGE_AI / CLAN_CHALLENGE") are testing unreachable code and give false confidence that the auto-battle path is covered.

Either drop the clause here (the caller already guards), or drop `summonsAllowedInBattle` and let `shouldPilotSummon` be the single `AutoBattleTypes` check — but not both.

app/src/layout/Combat.tsx
```diff
@@ -389,18 +437,22 @@ const Combat: React.FC<CombatProps> = (props) => {
...
+      // [suid, timeDiff], so those would be stale. calcActiveUser selects the actor
+      // from battle.activeUserId, which is all the "w" (wait) path needs.
+      const { actor } = calcActiveUser(battleRef.current, suid, timeDiff);
       switch (event.key) {
         case "w":
-          if (actor.userId === suid) {
+          // Controller-based: true on my own turn AND on my piloted summon's
+          // turn (actor.controllerId === suid). userId carries the acting
+          // entity's id (the summon's, on its turn).
+          if (actor.controllerId === suid) {
```
MathiasGruber: Two things here:

1. **Use the shared predicate.** `actor.controllerId === suid` is true for a *non-piloted* AI summon or a clone you own too, so "w" fires `performAction` on your AI summon's turn. The server then falls into the `isAITurn` branch and runs `performAIaction` instead of the wait — harmless, but it's the third hand-rolled variant of the same rule. `getTurnControl(actor, suid).isUserTurn` is exactly the condition you want and matches what the server authorises with.

2. **The precompute was dropped for no gain.** `calcActiveUser` without `precomputedActions` makes `hasNoAvailableActions` recompute `availableUserActions` on *every keydown anywhere on the page*. The staleness the comment worries about is already solved by the refs this PR introduced — `userIdRef.current` and `precomputedActionsRef.current` are both kept current and are exactly what the click handler passes at line 776-777. Passing them here restores the optimisation *and* makes it work on the summon's turn (the old `precomputedUserId: suid` only ever hit on your own turn).

Also note `userId: actor.userId` in the payload below is cosmetic — `performAction` never reads `input.userId`; the server derives the actor itself.

app/src/libs/combat/summon.ts
```diff
@@ -0,0 +1,128 @@
...
+  const removed: string[] = [];
+  for (let i = usersState.length - 1; i >= 0; i--) {
+    const u = usersState[i];
+    if (u && orphanIds.has(u.userId)) {
+      removed.push(u.userId);
+      usersState.splice(i, 1);
+    }
+  }
+  removed.reverse();
+  return removed;
```
MathiasGruber: Both call sites ignore the return value, so the `removed` array, the reverse-iteration and the final `removed.reverse()` are all dead work. Either make it `void`, or — better — use it: right now a summon just silently vanishes from the battlefield when its master falls. `summon()`/`clone()` both emit an `ActionEffect` ("was unsummoned!", "disappears!") for their own teardown; the orphan path should say something equivalent in the battle log rather than leaving players to wonder where their creature went.

app/src/libs/combat/ai_v2.ts
MathiasGruber: Done in a95c76f. The whole block is now:

```ts
case "does_not_have_summon":
  // Same predicate as the one-per-controller cap in summon(), so the
  // AI's re-summon gate can never drift from what blocks the cast.
  return !hasLiveSummon(nextBattle.usersState, user.userId, effects);
```

`isClone` / `isLiveSummon` are no longer imported here — `hasLiveSummon` already composes both.

app/src/server/api/routers/combat.ts
MathiasGruber: Done in a95c76f:

```ts
const { isAITurn: newIsAITurn } = getTurnControl(newActor, suid);
if (
  newIsAITurn && // Continue new loop if the next actor acts autonomously
  nActions < maxActions &&
  ...
```

So the top-of-block authorization and this continue-gate now read from the same helper. A third control mode would only need `getTurnControl` changed.

app/tests/libs/combat/summon_teardown.test.ts
MathiasGruber: Resolved, but not the way I expected — worth recording why.

The rebase alone fixed it: `main` had already removed the `applyEffects` stub from `actions.test.ts` and `barrier_aoe.test.ts` (both now override only `checkFriendlyFire`, with comments naming the summon/poison damage-credit suites as the victims). This branch was 258 commits behind, so it was still seeing the old stub.

There *was* a second, opposite instance of the same bug that the rebase did not fix, and CI caught it: this PR's own `ai_v2_piloted_guard.test.ts` mocked `@/libs/hexgrid` with `getPossibleActionTiles: () => new Set()`, which is process-global under `bun test` and crashed `barrier_aoe.test.ts` on `restrictGrid.getHex`. Fixed in 8a661da by dropping that mock — the real module works there, since both summons have no jutsu/item and empty pools.

Now green: `bun test` 1054 pass / 0 fail, and I check the combat directory in reversed file order too, since default ordering is what hid this.

app/src/libs/combat/summon.ts
MathiasGruber: Fixed in f88831e.

`isOrphanedSummon` now short-circuits on `isClone`, matching `hasLiveSummon`, the un-summon lookup and the `does_not_have_summon` gate. I also hardened `clone()` to expire its effect instead of throwing when the creator is missing, so no future removal path can brick a battle the same way.

Three regression tests, verified to fail without the fix — including an end-to-end one that runs the real `applyEffects` pipeline: clone spawns → creator flees → orphan cleanup → the ally's next action still resolves.

app/src/libs/combat/util.ts
MathiasGruber: Fixed in 0cea28a. `spliceOrphanedSummons` now runs at the top of `alignBattle`, before `calcActiveUser`, and unconditionally rather than gated on `progressRound`.

That closes both halves: the ordering hole, and the fact that the router's workaround only covered one of the three `alignBattle` call sites. Orphans now also clear mid-round instead of at the next round boundary — which is what the router pre-splice was already doing on every `performAction`, so that path is unchanged and only `getBattle` becomes more eager.

New `align_battle_orphan.test.ts`; 3 of its 4 cases fail without the fix, the fourth being the negative control (a summon whose controller is still fighting must survive).

app/src/server/api/routers/combat.ts
MathiasGruber: Removed in 0cea28a, along with the 8-line comment and the now-unused import. `alignBattle` owns orphan cleanup outright.

app/src/libs/combat/tags.ts
MathiasGruber: Fixed in ce45d181.

`summonedUserId` is recorded on the effect at spawn and matched exactly, so no heuristic is involved. It lives on `BattleEffect` (the runtime augmentation layer, alongside `creatorId`/`castThisRound`) rather than on `SummonTag`, which keeps it out of the content editor while `aiId` stays pointed at the template so re-cast still works.

The controllerId search survives only as a fallback for battles already in flight at deploy, and is marked deletable once none remain. Added a test for the case the heuristic actually got wrong: with a dead and a live summon both present, the live summon's effect now tears down the live summon instead of the dead one.

app/src/libs/combat/summon.ts
MathiasGruber: Not changed — flagging for your decision rather than picking one myself.

This is a gameplay rule, not a defect. Pre-PR the cap was per-creature (`u.username === ai?.username`), so two *different* summon jutsus could be active at once; `hasLiveSummon` keys only on `controllerId`, making it one summon per player across all existing summon content.

That may well be what you want, and the cap is load-bearing for piloting (two piloted summons would give one player three turns a round). But it silently rebalances existing jutsu and isn't in the Breaking changes list. Either is a one-line change from here — add the creature identity back to the predicate, or document it. Tell me which and I'll do it.

Related: a blocked cast still costs the full chakra/AP. If the cap stays, gating the jutsu in `availableUserActions` (next to the existing `isSummonPrevented` handling) would be kinder than charging for a red message.

app/src/libs/combat/tags.ts
MathiasGruber: Not changed — same reasoning as the cap; this is a balance call.

`usersState` order is turn order (`initiateBattle` seeds it with `sort((a, b) => b.initiative - a.initiative)` and `calcActiveUser` walks it as a ring). Inserting at `summonerIdx + 1` means the summon now acts before every enemy on the round it is cast, where previously it was appended and acted last. That is a tempo buff to all existing summon jutsu, not just piloted ones.

It's a good fit for piloting specifically — the player gets both turns back to back — so I've left it. If you want it kept, it belongs in Breaking changes; if you'd rather it not apply to AI-driven summons, gating the insert on `newAi.isPiloted` and appending otherwise is a two-line change.

app/src/libs/combat/actions.ts
MathiasGruber: Fixed in a95c76f. It's now a projection of `getTurnControl`:

```ts
const actor = battle.usersState.find((u) => u.userId === battle.activeUserId);
return getTurnControl(actor, myUserId).isUserTurn ? actor.userId : myUserId;
```

Moved to `util.ts` beside `getTurnControl`, and typed `Pick<ReturnedBattle, "activeUserId" | "usersState">` — which accepts both the full battle and the extraState-less `battleUpdate`, and immediately caught that the old test fixture wasn't a real user shape (it now builds users through `makeBattleUser`).

One predicate now decides who acts, on both sides of the wire.

app/src/app/combat/page.tsx
MathiasGruber: Fixed in a95c76f. `page.tsx` resolves the controlled actor once and passes it as `userId`; `Combat` reads `props.userId`.

That deleted the second `resolveControlledActorId` call, the `useEffect` that owned `userIdRef`, and the "do NOT reset this ref" comment — the props-sync effect just assigns `userIdRef.current = props.userId` again. `controlledActorId` is now an explicit dep of the `combat` memo rather than riding on `actions`.

app/src/libs/combat/summon.ts
MathiasGruber: Fixed in a95c76f. `shouldPilotSummon` is now `(controller, playerControlled)` with no battle parameter, and the two tests that pinned the unreachable branch are gone — `summonsAllowedInBattle` owns the `AutoBattleTypes` rule, tested in `summon.test.ts`.

app/src/layout/Combat.tsx
MathiasGruber: Fixed in a95c76f, both parts.

The guard is `getTurnControl(actor, suid).isUserTurn`, so "w" fires on your own turn and your piloted summon's, but not on your AI summon's. And it passes `userIdRef.current` / `precomputedActionsRef.current` into `calcActiveUser` — both kept current — so the precompute is reused instead of recomputing `availableUserActions` on every keypress, and it now hits on the summon's turn too (the old `precomputedUserId: suid` only ever matched your own).

app/src/libs/combat/summon.ts
MathiasGruber: Fixed in a95c76f. `spliceOrphanedSummons` now returns the removed combatants rather than ids — you can't look up a user you just deleted, which is a bug I wrote and caught while wiring this up — and `alignBattle` passes their names out so `performAction` can push an ActionEffect:

> Ashura vanishes without its summoner!

The reverse-iterate-then-reverse dance is gone too; it's a filter-and-refill now.

app/src/libs/combat/tags.ts
MathiasGruber: Fixed in a95c76f, at the layer you'd expect rather than in \`summon()\`.

The AI's candidate list now drops summon actions in \`AutoBattleTypes\`, right next to the existing \`isUser\` summon/move filter in \`ai_v2.ts\`. So the AI never selects one, never pays the cost for a no-op, and \`does_not_have_summon\` no longer keeps it re-casting every round.

I deliberately kept the \`summonsAllowedInBattle\` guard in \`summon()\` as a backstop rather than removing the restriction outright. I still think blocking AI-cast summons in auto battles is a gameplay change this PR didn't need — \`shouldPilotSummon\` already made piloting impossible there — but it's documented under Breaking changes, so reverting it is your call, not mine to make silently.

app/src/libs/combat/summon.ts
MathiasGruber: Resolved per your call: **keep "no two of the same creature", allow multiple different summons, and cap player-controlled summons at one.** Implemented in 563d3c3.

The two caps are now independent, in one predicate (`summonCastBlockedReason`):

```ts
const live = liveSummonsOf(usersState, controllerId, effects);
const duplicate = live.find(...);
if (duplicate) return `${duplicate.username} is already summoned`;
if (wantsPilot && live.some((u) => u.isPiloted)) {
  return "You can only control one summon at a time";
}
return null;
```

The pilot cap keys on `wantsPilot`, so a non-piloted summon neither blocks nor is blocked — a piloted Demon plus any number of AI-driven creatures is fine.

Identifying "the same creature" needed a new transient `summonSourceId`: the spawn rebinds `controllerId` to the summoner, so nothing else links a live summon back to its template. A test caught the backward-compat gap — summons in battles that started before this field exists have no `summonSourceId` — so the original `username` comparison is kept as the fallback for exactly those.

**On the cost**, you were right that `availableUserActions` is the wrong lever, so the check moved to the router instead, immediately before `performBattleAction`:

```ts
if (summonBlocked) {
  return { updateClient: false, notification: summonBlocked };
}
```

Pools are deducted inside `insertAction` (actions.ts:1110), well after entry, so returning early costs nothing. The jutsu stays visible in the action bar, and a blocked cast produces a toast naming the reason rather than silently eating the chakra.

I also considered throwing from `insertAction` next to the existing `Not enough chakra` checks, which reads more naturally — but `performAIaction` calls `performBattleAction` with no try/catch, so any AI that picked a capped summon would abort its whole turn. `summon()` keeps the same predicate as a backstop, which is the path AI casts actually hit; they expire the effect exactly as they did pre-PR.

12 new tests in `summon_caps.test.ts` cover both caps, their interaction, re-cast after death, other players' summons, the legacy fallback, and clones tripping neither. Full suite 1066 pass / 0 fail; CI green.

app/src/libs/combat/tags.ts
MathiasGruber: Both halves are now closed.

**AI wasting the turn** — fixed in a95c76f by the filter you pointed at. Summon actions are stripped from the AI's candidate list in auto-resolved battle types, right beside the existing pool pre-filter, so an AI kage never selects a cast that `summon()` will refuse:

```ts
if (
  AutoBattleTypes.includes(nextBattle.battleType) &&
  action.effects.some((e) => "type" in e && e.type === "summon")
) {
  return false;
}
```

**Cost paid before the refusal** — fixed in 563d3c3. The cap is now checked in the router before `performBattleAction`, so a blocked cast spends nothing and surfaces a toast naming the reason. Pools are deducted inside `insertAction` (actions.ts:1110), well after entry, so the early return is genuinely free.

`summon()` keeps the predicate as a backstop for paths that reach `applyEffects` directly, but a human no longer gets there.

app/src/libs/combat/tags.ts
MathiasGruber: Done in 6653a314 — initiative-based placement, as you asked.

```ts
newAi.initiative = user.initiative;
const insertAt = usersState.findIndex((u) => u.initiative < newAi.initiative);
if (insertAt === -1) {
  usersState.push(newAi);
} else {
  usersState.splice(insertAt, 0, newAi);
}
```

The summon carries its summoner's initiative and goes in at the matching sorted position, so the descending-initiative invariant that `initiateBattle` establishes still holds after a mid-battle spawn. With nobody else on the summoner's initiative this lands the summon directly after them, so a piloted pair still acts back to back — the ergonomics that motivated the original splice, without the tempo buff against faster enemies.

Four tests replace the old placement test: initiative is inherited; the summon lands directly after its summoner when the initiative is unshared; the array stays sorted when a *faster* enemy exists (the case the old splice got wrong); and it appends correctly when the summoner is slowest. Suite 1069 pass / 0 fail.

MathiasGruber (approved):

PR #1386, opened 2026-07-10
fix: stop treating pinned-session tabs as signed out under Clerk multi-session

MathiasGruber:
# Pull Request

## Summary

Fixes the regression reported after #1374 merged: **signed-in users being told they are not logged in** (bounced to the logged-out landing page, "You need to be signed in to perform this action." toasts) **until they log out and log in again**, with the underlying trigger being Clerk randomly flipping the browser-global active account when multiple accounts are signed in.

## Why / Root Cause

#1374 correctly made each tab's **identity** and **tRPC auth** follow its pinned Clerk session. But "am I signed in?" decisions still read Clerk's **browser-global** active-session state (`useUser().isSignedIn`), which multi-account usage flips or drops **without user action** — focused tabs overwrite the shared `__session` cookie and active-context claim on every token refresh (verified in `clerk-js`: `Session._getToken` dispatches `TokenUpdate` for non-active sessions too, and `AuthCookieService.updateSessionCookie` lets any focused tab write the cookie and claim the active context), and signing out of / expiring the *other* account can leave the global state signed-out entirely.

So a tab whose pinned session was perfectly valid could be treated as signed out:

1. **`useRequiredUserData`** redirected to `/` on global `!isSignedIn` alone. On `/`, the index page (also gating on global `isSignedIn`) rendered the logged-out `<Welcome>` landing page — or, when profile data was cached, entered a `/` ↔ `/profile` redirect loop.
2. **`useGlobalOnMutateProtect`** threw "You need to be signed in to perform this action." on global `!isSignedIn`, so every game action showed a sign-in toast in a working tab.
3. **`profile.getUser` had `retry: false`**, so the two *expected-transient* multi-session failures introduced by #1374 — the server's fail-closed `UNAUTHORIZED` (per-tab token blip) and the `FORBIDDEN` viewer/session mismatch guard — settled into a terminal error state with no data, which the hooks above read as "signed out" (`useRequiredUserData` boots to `/`; the index then pushes `/500`). Only re-login recovered.
4. **`decidePinnedSession` silently adopted the other signed-in account when the pinned session died** — the one remaining "randomly swapped me to my alt" path after #1374.

### 1. Pinned session is the signed-in authority for the tab

`useRequiredUserData`, `useGlobalOnMutateProtect`, the index page's `<Welcome>` gate/referral effect, and the anonymous-visitor tracker now treat the tab as signed out only when **neither** the pinned session **nor** the browser-global session is signed in. A global-state flip or drop can no longer log a pinned tab "out"; a genuine sign-out (both signals gone) still redirects as before.

### 2. Transient multi-session auth failures retry instead of settling

New `isTransientMultiSessionAuthError(code, message)` in `authHeaders.ts` (the isomorphic module that already owns `VIEWER_SESSION_MISMATCH_MESSAGE`). `profile.getUser` retries these up to 3 times (they self-heal on the next per-tab-token request by design); all other errors keep failing fast as before.

### 3. A dead pin never silently switches accounts

The pinned **user id** is now stored alongside the pinned session id in `sessionStorage` (backfilled for existing pins). When the pinned session is no longer signed in and the sessions list is populated:

- **Same account active** (matched via stored user id, e.g. fresh session after expiry) → silently re-adopt, as before.
- **Different or unknown account active** → new `release` decision: clear the pin (remounts the subtree, dropping the old account's caches) and route to `/login` so the user picks an account **explicitly**. A `releasedRef` guard blocks adoption between the release and the `/login` navigation completing, so a mid-navigation Clerk sessions refresh cannot re-adopt the alt.
- **On the sign-in flow** a dead pin is `clear`ed (previously kept) so the account chosen there adopts cleanly instead of reading as a cross-account move and bouncing back to `/login`.

First-load adoption of a fresh tab is unchanged.

## What Was NOT Changed (known follow-ups)

- Clerk's `<UserButton>` (sidebar) still displays the browser-global active account, which can differ from the tab's pinned account; its "switch account" action changes only the global active session, which pinned tabs deliberately ignore. An explicit in-tab account switcher wired to the pin is a follow-up.
- Other protected queries do not get the transient-auth retry; `profile.getUser` is the identity-critical one that gates the whole app.

## Testing

- Updated `sessionPin.test.ts` for the new `decidePinnedSession` inputs/decisions: release on dead pin + different account, release on unknown account, silent same-account re-adopt, dead-pin clear on the auth route, dead pin with no adoptable active session, plus the pinned-user-id storage round-trip. All existing cross-tab/auth-route protections re-asserted.
- New `authHeaders.test.ts` cases for `isTransientMultiSessionAuthError`.
- `make test` (398 tests), `make typecheck`, and biome lint on changed files all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multi-session sign-in handling across tabs and sessions.
  * Prevented accidental account switching when a pinned session expires; affected users are routed through sign-in.
  * Reduced transient authentication errors during session refreshes.
  * Improved redirects and loading states when session information is still being restored.
  * Preserved authenticated access for valid pinned sessions even when global session status changes.
* **Tests**
  * Expanded coverage for pinned-session recovery, account switching, authentication errors, and storage behavior.

### DIFF
--- a/app/src/app/_trpc/SessionPinProvider.tsx
+++ b/app/src/app/_trpc/SessionPinProvider.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useSession, useSessionList } from "@clerk/nextjs";
-import { usePathname } from "next/navigation";
+import { usePathname, useRouter } from "next/navigation";
 import {
   createContext,
   Fragment,
@@ -13,12 +13,15 @@ import {
   useState,
 } from "react";
 import {
+  clearPinnedSessionStorage,
   decidePinnedSession,
   isAuthRoute,
   pinChangeRequiresRemount,
   readPinnedSessionId,
+  readPinnedUserId,
   resolvePinnedSession,
   writePinnedSessionId,
+  writePinnedUserId,
 } from "./sessionPin";
 
 interface SessionPinContextValue {
@@ -47,10 +50,13 @@ const SessionPinContext = createContext<SessionPinContextValue | null>(null);
  * It also replaces Clerk's `MultisessionAppSupport`: that component remounts the
  * app keyed on the ACTIVE session, which would re-flip the tab on every global
  * active-session change. Here we remount the subtree (fresh tRPC client + React
- * Query cache) only on a cross-account pin change — i.e. when the pinned account is
- * no longer signed in and the tab adopts a different signed-in account (see
- * `pinChangeRequiresRemount`). A background active-session change does not remount,
- * and neither does the first-load adoption of the tab's own account, so the normal
+ * Query cache) only when the pin moves off a signed-in account (see
+ * `pinChangeRequiresRemount`) — re-pinning the same account's fresh session, or
+ * clearing a dead pin before the tab heads to the sign-in flow. The tab NEVER
+ * silently adopts a different signed-in account: when the pinned session dies while
+ * another account remains, the pin is released and the user picks an account on
+ * /login explicitly. A background active-session change does not remount, and
+ * neither does the first-load adoption of the tab's own account, so the normal
  * signed-in first load and post-sign-in redirect do not pay a tree-wide remount.
  */
 export function SessionPinProvider(props: { children: React.ReactNode }) {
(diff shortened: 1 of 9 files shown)


MathiasGruber:
Re: the two Greptile advisory notes, for the record:

1. **`releasedRef` depends on `SessionPinProvider` living in the root layout and never remounting** — correct, and that placement is also what the whole pin system (and the once-created tRPC client reading it) already relies on. The remount key deliberately sits *below* the provider (`<Fragment key={remountKey}>`), so pin-driven remounts never destroy the ref. If the provider is ever moved, the pin architecture as a whole needs revisiting, not just this ref.

2. **Every `UNAUTHORIZED` on `profile.getUser` is treated as transient** — the retry is bounded (3 attempts, ~7s worst case) and the query only runs while the tab's pinned session still resolves client-side. On a genuine revocation Clerk drops the session from the sessions list within its normal sync, which disables the query / releases the pin independently of the retry loop, so the terminal state is delayed by seconds at most — a deliberate trade against the regression this PR fixes (one transient blip logging users out).

PR #1389, opened 2026-07-11
Filter by bracket instead of level

Phrosfire:
# Pull Request

**Changes:**
- AS and Genin are bracket 0
- The level filter is now a bracket filter
- Setting the filter all the way to the left disables the filter and displays everyone

## License

## Summary by CodeRabbit

- **New Features**
  - PvP protection/attack eligibility and combat checks are now rank-aware and use XP brackets, with reserved bracket 0 for Academy students and Genin.
  - Scouting and nearby-user filtering now use minimum XP bracket selection (with an off option) and bracket-aware labels.
  - Live map/sector updates now include each user’s XP and rank so bracket context stays consistent.

- **Bug Fixes**
  - Restricted-rank players are correctly shown as protected or unable to attack.
  - Movement updates no longer erase existing XP/rank data when it’s missing.

- **Documentation**
  - Updated bracket definitions and tooltip wording for rank restrictions and reserved bracket 0.


app/src/layout/Sector.tsx
```diff
@@ -319,7 +328,7 @@ const Sector: React.FC<SectorProps> = (props) => {
             }
           }
         } else {
-          // New user enters
+          // New user enters — do not invent experience; unknown XP fails open in bracket filters
```
MathiasGruber: Comment is incorrect: it says "unknown XP fails open", but `passesBracketFilter` (in `libs/profile.ts`) fails **closed** when experience is nullish and a filter is active (`if (user.experience == null) return false;`), and its own doc string states "Unknown experience fails closed when a filter is active." A new user whose pusher payload arrives without an `experience` field will be **hidden** by the bracket filter, not shown. Please align this comment with the actual behavior.

Phrosfire: Corrected this comment

app/src/layout/Sector.tsx
```diff
@@ -1259,26 +1259,25 @@ const SorroundingUsers: React.FC<SorroundingUsersProps> = (props) => {
...
 
   // Filter users
   const users = props.users
     .filter((user) => user.userId !== userData.userId)
     .filter((user) => user.status === "AWAKE")
-    .filter((user) => user.level >= watchedLevel)
     .filter(
-      (user) =>
-        !props.hideStudentGeninScout || !RANKS_RESTRICTED_FROM_PVP.includes(user.rank),
+      (user) => !isBracketFilterActive || passesBracketFilter(user, watchedBracket),
```
MathiasGruber: The `!isBracketFilterActive ||` short-circuit here is redundant with `passesBracketFilter`, which already returns `true` when `filterBracket < 0` (i.e. when the filter is off). This can be simplified to just `.filter((user) => passesBracketFilter(user, watchedBracket))`. The `isBracketFilterActive` boolean is still useful for the different empty-state message just below, but it isn't doing anything in the filter predicate itself.

MathiasGruber (approved):
LGTM. Prior review feedback all resolved: rank-aware bracket calculation, fail-closed on missing XP, redundant filter short-circuit removed, comment corrected, and the pusher contract now guarantees experience/rank on all map updates. Sector filter, PublicUser eligibility badge, and combat guards all consistent.

app/src/layout/SliderField.tsx
```diff
@@ -34,13 +35,18 @@ const SliderField: React.FC<SliderFieldProps> = (props) => {
     }
   };
 
+  const selectedLabel =
+    typeof props.watchedValue === "number"
+      ? props.formatWatchedValue
+        ? props.formatWatchedValue(props.watchedValue)
+        : `Selected: ${props.watchedValue.toFixed(2)} ${props.watchedTotal ? `/ ${props.watchedTotal.toFixed(2)}` : ""} ${props.unit ?? ""}`
+      : "";
```
MathiasGruber, line 42: This change unintentionally alters behavior for existing `SliderField` consumers. The old check `props.watchedValue ? ...` treated `0` as falsy and hid the label; the new check `typeof props.watchedValue === "number"` is true for `0`, so consumers with `min={0}, default={0}` (e.g., the three "days" sliders in `app/src/app/event/page.tsx`) will now render `Selected: 0.00  days` on first load instead of hiding the selection line.

If the new bracket slider needs a label at `-1`, the intent is already captured by `formatWatchedValue`. Consider keeping the visibility rule as `props.watchedValue` (or explicitly `Number.isFinite(props.watchedValue) && props.watchedValue !== 0` when `formatWatchedValue` is not passed) so unrelated call sites keep their existing UX. As it stands the guard is also effectively a no-op — the prop is typed as `number` so `typeof === "number"` is essentially always true (aside from NaN, which would then render `Selected: NaN`).

app/src/libs/profile.ts
```diff
@@ -55,17 +56,34 @@ export const calcLevel = (experience: number) => {
...
+ * use filterBracket < 0 to disable filtering entirely.
+ * Unknown experience fails closed when a filter is active.
+ */
+export const passesBracketFilter = (
+  user: { experience?: number | null; rank?: UserRank },
+  filterBracket: number,
+): boolean => {
+  if (filterBracket < 0) return true;
+  if (user.experience == null) return false;
+  return getExpBracket(user.experience, user.rank) === filterBracket;
```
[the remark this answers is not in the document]
Phrosfire: This is intended

PR #1390, opened 2026-07-11
Fix skillpoints

Phrosfire:
# Pull Request

When we changed Chunin to start at level 30, we didn't adjust when skill points start which causes users to not have enough skillpoints.  This changes uses the appropriate level ranges and contains a migration to update all users to the correct number of skillpoints.

We will do a skill tree reset to put everyone to fresh when this update is merged.

## License

## Summary by CodeRabbit

* **New Features**
  * Skill points from leveling are now awarded between levels 31 and 50 for eligible ranks.
* **Bug Fixes**
  * Players can now level up when their experience exactly meets the required threshold.
  * Existing skill-point balances have been recalculated to reflect the updated leveling range.
* **Data Updates**
  * Added the database migration needed to apply the skill-point balance changes.


app/drizzle/migrations/0019_aberrant_morgan_stark.sql
```diff
@@ -0,0 +1,16 @@
...
+UPDATE `UserData`
+SET `skillPoints` = LEAST(
+  GREATEST(
+    `skillPoints` + (
+      LEAST(GREATEST(`level` - 30, 0), 20) - LEAST(GREATEST(`level` - 20, 0), 20)
+    ),
+    0
+  ),
+  100
+)
```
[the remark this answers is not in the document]
Phrosfire: We will be doing a mass skill tree reset after this goes live.  This is not a concern

app/src/server/api/routers/profile.ts
```diff
@@ -404,10 +404,10 @@ export const profileRouter = createTRPCRouter({
       const { trackers } = getNewTrackers(user, [
         { task: "user_level", value: newLevel },
       ]);
-      // Calculate skill points reward for chunin+ ranks - levels 21-40 give 1 skill point each
+      // Calculate skill points reward for chunin+ ranks - levels 31-50 give 1 skill point each
```
MathiasGruber: Not directly this line, but noting here since the target file is outside the diff: preview testing of this PR uncovered a client/server threshold mismatch on level-up. Server (line 396 of this file) allows leveling when `expRequired <= 0`, but the client CTA in `app/src/layout/LevelUpBtn.tsx:65` uses strict `userData.experience > expRequired`. A user sitting at exactly `calcLevelRequirements(level)` XP satisfies the server rule but the button stays hidden. Pre-existing bug; fix by changing the client check to `>=`.

MathiasGruber (commented):
Preview testing surfaced a pre-existing threshold bug worth fixing in a follow-up.

app/src/server/api/routers/profile.ts
```diff
@@ -404,10 +404,10 @@ export const profileRouter = createTRPCRouter({
...
       const isChunin = UserRolesWithSkillTreeAccess.includes(user.rank);
       // Calculate how many skillpoints they should have from leveling (max 20)
-      // Chunin+ ranks get 1 skillpoint per level from levels 21-40
+      // Chunin+ ranks get 1 skillpoint per level from levels 31-50
       const expectedSkillPointsFromLeveling =
```
MathiasGruber: Nit: `expectedSkillPointsFromLeveling` is now effectively dead code. `Math.min(newLevel - SKILL_POINT_MIN_LEVEL + 1, MAX_SKILL_POINTS_FROM_LEVELING)` is already clamped to `MAX_SKILL_POINTS_FROM_LEVELING`, so the `expectedSkillPointsFromLeveling <= MAX_SKILL_POINTS_FROM_LEVELING` check on line 423 is always true, and the variable isn't referenced anywhere else. The block simplifies to:

```ts
const skillPointsGain =
  isChunin &&
  newLevel >= SKILL_POINT_MIN_LEVEL &&
  newLevel <= SKILL_POINT_MAX_LEVEL
    ? 1
    : 0;
```

Behavior is preserved — level-up only fires once per level, and the `LEAST(..., MAX_SKILL_POINTS)` in the SQL update already caps the running total.

MathiasGruber (commented):
Skill point range change looks correct; migration recalibrates existing users' totals sensibly (delta between old/new leveling formulas, floored at 0 and capped at MAX_SKILL_POINTS). Rank list matches UserRolesWithSkillTreeAccess. One readability nit below.

app/drizzle/migrations/0020_massive_the_call.sql
```diff
@@ -0,0 +1,12 @@
+-- Recalibrate skill points after moving leveling rewards from levels 21-40 to 31-50.
+-- Preserves non-leveling skill points (e.g. quests) by applying the delta between formulas:
+--   old: LEAST(GREATEST(level - 20, 0), 20)
+--   new: LEAST(GREATEST(level - 30, 0), 20)
+-- Intentionally allows negative skillPoints so users with spent points keep the debt
+-- until the skill-tree reset refunds them to the correct final balance.
```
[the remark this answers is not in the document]
Phrosfire: UserData.skillPoints is total earned, not unspent. Purchases never decrement it (available = skillPoints - spent), and reset only deletes userSkill rows.

app/drizzle/constants.ts
```diff
@@ -1353,8 +1353,8 @@ export const WAR_DAILY_HEALTH_DRAIN = 400; // War health drained from both sides
 export const WAR_MAX_DURATION_DAYS = 14; // Maximum war duration before auto-resolution
 
 // Skill point leveling constants
-export const SKILL_POINT_MIN_LEVEL = 21; // Minimum level to start gaining skill points from leveling
-export const SKILL_POINT_MAX_LEVEL = 40; // Maximum level to gain skill points from leveling
+export const SKILL_POINT_MIN_LEVEL = 31; // Minimum level to start gaining skill points from leveling
+export const SKILL_POINT_MAX_LEVEL = 50; // Maximum level to gain skill points from leveling
```
MathiasGruber: `MAX_SKILL_POINTS_FROM_LEVELING` (line 1358) is no longer referenced anywhere after this PR removed its usage in `levelUp` and the corresponding import in `profile.ts`. Consider deleting it, or use it to derive the range (e.g. `SKILL_POINT_MAX_LEVEL = SKILL_POINT_MIN_LEVEL + MAX_SKILL_POINTS_FROM_LEVELING - 1`) so the three constants stay in sync.

MathiasGruber:

**Conflicts resolved:** the only conflicts were in the migration metadata, caused by `main` landing its own `0019` migration (`0019_even_vision`) while this PR also added a `0019`.

Resolution:
- Kept `main`'s `0019_even_vision` migration, its snapshot, and its journal entry unchanged.
- Deleted this PR's `0019_aberrant_morgan_stark.sql` and regenerated it as `0020_massive_the_call.sql` via `make emptymigration` (it is a custom data migration with no schema change), re-adding the skill point recalibration SQL verbatim.
- Snapshot chain verified: `0020.prevId` matches `0019.id`, and `0020` contains no schema drift relative to `0019`.

No application code changes were needed — `constants.ts`, `LevelUpBtn.tsx`, and `profile.ts` rebased cleanly and the resulting diff is identical to before.

Verification on the rebased branch: `make typecheck` passes, `make lint` passes, `make test` passes (445/445).

app/drizzle/migrations/0020_massive_the_call.sql
```diff
@@ -0,0 +1,12 @@
...
+UPDATE `UserData`
+SET `skillPoints` = `skillPoints` + (
+  LEAST(GREATEST(`level` - 30, 0), 20) - LEAST(GREATEST(`level` - 20, 0), 20)
+)
+WHERE `rank` IN ('CHUNIN', 'JONIN', 'ELITE JONIN', 'ELDER')
+  AND `isAi` = 0;
```
MathiasGruber: Consider narrowing the row set with `AND `level` BETWEEN 21 AND 50`. Both `rank` and `level` are indexed on `UserData`, and the delta collapses to 0 for anyone at level ≤ 20 or ≥ 51 (both LEAST/GREATEST branches clamp to the same value). Skipping the no-op rewrites for the (likely large) population of Chunin+ users at level 51-100 cuts binlog/replication traffic without changing the outcome.

Phrosfire: We can't narrow this migration as much as asked for.  We have affected users that are now well beyond level 51.  I did adjust it though to not hit users below level 20.  This will also fix users who still had skill points from before we increased the genin level cap.

app/drizzle/migrations/0025_hot_ken_ellis.sql
```diff
@@ -0,0 +1,15 @@
...
+-- until the skill-tree reset refunds them to the correct final balance.
+-- Narrowed to levels 21-50: outside that range both formulas clamp to the same value
+-- so the delta is 0 and rewriting those rows would only add binlog/replication traffic.
+UPDATE `UserData`
+SET `skillPoints` = `skillPoints` + (
+  LEAST(GREATEST(`level` - 30, 0), 20) - LEAST(GREATEST(`level` - 20, 0), 20)
+)
+WHERE `rank` IN ('CHUNIN', 'JONIN', 'ELITE JONIN', 'ELDER')
+  AND `isAi` = 0
+  AND `level` >= 20;
```
MathiasGruber: The narrowing this migration's own comment claims (lines 7-8, "Narrowed to levels 21-50") is not actually implemented by this predicate. `level >= 20` has no upper bound, so it matches every chunin+ human up to the rank level cap of 100 (`USER_CAPS` in `app/drizzle/constants.ts` gives CHUNIN/JONIN/ELITE JONIN/ELDER `LVL_CAP: 100`).

Working the delta `LEAST(GREATEST(level-30,0),20) - LEAST(GREATEST(level-20,0),20)` through the level ranges:

| level | new term | old term | delta |
|---|---|---|---|
| <= 20 | 0 | 0 | **0** |
| 21-30 | 0 | level-20 | -(level-20) |
| 31-40 | level-30 | level-20 | -10 |
| 41-49 | level-30 | 20 (clamped) | level-50 (-9..-1) |
| 50 | 20 (clamped) | 20 (clamped) | **0** |
| >= 51 | 20 (clamped) | 20 (clamped) | **0** |

So the only rows with a non-zero delta are **levels 21-49**. Both endpoints currently included are provable no-ops: level 20 exactly (the lower bound that was just added), and the entire level 50-100 band, which on a mature server is likely the bulk of the chunin+ population.

Why this costs something on a large `UserData`:

- MySQL skips the write when the new record compares equal to the old, so the no-op rows mostly avoid binlog traffic - but they are still **read and X-locked** by the UPDATE. Every scanned row in the range holds a row lock for the duration of the statement, so the migration takes locks across most of the chunin+ population of a hot, very wide table instead of a narrow slice.
- `UserData` has only single-column indexes on these predicates (`UserData_level_idx`, `UserData_rank_idx`, `UserData_isAi_idx` in `app/drizzle/schema.ts`) - no composite covering `(rank, isAi, level)`. With an unbounded `level >= 20` the range is unselective enough that the optimizer will likely abandon `UserData_level_idx` and full-scan `UserData`, reading every STUDENT/GENIN/AI row and every wide column (including the JSON blobs) along the way. Bounding the range to 21-49 makes the level index range genuinely selective and keeps the plan on the index.

Suggested fix - make the predicate match the stated intent:

```sql
  AND `level` BETWEEN 21 AND 49;
```

This is the same skip-the-no-ops shape already used by `0022_water_walkability.sql` (`AND \`defaultWalkCost\` <> 1`). The comment on lines 7-8 should be updated to say 21-49 as well, since 50 is a no-op too.

app/drizzle/migrations/0025_hot_ken_ellis.sql
```diff
@@ -0,0 +1,15 @@
...
+-- Preserves non-leveling skill points (e.g. quests) by applying the delta between formulas:
+--   old: LEAST(GREATEST(level - 20, 0), 20)
+--   new: LEAST(GREATEST(level - 30, 0), 20)
+-- Intentionally allows negative skillPoints so users with spent points keep the debt
+-- until the skill-tree reset refunds them to the correct final balance.
+-- Narrowed to levels 21-50: outside that range both formulas clamp to the same value
+-- so the delta is 0 and rewriting those rows would only add binlog/replication traffic.
+UPDATE `UserData`
+SET `skillPoints` = `skillPoints` + (
+  LEAST(GREATEST(`level` - 30, 0), 20) - LEAST(GREATEST(`level` - 20, 0), 20)
```
MathiasGruber: The `old` term assumes every Chunin+ user actually earned one point for each of levels 21..min(level, 40). That assumption does not hold, because `USER_CAPS.GENIN.LVL_CAP` is `30` (`app/drizzle/constants.ts:888`) while the pre-PR grant in `levelUp` was gated on `UserRolesWithSkillTreeAccess.includes(user.rank)` at the moment of leveling. A player who levelled 21..30 while still GENIN and only took the Chunin exam afterwards received **zero** leveling skill points for that span, yet this migration still debits them up to 10.

Concrete cases:
- Chunin at level 30 who promoted at 30: earned `Q` (quest points only). Delta is `0 - 10 = -10`, so `skillPoints` becomes `Q - 10` — the correct new-rule value is `Q`. For `Q < 10` this lands negative, and `availableSkillPoints = user.skillPoints - totalUsedSkillPoints` (`app/src/server/api/routers/skillTree.ts:147`) then blocks purchases the player legitimately earned from quests.
- Chunin at level 45 who promoted at 30: earned `Q + 10` (levels 31-40 under the old 21-40 window). Delta is `15 - 20 = -5`, giving `Q + 5`, but the new 31-50 rule should give `Q + 15`.

The planned mass skill-tree reset does not repair this: `resetSkillPoints` only deletes `userSkill` rows, it never recomputes `UserData.skillPoints`. Since `skillPoints` is the total-earned column, a wrong total stays wrong after the reset.

Because the actual historical grant is not reconstructible from `level` alone for this cohort, consider setting the leveling component absolutely rather than applying a delta — e.g. recompute `skillPoints` as `quest_component + LEAST(GREATEST(level - 30, 0), 20)` if the quest component is derivable, or at minimum floor the result so no account is pushed below its quest-earned balance.

Related: lines 5-6 state the negative balances are intentional so "users with spent points keep the debt until the skill-tree reset refunds them." That rationale was refuted on this PR — `skillPoints` is total earned, purchases never decrement it, and the reset issues no refund. The comment should be updated so it does not justify negatives on a mechanism that does not exist.

app/drizzle/migrations/0025_hot_ken_ellis.sql
MathiasGruber: Fixed in e9be99984. The migration now separates historical cohorts. The pre-cap-change delta is bounded to levels 21-49; the post-cap-change pass is bounded to levels 41-100 because those users receive a real +1..+10 catch-up for levels 41-50, including players now above 50. Every selected row now has a non-zero correction, and regression tests cover the boundaries.

app/drizzle/migrations/0025_hot_ken_ellis.sql
MathiasGruber: Fixed in e9be99984. The migration now uses Chunin exam completion relative to the 2026-04-22 Genin-cap change to distinguish users who could earn levels 21-40 from users who only earned levels 31-40. The later (and conservatively unclassified) cohort keeps its legitimate 31-40 points and receives only the missing 41-50 points. The incorrect negative-debt/reset-refund rationale was removed, and both cohort formulas have regression coverage.

MathiasGruber (approved):

PR #1394, opened 2026-07-17
Add disarm tag

Phrosfire:
# Pull Request

Adds a "Disarm" tag that prevents weapon usage while disarmed.

## License


app/src/libs/combat/actions.ts
```diff
@@ -213,6 +215,8 @@ export const availableUserActions = (
             if (NonActionItemTypes.includes(item.itemType)) return false;
             if (ui.equipped === "NONE") return false;
             if (item.itemType === "WEAPON") {
+              // Hide weapons when disarmed
+              if (isDisarmed) return false;
```
MathiasGruber, line 223: Hiding only item-type weapons leaves a semantic gap versus the effect message ("cannot use weapons"). Jutsu with `jutsuWeapon !== "NONE"` cannot normally be used without an equipped weapon (see `checkJutsuItems` in `libs/train.ts`), yet they remain selectable while disarmed because `availableUserActions` never re-checks the weapon requirement mid-battle. This mirrors how `isSummonPrevented` at lines 169-175 filters jutsu tagged with the summon effect — consider filtering `user.jutsus` where `getJutsu(battle, uj.jutsuId).jutsuWeapon !== "NONE"` when `isDisarmed` is set, otherwise a disarmed user can still cast weapon-required techniques like sword-slash jutsu.

app/src/libs/combat/util.ts
```diff
@@ -888,6 +888,21 @@ export const isUserSummonPrevented = (
...
+) => {
+  return userEffects?.some(
+    (e) =>
+      e.type === "disarm" &&
+      e.targetId === userId &&
+      !e.castThisRound &&
+      "rounds" in e &&
+      e.rounds &&
+      e.rounds > 0,
+  );
```
MathiasGruber: With this PR there are now four near-identical helpers in this file — `isUserStealthed`, `isUserSummonPrevented`, `isUserDisarmed`, and `isUserImmobilized` — each performing the same `some(e.type === X && e.targetId === userId && !e.castThisRound && "rounds" in e && e.rounds && e.rounds > 0)` scan. Worth extracting a shared helper (e.g. `hasActiveEffectOfType(userId, effects, type)`) that these four thin wrappers can delegate to, so any future refinement to the active-effect predicate only has to change in one place.

MathiasGruber (approved):

PR #1397, opened 2026-07-18
Apply new restrictions

Phrosfire:
# Pull Request

Don't allow banned/silenced users to create staff applications.  Only allow banned/silenced users to reply to their own support tickets.

## License

## Summary by CodeRabbit

* **Bug Fixes**
  * Hid the “Apply for Staff” confirmation/form unless the user is eligible, now accounting for staff application restrictions (not just pending status).
  * Blocked staff application submission when staff application restrictions apply.
  * Improved support conversation restrictions: banned or silenced users can only reply/compose on support tickets they originally created.
  * Updated restriction messaging and toasts to correctly distinguish the restricted support-ticket scenario in guidance and blocked-action feedback.

### DIFF
--- a/app/src/layout/Conversation.tsx
+++ b/app/src/layout/Conversation.tsx
@@ -23,7 +23,13 @@ import { useInfinitePagination } from "@/libs/pagination";
 import { showMutationToast } from "@/libs/toast";
 import { getNewReactions, processMentions } from "@/utils/chat";
 import { parseHtml } from "@/utils/parse";
-import { canPostAsAi, getMessagingRestriction } from "@/utils/permissions";
+import {
+  canPostAsAi,
+  canReplyToStaffConversationWhileRestricted,
+  getMessagingRestriction,
+  RESTRICTED_STAFF_CONVERSATION_MESSAGE,
+  RESTRICTED_SUPPORT_TICKET_REPLY_MESSAGE,
+} from "@/utils/permissions";
 import { stripBlockquotes } from "@/utils/sanitize";
 import { secondsFromNow } from "@/utils/time";
 import type { ArrayElement } from "@/utils/typeutils";
@@ -44,6 +50,8 @@ interface ConversationProps {
   topRightContent?: React.ReactNode;
   onCommentPosted?: () => void;
   onBack?: () => void;
+  /** When set, banned/silenced users may only compose if they created this ticket */
+  supportTicketCreatedByUserId?: string;
 }
 
 export const ConversationSkeleton: React.FC<ConversationProps> = (props) => {
@@ -114,8 +122,25 @@ const Conversation: React.FC<ConversationProps> = (props) => {
   });
   const allComments = comments?.pages.flatMap((page) => page.data);
   const conversation = comments?.pages[0]?.convo;
+  const canComposeDespiteRestriction =
+    !!userData &&
+    !!conversation?.isStaffAvailable &&
+    canReplyToStaffConversationWhileRestricted(
+      userData,
+      props.supportTicketCreatedByUserId,
+    );
   const canComposeMessage =
-    !!userData && (conversation?.isStaffAvailable || !composeRestriction);
+    !!userData && (!composeRestriction || canComposeDespiteRestriction);
+  // Single source of truth for toast + inline copy when compose is blocked.
+  const composeRestrictionMessage = !composeRestriction
+    ? null
+    : conversation?.isStaffAvailable &&
+        (userData?.isBanned || userData?.isSilenced) &&
+        !canComposeDespiteRestriction
+      ? props.supportTicketCreatedByUserId
+        ? RESTRICTED_SUPPORT_TICKET_REPLY_MESSAGE
+        : RESTRICTED_STAFF_CONVERSATION_MESSAGE
+      : composeRestriction;
   type ReturnedComment = ArrayElement<typeof allComments>;
 
   // Search functionality
@@ -588,10 +613,10 @@ const Conversation: React.FC<ConversationProps> = (props) => {
    * Submit comment
    */
   const handleSubmitComment = handleSubmit((data) => {
-    if (composeRestriction && !conversation?.isStaffAvailable) {
+    if (composeRestriction && !canComposeDespiteRestriction) {
       showMutationToast({
         success: false,
-        message: composeRestriction,
+        message: composeRestrictionMessage,
       });
       return;
     }
@@ -656,7 +681,7 @@ const Conversation: React.FC<ConversationProps> = (props) => {
             !canComposeMessage &&
             composeRestriction && (
               <p className="mb-2 text-center text-muted-foreground text-sm">
-                {composeRestriction}
+                {composeRestrictionMessage}
               </p>
             )}
           {conversation && !conversation.isLocked && canComposeMessage && (
--- a/app/src/server/api/routers/applications.ts
+++ b/app/src/server/api/routers/applications.ts
@@ -18,6 +18,7 @@ import {
   canDeleteStaffApplication,
   canViewAllApplications,
   getApprovalGroup,
+  getBanOrSilenceRestriction,
 } from "@/utils/permissions";
 import {
   createApplicationSchema,
@@ -40,6 +41,8 @@ export const applicationsRouter = createTRPCRouter({
         }),
       ]);
       // Guards
+      const applicationRestriction = getBanOrSilenceRestriction(user);
+      if (applicationRestriction) return errorResponse(applicationRestriction);
       if (!StaffApplicationTargetRoles.includes(input.targetRole)) {
         return errorResponse("Invalid target role");
       }
--- a/app/src/server/api/routers/comments.ts
+++ b/app/src/server/api/routers/comments.ts
@@ -13,6 +13,7 @@ import {
   conversationComment,
   forumPost,
   forumThread,
+  supportTicket,
   user2conversation,
   userBlackList,
   userData,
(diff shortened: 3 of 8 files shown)


app/src/layout/Conversation.tsx
```diff
@@ -656,7 +673,12 @@ const Conversation: React.FC<ConversationProps> = (props) => {
             !canComposeMessage &&
             composeRestriction && (
               <p className="mb-2 text-center text-muted-foreground text-sm">
-                {composeRestriction}
+                {conversation.isStaffAvailable &&
+                (userData.isBanned || userData.isSilenced) &&
+                props.supportTicketCreatedByUserId &&
+                props.supportTicketCreatedByUserId !== userData.userId
+                  ? "You can only reply to support tickets you created while banned or silenced"
+                  : composeRestriction}
```
MathiasGruber: When a banned/silenced user opens a staff-application conversation (rendered from `app/src/app/manual/staff/applications/[applicationId]/page.tsx`, which does not pass `supportTicketCreatedByUserId`), this ternary falls through to `composeRestriction` and shows the generic `You are banned` copy instead of the intended restriction guidance. The `props.supportTicketCreatedByUserId && ... !== userData.userId` guard means the specific message only appears for someone else`s support ticket, not for the staff-application case.

Suggested fix: drop the `supportTicketCreatedByUserId`-based conditions so the specific message shows for any staff-available conversation where a banned/silenced user cannot compose:

```tsx
{conversation.isStaffAvailable && (userData.isBanned || userData.isSilenced)
  ? "You can only reply to support tickets you created while banned or silenced"
  : composeRestriction}
```

`canComposeMessage` already gates whether the block is rendered at all, so this cannot regress the own-ticket case.

MathiasGruber: Banned/silenced users viewing a staff-application conversation still see the generic `You are banned` copy here instead of the clearer restriction guidance. The applications page (`app/src/app/manual/staff/applications/[applicationId]/page.tsx`) renders `<Conversation>` without a `supportTicketCreatedByUserId` prop (there is no ticket for an application), so `props.supportTicketCreatedByUserId` is falsy and this ternary always falls through to `composeRestriction`. Either broaden the condition to trigger whenever `conversation.isStaffAvailable && (userData.isBanned || userData.isSilenced)` (so any staff-only convo — support ticket or application — shows the clearer message), or thread an equivalent prop through the applications page. Keep the fallback consistent with the `handleSubmitComment` toast at ~L606-613.

app/src/layout/Conversation.tsx
```diff
@@ -588,10 +602,13 @@ const Conversation: React.FC<ConversationProps> = (props) => {
...
   const handleSubmitComment = handleSubmit((data) => {
-    if (composeRestriction && !conversation?.isStaffAvailable) {
+    if (composeRestriction && !canComposeDespiteRestriction) {
       showMutationToast({
         success: false,
-        message: composeRestriction,
+        message:
+          conversation?.isStaffAvailable && (userData?.isBanned || userData?.isSilenced)
+            ? "You can only reply to support tickets you created while banned or silenced"
+            : composeRestriction,
```
MathiasGruber: This toast condition (`conversation?.isStaffAvailable && (userData?.isBanned || userData?.isSilenced)`) is looser than the one used for the inline message on lines 676–681, which also requires `props.supportTicketCreatedByUserId && props.supportTicketCreatedByUserId !== userData.userId`. For a banned/silenced user in a staff conversation where no `supportTicketCreatedByUserId` prop is passed (e.g. a staff application conversation), the inline message will show `composeRestriction` ("You are banned") but this toast would say "You can only reply to support tickets you created" — misleading because no ticket is involved. Compute one `composeRestrictionMessage` derived from `canComposeDespiteRestriction`/props and reuse it in both places so they can never drift.

app/src/layout/Conversation.tsx
MathiasGruber: The string "You can only reply to support tickets you created while banned or silenced" is duplicated three times (here, in `handleSubmitComment` on line 610, and in `comments.ts` line 705). Extract it to a shared constant (e.g. exported from `@/utils/permissions` next to `canReplyToStaffConversationWhileRestricted`) so the wording stays in sync across UI and server.

app/src/layout/Conversation.tsx
```diff
@@ -656,7 +673,12 @@ const Conversation: React.FC<ConversationProps> = (props) => {
             !canComposeMessage &&
             composeRestriction && (
               <p className="mb-2 text-center text-muted-foreground text-sm">
-                {composeRestriction}
+                {conversation.isStaffAvailable &&
+                (userData.isBanned || userData.isSilenced) &&
+                props.supportTicketCreatedByUserId &&
+                props.supportTicketCreatedByUserId !== userData.userId
```
MathiasGruber: `props.supportTicketCreatedByUserId !== userData.userId` is effectively dead: if the user owns the ticket then `canComposeMessage` is true and this whole `!canComposeMessage && composeRestriction` branch is skipped, so this branch only ever runs when the user is not the creator. Either drop the redundant check or, better, drive this message off the same computed value as the toast (see other comment) to avoid two subtly-different conditions.

app/src/server/api/routers/applications.ts
```diff
@@ -40,6 +40,8 @@ export const applicationsRouter = createTRPCRouter({
         }),
       ]);
       // Guards
+      if (user.isBanned) return errorResponse("You are banned");
+      if (user.isSilenced) return errorResponse("You are silenced");
```
MathiasGruber: Instead of inlining `"You are banned"` / `"You are silenced"` here, reuse the existing pattern from `getMessagingRestriction` / `getConceptArtCreateRestriction` (e.g. add a `getStaffApplicationRestriction` helper in `@/utils/permissions.ts`). Keeps ban/silence wording and precedence in one place and matches how the other restricted actions in the codebase are gated.

app/src/server/api/routers/comments.ts
```diff
@@ -682,8 +685,25 @@ export const commentsRouter = createTRPCRouter({
...
+          where: eq(supportTicket.conversationId, convo.id),
+          columns: { createdByUserId: true },
+        });
+        if (
+          !canReplyToStaffConversationWhileRestricted(
+            user,
+            linkedSupportTicket?.createdByUserId,
+          )
+        ) {
+          return errorResponse(RESTRICTED_SUPPORT_TICKET_REPLY_MESSAGE);
```
MathiasGruber: Backend returns `RESTRICTED_SUPPORT_TICKET_REPLY_MESSAGE` ("You can only reply to support tickets you created…") for any `isStaffAvailable` conversation, but this branch also runs when `linkedSupportTicket` was not found (e.g., a banned/silenced applicant POSTing into a staff-application conversation, which has no linked `supportTicket` row). The frontend already distinguishes the two cases via `RESTRICTED_STAFF_CONVERSATION_MESSAGE` in `Conversation.tsx`, so the toast a banned user sees on the client ("You cannot participate in this conversation while banned or silenced") will disagree with the tRPC error toast surfaced from the server. Mirror the client's split here:

```ts
return errorResponse(
  linkedSupportTicket
    ? RESTRICTED_SUPPORT_TICKET_REPLY_MESSAGE
    : RESTRICTED_STAFF_CONVERSATION_MESSAGE,
);
```

app/src/utils/permissions.ts
```diff
@@ -420,6 +420,15 @@ export const getConceptArtCreateRestriction = (
...
 };
 
+/** Returns a restriction message if the user cannot create a staff application, otherwise null. */
+export const getStaffApplicationRestriction = (
+  user: Pick<UserData, "isBanned" | "isSilenced">,
+): string | null => {
+  if (user.isBanned) return "You are banned";
+  if (user.isSilenced) return "You are silenced";
+  return null;
+};
```
MathiasGruber, line 421: `getStaffApplicationRestriction` has a body byte-identical to `getConceptArtCreateRestriction` a few lines above (same signature, same two checks, same messages). Two side-by-side copies of the same rule will drift the first time "banned"/"silenced" phrasing or precedence changes. Collapse them into a single shared helper (e.g. `getBanOrSilenceRestriction(user)`) and have both the concept-art and staff-application call sites use it — the message strings already leak into UI copy, so keeping one source of truth is worth it.

app/src/utils/permissions.ts
```diff
@@ -430,6 +439,25 @@ export const getMessagingRestriction = (
...
+  "You can only reply to support tickets you created while banned or silenced";
+
+export const RESTRICTED_STAFF_CONVERSATION_MESSAGE =
+  "You cannot participate in this conversation while banned or silenced";
+
+export const canReplyToStaffConversationWhileRestricted = (
+  user: Pick<UserData, "userId" | "isBanned" | "isSilenced">,
+  supportTicketCreatedByUserId: string | null | undefined,
+): boolean => {
+  if (!user.isBanned && !user.isSilenced) return true;
```
MathiasGruber: `if (!user.isBanned && !user.isSilenced) return true;` is dead at both call sites. `comments.ts` L692-695 short-circuits on `!isBannedOrSilenced` before invoking this helper, and `Conversation.tsx` L127 gates the whole `canComposeDespiteRestriction` expression on `userData.isBanned || userData.isSilenced`. Given that, this helper is really "does the user own the ticket?" — the name and the extra early-return give the impression the helper is safe to call on any user, which invites future callers to skip the ban/silence guard and let level-gated (or any future) restricted users through. Either drop the early return so the helper enforces its precondition (`user.isBanned || user.isSilenced`), or rename to `didUserCreateSupportTicket` and remove the redundant caller guards.

app/src/layout/Conversation.tsx
```diff
@@ -588,10 +614,10 @@ const Conversation: React.FC<ConversationProps> = (props) => {
    * Submit comment
    */
   const handleSubmitComment = handleSubmit((data) => {
-    if (composeRestriction && !conversation?.isStaffAvailable) {
+    if (composeRestriction && !canComposeDespiteRestriction) {
       showMutationToast({
         success: false,
-        message: composeRestriction,
+        message: composeRestrictionMessage || composeRestriction,
```
MathiasGruber: `composeRestrictionMessage || composeRestriction` fallback is unreachable. Whenever this guard fires (`composeRestriction && !canComposeDespiteRestriction`), `composeRestrictionMessage` is computed on L136-144 with `composeRestriction` truthy, so it resolves to either one of the two `RESTRICTED_*` constants or `composeRestriction` itself — never `null`/empty. The `|| composeRestriction` therefore only obscures the fact that `composeRestrictionMessage` is the single source of truth here (which is the whole point of the L135 comment). Drop the fallback so the intent is unambiguous.

MathiasGruber (approved):
LGTM

PR #1398, opened 2026-07-18
Fix Issue #1330: Add consume tag

### LINKED ISSUE #1330: Consume
This is a new combat tag where a % of damage dealt is converted into a shield.

- [ ] Shield Becomes Active same round. 
- [ ] The shield will last based on the round on the consume tag.
- [ ] Consume works on Pierce and Damage Tags.

Phrosfire:
# Pull Request

This is a new combat tag where a % of damage dealt is converted into a shield.  Also removes shield from gcd

## License

## Summary by CodeRabbit

* **New Features**
  * Added **Consume** combat effect, converting a percentage of relevant outgoing **damage** (and **pierce**) into a temporary **self shield** for the attacker, with combat action-log messaging.
  * Updated effect handling to create the shield during resolution and to support configurable shield power and duration.
  * Expanded validation and objective/tag support so **Consume** is accepted as a valid battle/combat tag, including a rule that it requires the same action to also include **damage** or **pierce**.

* **Bug Fixes**
  * Ensure Consume shield values are merged correctly across combined outcomes and that the Consume effect does not persist after triggering.

* **Tests**
  * Added automated coverage for Consume behavior, shield creation, and cleanup.

### DIFF
--- a/app/drizzle/constants.ts
+++ b/app/drizzle/constants.ts
@@ -161,6 +161,7 @@ export const SHARED_COOLDOWN_TAGS = [
   "cleanseprevent",
   "clear",
   "clearprevent",
+  "consume",
   "debuffprevent",
   "drain",
   "increasepoolcost",
@@ -168,7 +169,6 @@ export const SHARED_COOLDOWN_TAGS = [
   "pierce",
   "poison",
   "seal",
-  "shield",
   "stun",
   "summon",
   "vamp",
@@ -1243,6 +1243,7 @@ export const POST_DAMAGE_MODIFIER_TYPES: string[] = [
   "lifesteal",
   "absorb",
   "vamp",
+  "consume",
 ];
 
 // Black market config
--- a/app/drizzle/constants.ts
+++ b/app/drizzle/constants.ts
@@ -1232,6 +1232,13 @@ export const RANKS_RESTRICTED_FROM_PVP = ["STUDENT", "GENIN"];
 export const STREAK_LEVEL_DIFF = 10;
 
 /**
+ * Upper bound on the HP a single shield effect may hold. Enforced by ShieldTag.health
+ * and used to clamp shields generated in combat (e.g. by the consume tag) so they can
+ * never exceed what the schema accepts.
+ */
+export const SHIELD_MAX_HEALTH = 100000;
+
+/**
  * Effect types that depend on post-mitigated damage values.
  * These must be processed AFTER damage modifiers and pierce have been applied.
  */
--- a/app/src/libs/combat/process.ts
+++ b/app/src/libs/combat/process.ts
@@ -12,7 +12,7 @@ import {
   POST_DAMAGE_MODIFIER_TYPES,
 } from "@/drizzle/constants";
 import type { ShieldTagType } from "@/validators/combat";
-import { VisualTag } from "@/validators/combat";
+import { ShieldTag, VisualTag } from "@/validators/combat";
 import {
   BARRIER_DAMAGE_TAG_TYPES,
   DAMAGE_LEECH_CAP_RATIO,
@@ -33,6 +33,7 @@ import {
   clear,
   clearPrevent,
   clone,
+  consume,
   copy,
   damageBarrier,
   damageUser,
@@ -197,6 +198,58 @@ const getVisualOrSound = (
     castThisRound: true,
     longitude,
     latitude,
+  };
+};
+
+/**
+ * Build the temporary shield effect granted by the consume tag.
+ *
+ * Parsing through ShieldTag keeps the shield contract in a single place: any field
+ * later added to ShieldTag (with a default) flows through here automatically instead
+ * of silently missing this call site. shieldHp is clamped to the schema's max health
+ * so a very large consumed hit can never throw during parsing mid-combat.
+ *
+ * isNew is intentionally false: shield()'s cast-round branch rolls a
+ * `Math.random() < power / 100` primary check to decide whether the shield lands, but
+ * here `power` is the raw shield HP (not a 0-100 probability). Marking the effect as
+ * not-new skips that roll so the consume shield is applied deterministically.
+ */
+const CONSUME_SHIELD_MAX_HP = 100000;
+const createConsumeShieldEffect = (
+  user: BattleUserState,
+  shieldHp: number,
+  rounds: number,
+  round: number,
+  targetId: string,
+): UserEffect => {
+  const cappedHp = Math.min(shieldHp, CONSUME_SHIELD_MAX_HP);
+  return {
+    ...ShieldTag.parse({
+      type: "shield",
+      description: "Temporary shield from consume",
+      rounds,
+      power: cappedHp,
+      powerPerLevel: 0,
+      direction: "offence",
+      calculation: "static",
+      health: cappedHp,
+      target: "SELF",
+    }),
+    id: nanoid(),
+    creatorId: user.userId,
+    targetId: user.userId,
+    level: 0,
+    // Regular shields are realized via realizeTag() as isNew, so shield() rolls its
+    // cast-round activation. Here power is already the raw HP, so keep isNew false to
+    // skip that roll and apply the shield deterministically.
+    isNew: false,
+    castThisRound: true,
+    createdRound: round,
+    longitude: user.longitude,
+    latitude: user.latitude,
+    barrierAbsorb: 0,
+    actionId: `consume-${user.userId}-${targetId}`,
+    targetType: "user",
   };
 };
 
@@ -693,6 +746,25 @@ export const applyEffects = (
               });
             }
           }
+          // Consume: convert a % of pre-shield damage into a temporary shield on the attacker.
+          // Not affected by heal modifiers and does not share the vamp/lifesteal leech budget.
+          const consumeShield = Math.floor(c.consumeShield ?? 0);
+          const consumeRounds = c.consumeRounds ?? 0;
+          if (consumeShield > 0 && consumeRounds > 0 && user.curHealth > 0) {
+            newUsersEffects.push(
+              createConsumeShieldEffect(
+                user,
+                consumeShield,
+                consumeRounds,
+                battle.round,
+                c.targetId,
+              ),
+            );
+            actionEffects.push({
+              txt: `${user.username} consumes ${consumeShield} damage as a shield for ${consumeRounds} rounds`,
+              color: "blue",
+            });
+          }
           // Reduce armor durability by 1 when hit (skip for battles that don't lose durability)
           if (!NO_DURABILITY_LOSS_COMBATS.includes(battle.battleType)) {
             const t = newUsersState.find((u) => u.userId === target.userId);
@@ -1128,6 +1200,8 @@ export const applySingleEffect = (
           info = lifesteal(effect, usersEffects, consequences, curTarget);
         } else if (effect.type === "vamp") {
           info = vamp(effect, usersEffects, consequences, curTarget);
+        } else if (effect.type === "consume") {
+          info = consume(effect, consequences);
         } else if (effect.type === "fleeprevent") {
           info = fleePrevent(effect, usersEffects, curTarget);
         } else if (effect.type === "healprevent") {
--- a/app/src/libs/combat/process.ts
+++ b/app/src/libs/combat/process.ts
@@ -10,6 +10,7 @@ import {
   OUT_OF_COMBAT_BASE_DAMAGE_INCREASE,
   OUT_OF_COMBAT_BASE_DAMAGE_REDUCTION,
   POST_DAMAGE_MODIFIER_TYPES,
+  SHIELD_MAX_HEALTH,
 } from "@/drizzle/constants";
 import type { ShieldTagType } from "@/validators/combat";
 import { ShieldTag, VisualTag } from "@/validators/combat";
@@ -202,53 +203,60 @@ const getVisualOrSound = (
 };
 
 /**
+ * Base shape of the temporary shield granted by the consume tag, parsed once at import.
+ *
+ * Going through ShieldTag keeps the shield contract in a single place: any field later
+ * added to ShieldTag (with a default) flows through here automatically instead of
+ * silently missing this call site. Everything the parse validates is static - rounds,
+ * power and health are overridden per cast below - so the parse never runs on the
+ * per-action combat path.
+ */
+const CONSUME_SHIELD_BASE = ShieldTag.parse({
+  type: "shield",
+  description: "Temporary shield from consume",
+  rounds: 1,
+  power: 1,
+  powerPerLevel: 0,
+  direction: "offence",
+  calculation: "static",
+  health: 1,
+  target: "SELF",
+});
+
+/**
  * Build the temporary shield effect granted by the consume tag.
  *
- * Parsing through ShieldTag keeps the shield contract in a single place: any field
- * later added to ShieldTag (with a default) flows through here automatically instead
- * of silently missing this call site. shieldHp is clamped to the schema's max health
- * so a very large consumed hit can never throw during parsing mid-combat.
+ * shieldHp and rounds are clamped to the bounds ShieldTag accepts, so a very large
+ * consumed hit can never produce a shield outside the schema's range.
  *
  * isNew is intentionally false: shield()'s cast-round branch rolls a
  * `Math.random() < power / 100` primary check to decide whether the shield lands, but
  * here `power` is the raw shield HP (not a 0-100 probability). Marking the effect as
  * not-new skips that roll so the consume shield is applied deterministically.
  */
-const CONSUME_SHIELD_MAX_HP = 100000;
 const createConsumeShieldEffect = (
   user: BattleUserState,
   shieldHp: number,
   rounds: number,
   round: number,
-  targetId: string,
 ): UserEffect => {
-  const cappedHp = Math.min(shieldHp, CONSUME_SHIELD_MAX_HP);
+  const cappedHp = Math.min(shieldHp, SHIELD_MAX_HEALTH);
   return {
-    ...ShieldTag.parse({
-      type: "shield",
-      description: "Temporary shield from consume",
-      rounds,
-      power: cappedHp,
-      powerPerLevel: 0,
-      direction: "offence",
-      calculation: "static",
-      health: cappedHp,
-      target: "SELF",
-    }),
+    ...CONSUME_SHIELD_BASE,
+    rounds: Math.min(rounds, 100),
+    power: cappedHp,
+    health: cappedHp,
     id: nanoid(),
     creatorId: user.userId,
     targetId: user.userId,
     level: 0,
-    // Regular shields are realized via realizeTag() as isNew, so shield() rolls its
-    // cast-round activation. Here power is already the raw HP, so keep isNew false to
-    // skip that roll and apply the shield deterministically.
     isNew: false,
     castThisRound: true,
     createdRound: round,
     longitude: user.longitude,
     latitude: user.latitude,
     barrierAbsorb: 0,
-    actionId: `consume-${user.userId}-${targetId}`,
+    actionId: `consume-${user.userId}-${round}`,
     targetType: "user",
   };
 };
@@ -592,6 +600,11 @@ export const applyEffects = (
     );
   });
 
+  // Consume shields are accumulated per attacker across all collapsed consequences and
+  // pushed once after the loop, so a multi-target action grants a single shield sized
+  // off the whole hit rather than one shield per damaged target.
+  const consumeShieldByUser = new Map<string, { hp: number; rounds: number }>();
+
   // Apply consequences to users
   Array.from(consequences.values())
     // Before collapsing consequences, we process each consequence indicidually
@@ -746,25 +759,6 @@ export const applyEffects = (
               });
             }
           }
-          // Consume: convert a % of pre-shield damage into a temporary shield on the attacker.
-          // Not affected by heal modifiers and does not share the vamp/lifesteal leech budget.
-          const consumeShield = Math.floor(c.consumeShield ?? 0);
-          const consumeRounds = c.consumeRounds ?? 0;
-          if (consumeShield > 0 && consumeRounds > 0 && user.curHealth > 0) {
-            newUsersEffects.push(
-              createConsumeShieldEffect(
-                user,
-                consumeShield,
-                consumeRounds,
-                battle.round,
-                c.targetId,
-              ),
-            );
-            actionEffects.push({
-              txt: `${user.username} consumes ${consumeShield} damage as a shield for ${consumeRounds} rounds`,
-              color: "blue",
-            });
-          }
           // Reduce armor durability by 1 when hit (skip for battles that don't lose durability)
           if (!NO_DURABILITY_LOSS_COMBATS.includes(battle.battleType)) {
             const t = newUsersState.find((u) => u.userId === target.userId);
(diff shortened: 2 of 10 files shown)


app/src/libs/combat/process.ts
```diff
@@ -693,6 +694,45 @@ export const applyEffects = (
...
+              direction: "offence",
+              calculation: "static",
+              health: consumeShield,
+              target: "SELF",
+              id: nanoid(),
+              creatorId: user.userId,
+              targetId: user.userId,
+              level: 0,
+              isNew: false,
+              castThisRound: false,
```
MathiasGruber: Setting `castThisRound: false` on a shield created mid-round makes it expire one game round earlier than a `ShieldTag` with the same `rounds` value.

`alignBattle` (util.ts:2429-2438) decrements `rounds` on the next round transition unless `castThisRound` is true. A shield placed via the shield tag runs through `applySingleEffect`, which sets `castThisRound = true` at line 1027, so a `rounds: 3` shield survives the first transition (rounds stays 3) and is active for 4 game rounds. This consume shield starts with `castThisRound: false`, so the very next transition decrements to 2, and a `consumeRounds: 3` shield is only active for 3 rounds.

The other place in this file that constructs an effect mid-processing (`getVisualOrSound` at line 198) uses `castThisRound: true`, as does `summon` in tags.ts:3072/3090. Recommend setting `castThisRound: true` here (or documenting that `consumeRounds` is intentionally one round shorter than `ShieldTag.rounds`).

app/src/libs/combat/tags.ts
```diff
@@ -1939,6 +1939,35 @@ export const vamp = (
...
+        c.userId === effect.creatorId &&
+        c.targetId !== effect.creatorId &&
+        typeof c.damage === "number" &&
+        c.damage > 0
+      ) {
+        c.consumeRatio = (c.consumeRatio ?? 0) + power / 100;
+        c.consumeRounds = Math.max(c.consumeRounds ?? 0, shieldRounds);
+      }
+    });
+  }
```
MathiasGruber: The `consequences.forEach` block here is a copy of the same block in `vamp` (lines 1927-1937) — same `isNew && castThisRound` gate, same predicate (`c.userId === effect.creatorId && c.targetId !== effect.creatorId && typeof c.damage === "number" && c.damage > 0`), only the assigned field differs. Consider extracting a small helper like:

```ts
const forEachOutgoingDamage = (
  effect: UserEffect,
  consequences: Map<string, Consequence>,
  fn: (c: Consequence) => void,
) => {
  if (!effect.isNew || !effect.castThisRound) return;
  consequences.forEach((c) => {
    if (
      c.userId === effect.creatorId &&
      c.targetId !== effect.creatorId &&
      typeof c.damage === "number" &&
      c.damage > 0
    ) {
      fn(c);
    }
  });
};
```

so `vamp` and `consume` (and any future damage-conversion tag) share one predicate.

app/src/libs/combat/tags.ts
```diff
@@ -1939,6 +1939,35 @@ export const vamp = (
...
+ * Convert a percentage of damage dealt into a temporary shield on the caster.
+ * Uses the same pre-shield damage basis as vamp, but is not affected by heal modifiers.
+ * Intentionally returns no cast-round announcement; process.ts logs when the shield is created.
+ */
+export const consume = (
+  effect: UserEffect,
+  consequences: Map<string, Consequence>,
+): ActionEffect | undefined => {
+  const { power } = getPower(effect);
+  const shieldRounds = effect.rounds ?? 1;
```
MathiasGruber: `effect.rounds ?? 1` — the schema prefaults `rounds` to 3 with a min of 1 (`ConsumeTag` in validators/combat.ts:512), so this fallback is unreachable in normal flow, but if it ever were hit it would silently create a 1-round shield instead of the intended 3-round default. Consider either dropping the fallback (`const shieldRounds = effect.rounds;` with a non-null assertion or narrowing) or matching the schema prefault. The current mismatch is a small readability trap for the next editor.

Phrosfire:
@coderabbitai review

Phrosfire:
@coderabbitai review

app/src/libs/combat/tags.ts
```diff
@@ -3283,6 +3294,28 @@ export const getEfficiencyRatio = (dmgEffect: UserEffect, effect: UserEffect) =>
...
+  consequences: Map<string, Consequence>,
+  fn: (c: Consequence) => void,
+) => {
+  if (!effect.isNew || !effect.castThisRound) return;
+  consequences.forEach((c) => {
+    if (
+      c.userId === effect.creatorId &&
+      c.targetId !== effect.creatorId &&
+      typeof c.damage === "number" &&
+      c.damage > 0
```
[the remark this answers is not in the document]
Phrosfire: @coderabbitai This is a false fail.  We were already comparing against c.damage before turning this into a helper function.  The determined value is unchanged

app/drizzle/constants.ts
```diff
@@ -167,6 +167,7 @@ export const SHARED_COOLDOWN_TAGS = [
   "stun",
   "summon",
   "vamp",
+  "consume",
```
MathiasGruber: `SHARED_COOLDOWN_TAGS` is alphabetically sorted; `consume` is appended at the end instead of being inserted between `clearprevent` and `debuffprevent`. Please place it in order to keep the ordering intact.

app/src/libs/combat/tags.ts
```diff
@@ -1924,21 +1924,32 @@ export const vamp = (
...
+  consequences: Map<string, Consequence>,
+): ActionEffect | undefined => {
+  const { power } = getPower(effect);
+  const shieldRounds = effect.rounds ?? 3;
+  forEachOutgoingDamage(effect, consequences, (c) => {
+    c.consumeRatio = (c.consumeRatio ?? 0) + power / 100;
+    c.consumeRounds = Math.max(c.consumeRounds ?? 0, shieldRounds);
+  });
+  // Only the generated shield should persist; zero the consume tag so it does not linger.
+  effect.rounds = 0;
```
MathiasGruber: `ConsumeTag` repurposes `rounds` to mean "shield duration" and then forcibly zeroes `effect.rounds` here so the tag itself does not linger. That dual meaning is confusing at read time. Consider a dedicated `shieldRounds` field on `ConsumeTag` (leaving `rounds` unset, like `VampTag`) - that removes the need for the manual `effect.rounds = 0` reset. Also the `effect.rounds ?? 3` fallback on line 1943 is dead code: the validator enforces `min(1).prefault(3)`, so `rounds` is always defined.

app/src/libs/combat/process.ts
```diff
@@ -693,6 +694,45 @@ export const applyEffects = (
...
+              level: 0,
+              isNew: false,
+              castThisRound: true,
+              createdRound: battle.round,
+              longitude: user.longitude,
+              latitude: user.latitude,
+              barrierAbsorb: 0,
+              actionId: `consume-${c.userId}-${c.targetId}`,
+              targetType: "user",
+            });
```
MathiasGruber: This 28-line inline `UserEffect` construction duplicates the shape of `ShieldTag`. Please extract to a small factory (e.g. `createConsumeShieldEffect(user, shieldHp, rounds, round)`) so the shield contract lives in one place - future required fields on shield effects would otherwise silently miss this call site. Also worth a brief comment on why `isNew: false` is chosen (it deliberately bypasses `shield()`'s cast-round `Math.random() < power/100` primary check, since `power` here is set to raw HP).

app/src/libs/combat/tags.ts
```diff
@@ -1924,21 +1924,30 @@ export const vamp = (
...
+ * Intentionally returns no cast-round announcement; process.ts logs when the shield is created.
+ */
+export const consume = (
+  effect: UserEffect,
+  consequences: Map<string, Consequence>,
+): ActionEffect | undefined => {
+  const { power } = getPower(effect);
+  const shieldRounds = "shieldRounds" in effect ? effect.shieldRounds : 3;
+  forEachOutgoingDamage(effect, consequences, (c) => {
+    c.consumeRatio = (c.consumeRatio ?? 0) + power / 100;
```
MathiasGruber: `consumeRatio` accumulates with no upper clamp, so the generated shield can exceed the damage that was actually dealt.

`getPower()` hard-caps a single percentage tag at 100 (`power = power > 100 ? 100 : power`), so the intent is clearly "at most 100% of the hit". But nothing stops an action from carrying two `consume` effects, and each one adds another `power / 100` here. Verified against this branch: an action with two `consume` tags at power 100 dealing 404 damage produces a single 808 HP shield (`consumeRatio` = 2.0) — a 200% conversion, with the log line reading `consumes 808 damage as a shield`.

This is also the one leech-style tag with no ceiling at all: vamp + lifesteal are jointly bounded by `DAMAGE_LEECH_CAP_RATIO` (60% of `preShieldDamage`) in `process.ts`, whereas consume shields are uncapped, add a fresh effect every round, and are created once per target on AOE (a 2-target AOE yields two full-ratio shields).

Suggest clamping the accumulated ratio, e.g. `c.consumeRatio = Math.min(1, (c.consumeRatio ?? 0) + power / 100);`

app/src/validators/combat.ts
```diff
@@ -501,6 +501,23 @@ export const VampTag = z.object({
...
+  type: z.literal("consume").prefault("consume"),
+  description: msg(
+    "Convert a percentage of damage dealt by this jutsu into a temporary shield on yourself. Not affected by heal modifiers.",
+  ),
+  calculation: z.enum(["percentage"]).prefault("percentage"),
+  // The tag itself is instant (never lingers); shieldRounds is the duration of the
+  // shield it generates. Keeping rounds locked to 0 avoids a manual reset while the
+  // dedicated shieldRounds field carries the shield duration.
+  rounds: z.coerce.number().int().min(0).max(0).prefault(0),
+  shieldRounds: z.coerce.number().int().min(1).max(100).prefault(3),
```
MathiasGruber, line 524: `shieldRounds` is never surfaced to players, and the field that *is* surfaced is actively misleading.

`ItemWithEffects.tsx` (the jutsu/item card) renders a hand-curated list of effect fields — `rounds`, `calculation`, `blocks`, `power`, `rank`, `aiHp`, `target`, `powerPerLevel`, `residualModifier`, `generalTypes`, `statTypes`, `elements`, the rewards, `direction`, `poolsAffected`. `shieldRounds` is not in that list, so the shield duration is invisible on the jutsu/item description.

Worse, that component guards the rounds row with `"rounds" in parsedEffect && parsedEffect.rounds !== undefined`. Because `ConsumeTag.rounds` is `prefault(0)` rather than optional, `0 !== undefined` is true and every consume jutsu will render **"Rounds: 0"** — telling the player the effect has no duration while the shield it grants actually lasts `shieldRounds` rounds.

Adding a `shieldRounds` row to the field list in `ItemWithEffects.tsx` (next to the existing rounds row) would fix both halves.

app/tests/libs/combat/consume.test.ts
```diff
@@ -0,0 +1,98 @@
...
+      defender.curHealth -
+      (newBattle.usersState.find((u) => u.userId === "defender")?.curHealth ?? 0);
+    expect(healthLost).toBeGreaterThan(0);
+
+    const shields = newBattle.usersEffects.filter(
+      (e) => e.type === "shield" && e.targetId === "attacker" && (e.power ?? 0) > 0,
+    );
+    expect(shields).toHaveLength(1);
+    expect(shields[0]?.power).toBeCloseTo(Math.floor(healthLost * 0.5), 0);
+    expect(shields[0]?.rounds).toBe(3);
```
MathiasGruber, line 107: Both tests only exercise the case where the target has no shield of its own and only one damage consequence exists, which leaves the two pieces of new logic most likely to regress completely uncovered:

1. **The pre-shield basis.** The point of the `collapseConsequences` addition (`val.consumeRatio * (val.preShieldDamage ?? val.damage ?? 0)`) is that the shield is sized off the *full* hit even when the defender's own shield absorbed it. Here `preShieldDamage === damage`, so the intended value and the fallback are indistinguishable and `healthLost * 0.5` would pass either way. A case with a pre-existing shield on the defender (post-shield `damage` collapses to 0, consume shield still ~50% of the raw hit) would pin this down.
2. **The merge branch.** `current.consumeShield += ...` / `current.consumeRounds = Math.max(...)` in `collapseConsequences` only runs when two consequences share a `targetId`; with a single damage effect the reducer always takes the `else` branch, so that code never executes in these tests.

Also worth covering: the effect is hand-built by `createConsumeShieldEffect` instead of going through `realizeTag`, so a missing or wrong field could produce a shield that sits in `usersEffects` but never absorbs anything — and both current tests would still pass. A test that advances the round and confirms the shield soaks incoming damage would catch that.

app/src/libs/combat/process.ts
```diff
@@ -200,6 +201,58 @@ const getVisualOrSound = (
...
+const createConsumeShieldEffect = (
+  user: BattleUserState,
+  shieldHp: number,
+  rounds: number,
+  round: number,
+  targetId: string,
+): UserEffect => {
+  const cappedHp = Math.min(shieldHp, CONSUME_SHIELD_MAX_HP);
+  return {
+    ...ShieldTag.parse({
```
MathiasGruber: `ShieldTag.parse()` runs here on every consume hit, i.e. inside the collapsed-consequence loop of `applyEffects`, which executes for every action of every battle. All the fields the parse actually validates are static except `rounds`, `power` and `health`, and those are already guarded at the call site (`consumeShield > 0` after `Math.floor`, `consumeRounds > 0`, `cappedHp` clamped to `CONSUME_SHIELD_MAX_HP`, `shieldRounds` schema-bounded to 1-100). Zod object parsing with `coerce`/`prefault` on ~17 fields is not free, and none of that work varies per call.

Hoist the parse to module scope so it happens once at import time and keep the schema-driven defaults:

```ts
const CONSUME_SHIELD_BASE = ShieldTag.parse({
  type: "shield",
  description: "Temporary shield from consume",
  rounds: 1,
  power: 1,
  powerPerLevel: 0,
  direction: "offence",
  calculation: "static",
  health: 1,
  target: "SELF",
});

// in createConsumeShieldEffect:
return {
  ...CONSUME_SHIELD_BASE,
  rounds: Math.min(rounds, 100),
  power: cappedHp,
  health: cappedHp,
  id: nanoid(),
  ...
};
```

That keeps the "any field later added to ShieldTag flows through here" property the comment is after, while removing the parse from the per-hit path.

app/src/libs/combat/process.ts
```diff
@@ -693,6 +746,25 @@ export const applyEffects = (
...
+          if (consumeShield > 0 && consumeRounds > 0 && user.curHealth > 0) {
+            newUsersEffects.push(
+              createConsumeShieldEffect(
+                user,
+                consumeShield,
+                consumeRounds,
+                battle.round,
+                c.targetId,
+              ),
+            );
```
MathiasGruber: This pushes one new `UserEffect` per collapsed consequence. `collapseConsequences` merges by `targetId` only, so a consume jutsu that hits N targets creates N separate shield effects on the same attacker in a single round — and `actionId` is keyed by `targetId`, so nothing downstream ever merges them.

Cost compounds: `battle.usersEffects` is serialized into the battle-state JSON on every action, `applyEffects` sorts and iterates it several times per round, and `targetShields` (line 602) re-filters the whole effect array once per consequence. With `shieldRounds` allowed up to 100 and a multi-target jutsu cast each round, the attacker can carry hundreds of live shield effects that all have identical `rounds`/`createdRound`.

Aggregate before pushing — the shields are interchangeable, so summing the HP into one effect is behaviourally equivalent and produces one effect and one log line per cast:

```ts
// accumulate while iterating consequences
const consumeByUser = new Map<string, { hp: number; rounds: number }>();
...
const prev = consumeByUser.get(user.userId);
consumeByUser.set(user.userId, {
  hp: (prev?.hp ?? 0) + consumeShield,
  rounds: Math.max(prev?.rounds ?? 0, consumeRounds),
});

// after the consequence loop
consumeByUser.forEach(({ hp, rounds }, userId) => { /* push one effect + one actionEffect */ });
```

That also drops the per-target `actionId` suffix, so a single `consume-${userId}-${round}` id is enough.

MathiasGruber (commented):
Performance pass on the consume tag. No DB access is introduced anywhere in the round-processing path, so the `performAction` rule is respected. Two hot-path notes below.

app/src/validators/combat.ts
```diff
@@ -501,6 +501,23 @@ export const VampTag = z.object({
...
+  ...PowerAttributes,
+  type: z.literal("consume").prefault("consume"),
+  description: msg(
+    "Convert a percentage of damage dealt by this jutsu into a temporary shield on yourself. Not affected by heal modifiers.",
+  ),
+  calculation: z.enum(["percentage"]).prefault("percentage"),
+  // The tag itself is instant (never lingers); shieldRounds is the duration of the
+  // shield it generates. Keeping rounds locked to 0 avoids a manual reset while the
+  // dedicated shieldRounds field carries the shield duration.
+  rounds: z.coerce.number().int().min(0).max(0).prefault(0),
```
MathiasGruber: `rounds: ...min(0).max(0)` makes the admin tag editor unrecoverable when switching tag types.

The type-switch effect in `layout/EditContent.tsx` (~L1546-1566) copies every key of the newly-parsed tag from the previous tag: `if (!['type','calculation','direction'].includes(key) && key in curTag) shownTag[key] = curTag[key]`. So switching *poison → consume* copies `rounds: 3` onto the consume tag. Verified:

```
ConsumeTag.safeParse({...consume, rounds: 3})
// => false, ["rounds: Too big: expected number to be <=0"]
```

The form is then permanently invalid (`form.formState.isValid` never becomes true, so the "automatically update the effects" effect never commits), and because `ignore.push("rounds")` hides the field for consume, the admin cannot see or fix it. The reverse direction breaks too: *consume → poison* copies `rounds: 0` into `PoisonTag`, which requires `min(1)`.

Suggested fix: keep the field permissive and normalize instead of rejecting, e.g. `rounds: z.coerce.number().int().prefault(0).transform(() => 0)`, or add `"rounds"` to the skip list in the type-switch copy loop in `EditContent.tsx`.

app/src/libs/combat/process.ts
```diff
@@ -200,6 +201,58 @@ const getVisualOrSound = (
...
+ * later added to ShieldTag (with a default) flows through here automatically instead
+ * of silently missing this call site. shieldHp is clamped to the schema's max health
+ * so a very large consumed hit can never throw during parsing mid-combat.
+ *
+ * isNew is intentionally false: shield()'s cast-round branch rolls a
+ * `Math.random() < power / 100` primary check to decide whether the shield lands, but
+ * here `power` is the raw shield HP (not a 0-100 probability). Marking the effect as
+ * not-new skips that roll so the consume shield is applied deterministically.
+ */
+const CONSUME_SHIELD_MAX_HP = 100000;
```
MathiasGruber: Two things here:

1. `100000` is a copy of `ShieldTag.health`'s `.max(100000)` in `validators/combat.ts:559`. If that schema bound is ever lowered, this clamp silently stops protecting and `ShieldTag.parse(...)` below throws inside `applyEffects` — i.e. the whole combat action 500s. Per the project rule, put the bound in `drizzle/constants.ts` and reference it from both the schema's `.max()` and this clamp so they cannot drift.

2. The JSDoc block directly above (L204-216) is written to describe `createConsumeShieldEffect`, but with this `const` inserted between them it now documents `CONSUME_SHIELD_MAX_HP`. Move the constant above the doc comment.

app/src/libs/combat/process.ts
MathiasGruber: This runs once per collapsed consequence, and `collapseConsequences` merges by `targetId` only — so a multi-target action creates one separate shield effect per damaged target rather than one shield sized off the whole hit. Confirmed with a two-target AOE (50% consume, identical hits): the attacker ends up with `2` shield effects of `202` HP each rather than a single `404` HP shield.

That diverges from the tag's own description ("a percentage of damage dealt **by this jutsu**"), scales shield HP linearly with target count with no cap relative to `maxHealth` (unlike vamp, which at least clamps to `maxHealth`), and adds N entries to `usersEffects` that each persist for `shieldRounds`.

Suggest accumulating `consumeShield`/`consumeRounds` per attacker across the collapsed consequences and pushing a single shield after the loop. `consume.test.ts` covers multiple hits on the *same* target but not the multi-target case — worth adding once the behaviour is settled.

app/src/validators/combat.ts
```diff
@@ -1177,6 +1196,16 @@ export const SuperRefineEffects = (effects: ZodAllTags[], ctx: z.RefinementCtx)
...
+    } else if (e.type === "consume") {
+      const hasDamageOrPierce = effects.some(
+        (x) => x.type === "damage" || x.type === "pierce",
+      );
+      if (!hasDamageOrPierce) {
+        addIssue(
+          ctx,
+          "ConsumeTag must be used together with a damage or pierce effect on the same action",
+        );
+      }
```
MathiasGruber: This branch is a verbatim copy of the `vamp` branch immediately above, differing only in the message text. Collapse them:

```ts
} else if (e.type === "vamp" || e.type === "consume") {
  const hasDamageOrPierce = effects.some(
    (x) => x.type === "damage" || x.type === "pierce",
  );
  if (!hasDamageOrPierce) {
    addIssue(
      ctx,
      `${e.type === "vamp" ? "VampTag" : "ConsumeTag"} must be used together with a damage or pierce effect on the same action`,
    );
  }
}
```

app/src/libs/combat/process.ts
MathiasGruber: Done in 51f47480e. The `ShieldTag.parse()` is now a module-scope `CONSUME_SHIELD_BASE`, parsed once at import; `createConsumeShieldEffect` spreads it and overrides `rounds`/`power`/`health` per cast, so nothing schema-driven runs on the per-action path.

app/src/libs/combat/process.ts
MathiasGruber: Done in 51f47480e. `applyEffects` now accumulates into a `consumeShieldByUser` map keyed by attacker (summing HP, `Math.max` on rounds) and pushes a single effect plus one log line after the consequence loop, so `actionId` is just `consume-${userId}-${round}`. Added a two-target regression test in `consume.test.ts` that asserts exactly one shield and one log entry, with both targets confirmed hit so the assertion cannot go vacuous.

One thing I deliberately left alone: the aggregated shield is still sized off total damage across all targets, with no clamp relative to `maxHealth`. After aggregation that matches the tag description ("a percentage of damage dealt by this jutsu"), and `SHIELD_MAX_HEALTH` is the hard ceiling — but if you want a `maxHealth` clamp like vamp has, say the word and it is a one-liner.

app/src/libs/combat/process.ts
MathiasGruber: Done in 51f47480e. Added `SHIELD_MAX_HEALTH` to `drizzle/constants.ts`, referenced from both `ShieldTag.health.max()` and the combat-side clamp so they cannot drift. The JSDoc split in two: one block documents the new module-scope `CONSUME_SHIELD_BASE`, the other sits directly above `createConsumeShieldEffect`.

app/src/libs/combat/process.ts
MathiasGruber: Done in 51f47480e — same fix as the thread above: accumulated per attacker and granted once after the loop, with a two-target test covering it.

app/src/validators/combat.ts
MathiasGruber: Done in 51f47480e. `rounds` now normalizes instead of rejecting: `z.coerce.number().int().min(0).max(100).overwrite(() => 0).prefault(0)`. `.overwrite()` keeps the schema introspectable as a `ZodNumber` (so `getInner` in `EditContent` still unwraps it), and the bounds now match `BaseAttributes.rounds`, so nothing another tag can hold gets rejected. Verified `ConsumeTag.safeParse({...consume, rounds: 3})` now succeeds with `rounds: 0`.

Also covered the reverse direction you flagged: the type-switch copy loop in `EditContent.tsx` skips `rounds` when switching into or out of consume, so `consume -> poison` keeps PoisonTag's own prefault rather than copying a literal 0 into a `min(1)` field.

app/src/validators/combat.ts
MathiasGruber: Done in 51f47480e — collapsed into a single `e.type === "vamp" || e.type === "consume"` branch with the tag name interpolated into the message.

MathiasGruber (approved):

PR #1399, opened 2026-07-18
feat: copy and mirror max 4 tags with prio list

ifroshty:
## What was implemented

The `copy` and `mirror` combat tags are now capped at **4 transferred effects each**, selected by a fixed commitment-priority order instead of transferring everything at once:

- **Persistent per-caster ceiling of 4** — the cap counts *active* transferred effects across casts, not per cast. A new cast only fills the remaining open slots (`remaining = 4 - active`), and slots free up naturally as transferred effects expire. Copies are identified by `creatorId` + `fromEffectId`, so an opponent mirroring a debuff onto you does not consume your own copy budget.
- **Unique per type** — at most one transferred effect of each type may be held at a time; among duplicates the highest-power one wins.
- **Commitment-priority tiers** (new constants in `combat/constants.ts`):
  - Copy: `increasedamagegiven`/`decreasedamagetaken` → `lifesteal` → `absorb`/`reflect` → `shield` → `increaseheal` → `increasestat`
  - Mirror: `increasedamagetaken`/`decreasedamagegiven` → `afterburn` → `poison`/`increasepoolcost` → `wound` → all other negative types (tail, tie-broken by power)
- **Copy's eligible set expanded to 8 positive types** (adds `absorb`, `lifesteal`, `reflect`, `increaseheal` to the previous set).
- **Mirror can now transfer `wound`** (previously excluded); the existing drain-halving behavior is preserved.
- Selection logic lives in a new pure helper `selectTransferEffects` in `combat/util.ts` (dedupe by type → rank by tier, power, id → cap), shared by both tags. All caps/tier data come from `combat/constants.ts` — no hardcoded values in `tags.ts`.

## Why

Uncapped copy/mirror could transfer every effect on a target in a single cast, which scaled uncontrollably with effect-stacking builds. A small, priority-ordered budget makes both tags predictable and forces a meaningful choice about what gets transferred, while the persistent ceiling prevents re-casting to stack past the cap.

## Testing

- New suite `app/tests/libs/combat/copy_mirror_priority.test.ts` (17 tests): constants/tier ranks, helper dedupe-before-cap ordering and tie-breaks, copy ceiling/uniqueness/creator-scoping, mirror ceiling/wound eligibility/drain halving.
- Full suite green (385/385), typecheck clean, lint clean on touched files.
- All combat logic is in-memory during action processing — no schema changes, no new DB round-trips.

## Breaking / behavior changes

No schema or API changes; battles in flight at deploy are safe (pre-existing transferred effects count toward the ceiling and expire naturally). Gameplay behavior changes players will notice:

- Copy/mirror now transfer at most 4 effects (previously unlimited) and skip types already held.
- Copy can steal 4 new effect types; mirror can now reflect `wound`.
- Mirror's ceiling is scoped per caster–target pair (4 per opponent in multi-fighter battles), while copy's is global per caster — intentional, matching the transfer direction of each tag.

## License

## Summary by CodeRabbit

* **Improvements**
  * Copy and mirror transfers now cap at **up to 4** effects per cast, with **one effect per type** chosen by consistent priority rules.
  * Transfers respect per-target active effects, prevent overfilling on repeated casts, and ignore ineligible sources/effect types.
  * Mirrored **drain** scales correctly with mirror duration, and combat text now reports the correct transferred count.
* **Bug Fixes**
  * Improved deterministic selection with consistent tie-breaking for equal-strength candidates.
* **Tests**
  * Added automated coverage for priority, capping, exclusions, no-op cases, and selection determinism.


app/src/libs/combat/tags.ts
```diff
@@ -341,67 +342,80 @@ export const mirror = (
...
+            mirroredEffect.power / (effect.rounds || 1),
+          );
+          mirroredEffect.powerPerLevel = Math.floor(
+            mirroredEffect.powerPerLevel / (effect.rounds || 1),
+          );
         }
+        usersEffects.push(mirroredEffect);
+
+        const effectPower = negEffect.power + negEffect.level * negEffect.powerPerLevel;
+        const effectDesc = `${negEffect.type} (${effectPower}${negEffect.calculation === "percentage" ? "%" : ""})`;
```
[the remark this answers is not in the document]
ifroshty: Checked this against `main` (the PR base, `ac33a596e`): the drain block there already divided by `effect.rounds`, not by 2 — `Math.floor(mirroredEffect.power / (effect.rounds || 1))`. Only the *comment* said "Cut drain effects in half"; the code never halved. This PR keeps that exact division-by-duration behavior and just replaced the misleading comment, so there's no contract to restore. Applying `/2` would be an actual balance change and would break the current test, which asserts `Math.floor(40 / 3)` for a 3-round mirror. I've renamed the test from "halves drain power when mirrored" to "scales mirrored drain power down by the mirror tag's duration" so the wording matches the behavior.

app/src/libs/combat/constants.ts
```diff
@@ -172,3 +172,77 @@ export const POST_PIERCE_TAGS: string[] = [
...
+      for (const type of tier) {
+        rank.set(type, index);
+      }
+    }
+  }
+  return rank;
+};
+
+export const COPY_PRIORITY_RANK = buildTagPriorityRank(COPY_PRIORITY_TIERS);
+export const MIRROR_PRIORITY_RANK = buildTagPriorityRank(MIRROR_PRIORITY_TIERS);
```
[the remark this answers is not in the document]
ifroshty: Moved. `COPY_MAX_TAGS`, `MIRROR_MAX_TAGS`, `COPY_PRIORITY_TIERS`, `MIRROR_PRIORITY_TIERS`, `COPYABLE_EFFECT_TYPES`, `TRANSFER_EXCLUDED_SOURCE_TYPES`, `MIRROR_EXCLUDED_EFFECT_TYPES`, the `buildTagPriorityRank` helper, and both rank maps now live in `@/drizzle/constants` (grouped with the other combat-domain constants, just above the DMG-setting block). `combat/constants.ts` no longer holds them; `tags.ts` and the test import them from `@/drizzle/constants`. typecheck and the 18-test suite pass.

app/src/libs/combat/tags.ts
```diff
@@ -231,71 +240,71 @@ export const copy = (
...
+      // with a fromEffectId and must not consume the copy budget.
+      // Clones always carry numeric rounds (the tag gate requires effect.rounds); undefined-rounds effects are never clones.
+      const activeCopied = usersEffects.filter(
+        (e) =>
+          e.targetId === user.userId &&
+          e.creatorId === user.userId &&
+          e.fromEffectId &&
+          e.rounds &&
+          e.rounds > 0,
+      );
```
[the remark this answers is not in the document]
ifroshty: Done in `001db12`. `activeCopied` now includes `COPYABLE_EFFECT_TYPES.includes(e.type)`, so a self-mirror (a negative type landing on the caster) can't consume a copy slot. The comment above the filter documents that invariant.

ifroshty:
@coderabbitai review

app/src/libs/combat/tags.ts
```diff
@@ -324,84 +349,77 @@ export const mirror = (
...
-      const negativeEffects = usersEffects.filter(
+      // Persistent, per-caster ceiling: only this caster's active mirrors on the target.
+      const activeMirrored = usersEffects.filter(
+        (e) =>
+          e.targetId === target.userId &&
+          e.creatorId === user.userId &&
+          e.fromEffectId &&
+          e.rounds &&
+          e.rounds > 0,
+      );
```
MathiasGruber, line 428: Consider mirroring the forward-safety guard that was added to `activeCopied` (line 261: `COPYABLE_EFFECT_TYPES.includes(e.type)`). The `activeMirrored` filter here counts any effect on the target created by this user with a `fromEffectId`, without gating on effect type. Today only the `mirror` tag itself produces this signature so there is no current-path bug — but if any future tag ever places a `fromEffectId`-bearing effect on an opponent with `creatorId === user.userId`, it will silently consume the mirror budget. The symmetric guard would be `!MIRROR_EXCLUDED_EFFECT_TYPES.includes(e.type)` (or, if you want the tightest analogue to `COPYABLE_EFFECT_TYPES`, a `MIRRORABLE_EFFECT_TYPES` allowlist alongside the exclude list). This keeps the two tags' budget-accounting semantics parallel and self-documenting.

MathiasGruber (commented):
Follow-up on the Greptile summary review: one unaddressed forward-safety concern about `activeMirrored` budget accounting. Not a current-path bug — flagging for symmetry with `activeCopied`.

app/src/libs/combat/tags.ts
ifroshty: Fixed the underlying cause rather than mirroring the guard. `copy` and `mirror` now early-return on a self-cast (`user.userId === target.userId`). Self-copy/self-mirror are degenerate (copying your own buffs onto yourself / doubling your own debuffs), and the self-cast is exactly what would let the asymmetry bite: when caster === target, copy clones and mirror clones share the same `(targetId, creatorId, fromEffectId)` identity, so one tag's clones could count against the other's ceiling. With self-casts blocked, `activeMirrored` only ever sees this caster's mirror clones — it doesn't need the type guard `activeCopied` carries, which now stands purely as forward-safety. Added self-copy and self-mirror no-op tests (suite 18 → 20, all green).

app/tests/libs/combat/copy_mirror_priority.test.ts
```diff
@@ -0,0 +1,406 @@
...
+      targetId: OPP,
+      creatorId: SELF,
+      level: 0,
+      isNew: true,
+      castThisRound: true,
+      createdRound: 1,
+    },
+  );
+
+describe("mirror tag: ceiling + priority + wound", () => {
```
MathiasGruber, line 382: Re-opening the one review item from the Greptile summary that was never acted on: this `describe` block still has no persistent-ceiling test for `mirror`, while `copy` has the symmetric one (`respects the persistent ceiling across casts (already at 4 -> copies nothing)`, line 263).

The mirror ceiling path is the least-covered new logic in the PR — `activeMirrored` in `tags.ts` uses a denylist (`!MIRROR_EXCLUDED_EFFECT_TYPES.includes(e.type)`) rather than the allowlist `activeCopied` uses, and nothing currently exercises the `remaining <= 0` branch or its `is already holding the maximum mirrored effects.` message.

Please add a test that seeds 4 effects with `targetId: OPP, creatorId: SELF, fromEffectId: ...` plus one fresh mirrorable debuff on SELF, casts `mirror`, and asserts nothing new is added. Worth also covering the creator-scoping counterpart of the existing copy test (an opponent-created `fromEffectId` effect on OPP must not consume this caster's mirror budget).

app/src/libs/combat/tags.ts
```diff
@@ -324,84 +349,82 @@ export const mirror = (
...
+        mirroredEffect.id = nanoid();
+        mirroredEffect.fromEffectId = negEffect.id;
+        mirroredEffect.targetId = target.userId;
+        mirroredEffect.creatorId = user.userId;
+        mirroredEffect.rounds = effect.rounds;
+        mirroredEffect.isNew = true;
+        mirroredEffect.castThisRound = true;
+        mirroredEffect.createdRound = effect.createdRound;
+        // Divide drain power by the mirror duration so the total drain transferred stays bounded
+        if (negEffect.type === "drain") {
```
MathiasGruber: `drain` competes for a capped mirror slot at its **full source power**, but is then scaled down to `power / effect.rounds` *after* selection — so its selection weight does not reflect what it actually delivers.

`selectTransferEffects` ranks candidates by `getPower(negEffect).power` (the untouched source value), and `drain` is not listed in `MIRROR_PRIORITY_TIERS`, so it lands in the tail tier where ordering is decided purely by that power number. Concrete case: self has `drain` (power 60) and `decreasestat` (power 50), the mirror tag has `rounds: 3`, and only one tail slot is left. `drain` wins the slot on 60 vs 50, then gets written to the target with `power = floor(60 / 3) = 20` — i.e. the weaker transfer displaced the stronger one.

Either divide before ranking (pass the effective power into the selector) or give `drain` an explicit tier so its placement isn't decided by a number the code itself immediately discards.

app/src/libs/combat/tags.ts
```diff
@@ -324,84 +349,82 @@ export const mirror = (
...
-        "timecompression",
-      ];
-
-      const negativeEffects = usersEffects.filter(
+      // Persistent, per-caster ceiling: only this caster's active mirrors on the target.
+      // The MIRROR_EXCLUDED_EFFECT_TYPES guard scopes the budget to effects `mirror`
+      // can actually produce (mirroring the COPYABLE_EFFECT_TYPES guard on activeCopied),
+      // so a future tag that lands a fromEffectId-bearing effect on the target with
+      // creatorId === user.userId can't silently consume a mirror slot.
+      const activeMirrored = usersEffects.filter(
```
MathiasGruber: The comment above this filter says the `MIRROR_EXCLUDED_EFFECT_TYPES` guard means "a future tag that lands a `fromEffectId`-bearing effect on the target with `creatorId === user.userId` can't silently consume a mirror slot" — but a **denylist cannot provide that guarantee**. Only the 8 types currently on the list are rejected; any new or unlisted type passes the filter and *does* consume a slot.

The copy path gets this right because `activeCopied` uses `COPYABLE_EFFECT_TYPES` — an allowlist of exactly what `copy` can produce, so anything else is rejected by construction. For symmetry and to make the comment true, scope the mirror budget with an allowlist too (or at minimum add `isNegativeUserEffect(e)`, since mirror only ever produces negative effects).

This asymmetry is a symptom of the two paths still being ~50 lines of near-duplicated logic (budget filter → `heldTypes` → `remaining` guard → candidate filter → clone/push loop → message). Only the ranking step was factored into `selectTransferEffects`; the budget-accounting step, which is where the two have already drifted apart, was not.

app/drizzle/constants.ts
```diff
@@ -2644,6 +2644,90 @@ export const isPreBattleKeystoneFromType = (
...
+];
+
+/** Effect types the `copy` tag may transfer (flattened copy priority tiers). */
+export const COPYABLE_EFFECT_TYPES: string[] = COPY_PRIORITY_TIERS.flat();
+
+/**
+ * Source (`fromType`) origins that copy/mirror never transfer — passive/gear
+ * effects are not stealable/reflectable. Shared by both tags.
+ */
+export const TRANSFER_EXCLUDED_SOURCE_TYPES: string[] = [
```
MathiasGruber: This list is documented as covering "passive/gear effects", but it omits `"ranked"`, which is a declared member of `UserEffect["fromType"]` (`app/src/libs/combat/types.ts`) and is treated as a passive/pre-battle source everywhere else in the combat engine:

- `getEffectStage` (`util.ts`) classifies it Stage 1 alongside `armor`/`accessory`/`keystone`/`skill`/`village`
- `persistentEffectSourceTypes` (`tags.ts`, used by `clear`/`cleanse`) protects it from removal alongside the same set

So a `ranked`-sourced passive is uncleansable but freely stealable/reflectable — and it's worse under this PR than before, because `increasedamagegiven` / `decreasedamagetaken` are now tier 0, i.e. the *first* things `copy` reaches for. Nothing emits `fromType: "ranked"` today so this is latent rather than live, but since this PR is the place that turns the list into the canonical shared contract, adding `"ranked"` here is a one-line fix that keeps the three lists consistent.

app/tests/libs/combat/copy_mirror_priority.test.ts
```diff
@@ -0,0 +1,406 @@
...
+      oppBuff("absorb", 50, { id: "c4", targetId: SELF, creatorId: SELF, fromEffectId: "src4" }),
+      // a fresh copyable buff on the opponent
+      oppBuff("shield", 90),
+    ];
+    const before = effects.length;
+    copy(copyTag(), effects, asUser(SELF, "Self"), asUser(OPP, "Opp"));
+    expect(effects.length).toBe(before); // nothing added
+  });
+
+  it("does not count an opponent's mirror on self against the copy ceiling", () => {
```
MathiasGruber: This test does not exercise the behavior it names — it passes even if the `creatorId === user.userId` guard is deleted.

The fixture effect is `increasedamagetaken`, which is rejected by `activeCopied` on *two* independent grounds: `creatorId !== user.userId` **and** `!COPYABLE_EFFECT_TYPES.includes("increasedamagetaken")`. On top of that, even if both guards were removed, `activeCopied.length` would be 1, `remaining` would be 3, and the `increasedamagegiven` copy the test asserts on would still happen. So no assertion here can fail regardless of the creator-scoping logic.

To actually pin the guard, use a **copyable** type on SELF with `creatorId: OPP` and enough of them to matter — e.g. 4 copyable-typed effects on SELF with `creatorId: OPP` plus a fresh buff on the opponent; without the `creatorId` scope those 4 would exhaust the budget and the copy would be blocked, with it the copy must still succeed.

app/tests/libs/combat/copy_mirror_priority.test.ts
MathiasGruber, line 382: Coverage gaps in this block relative to what the PR claims:

1. **No test of mirror's persistent ceiling across casts.** The copy path has one ("already at 4 -> copies nothing"); mirror has none, even though `activeMirrored` uses a *different* predicate from `activeCopied` (denylist vs allowlist) and is therefore the riskier of the two. Add the mirror analogue: 4 pre-existing `fromEffectId` effects on OPP with `creatorId: SELF`, plus a fresh debuff on SELF, and assert nothing is added.
2. **No test that `heldTypes` skipping works for mirror** (the copy path covers this with the `lifesteal` test).
3. **The exclusion lists are only asserted as array membership, never as behavior.** There is no test that a `fromType: "bloodline"` buff is skipped by `copy`, and none that a `MIRROR_EXCLUDED_EFFECT_TYPES` entry (e.g. `damage`, `moveprevent`) is skipped by `mirror`. `expect(TRANSFER_EXCLUDED_SOURCE_TYPES).toContain("bloodline")` re-states the constant; it would still pass if `tags.ts` stopped consulting the list at all.

MathiasGruber: Both added in 531531aed: `respects the persistent ceiling across casts (already at 4 -> mirrors nothing)` seeds 4 fromEffectId effects on OPP with creatorId: SELF plus a fresh debuff on SELF, asserts nothing is added, and pins the `maximum mirrored` message — the `remaining <= 0` branch is now exercised. `does not count opponent-created transfers on the target against the mirror ceiling` covers the creator-scoping counterpart (4 opponent-created fromEffectId effects on OPP must not consume this caster's budget).

app/src/libs/combat/tags.ts
MathiasGruber: Fixed in 531531aed (refined in 713c82a1d): drain candidates are scaled to delivered power BEFORE selection — power = floor((power + level×powerPerLevel) / mirror rounds), with the level contribution baked into the base so fractional powerPerLevel is not floored to zero — so drain competes for a capped slot at the value it actually delivers. The in-loop special case is gone; the pre-scaled stand-in keeps the source id, so fromEffectId provenance is unchanged. Regression test: `ranks drain by delivered (duration-scaled) power when competing for a slot` (drain 60 vs decreasestat 50, one tail slot → decreasestat wins).

app/src/libs/combat/tags.ts
MathiasGruber: Fixed in 531531aed: the budget is now scoped by `isMirrorableEffect` = `isNegativeUserEffect(e) && !MIRROR_EXCLUDED_EFFECT_TYPES.has(e.type)` — the exact predicate that gates the candidates — so it counts precisely what mirror can produce, by construction (same shape as copy's `isCopyableEffect`). The drifted budget accounting is deduplicated as well: both tags share `getTransferBudget`, `isTransferCandidate`, and `applyTransferClones` now; only the direction parameters and messages differ. The misleading comment is gone.

app/drizzle/constants.ts
MathiasGruber: Added in 531531aed — `ranked` is in TRANSFER_EXCLUDED_SOURCE_TYPES (a ReadonlySet as of 713c82a1d), with a doc note tying the list to persistentEffectSourceTypes / getEffectStage, a membership assertion, and a behavior test that fromType "ranked" (and "bloodline") buffs are skipped by copy.

app/tests/libs/combat/copy_mirror_priority.test.ts
MathiasGruber: Replaced in 531531aed with the fixture you outlined: 4 copyable-typed, fromEffectId-bearing effects on SELF with creatorId: OPP plus a fresh opponent buff — without the creatorId scope those exhaust the ceiling and block the cast, so the test fails if the guard is deleted. A second test (`does not count non-copyable-typed clones (e.g. mirrors) against the copy ceiling`) pins the COPYABLE_EFFECT_TYPES guard independently with 4 negative-typed self-created clones.

app/tests/libs/combat/copy_mirror_priority.test.ts
MathiasGruber: All three added in 531531aed: (1) the mirror persistent-ceiling test; (2) `skips a type already actively mirrored` (held poison vs stronger fresh poison); (3) exclusion lists as behavior — `skips buffs from excluded passive sources (bloodline, ranked)` for copy and `skips excluded effect types (damage)` for mirror. 713c82a1d additionally covers depleted (power-0) and ground-derived sources, magnitude ranking for in-place-negated decrease* powers, and tier/classifier alignment as a typo guard.

MathiasGruber (approved):

PR #1400, opened 2026-07-18
Better anbu management

Phrosfire:
# Pull Request

Allow staff to remove & promote anbu members when necessary.  Allow kage to accept anbu join requests

## License

## Summary by CodeRabbit

* **New Features**
  * Expanded ANBU staff permissions using clan-edit eligibility to manage squads and moderate requests beyond prior village limits.
  * Enhanced ANBU request handling with squad-aware moderation, improved “send request” gating, and persistent squad identity across leader changes.
  * Updated sparring request fetching to support optional squad/related targeting.

* **Bug Fixes**
  * Improved authorization for accepting/rejecting requests and relaxed promote/kick guards for eligible staff.
  * Added safer request acceptance with requester-rank validation and atomic join/rollback behavior.
  * Prevented unintended membership clearing by only removing ANBU links for the targeted squad.

* **Tests**
  * Added coverage for squad identity resolution and pending request reassignment on promotion.

### DIFF
--- a/app/drizzle/migrations/0027_backfill_anbu_request_related_id.sql
+++ b/app/drizzle/migrations/0027_backfill_anbu_request_related_id.sql
@@ -1,0 +1,39 @@
+-- Backfill UserRequest.relatedId for ANBU join requests created before squad id
+-- was persisted on relatedId. Without this, getRequests filters by relatedId =
+-- squadId and silently drops legacy rows (relatedId IS NULL), hiding pending
+-- requests from leaders and applicants after deploy.
+--
+-- Only match via the current squad leader. Do not join through the receiver's
+-- current anbuId — a demoted/kicked leader who later joined a different squad
+-- would incorrectly re-link the original squad's requests. Remaining unmatched
+-- rows (relatedId still null) are covered by getRequests' legacy receiverId ===
+-- leaderId fallback while that leader still leads the squad.
+
+UPDATE `UserRequest` AS ur
+INNER JOIN `AnbuSquad` AS a ON a.leaderId = ur.receiverId
+SET ur.relatedId = a.id
+WHERE ur.type = 'ANBU'
+  AND ur.status = 'PENDING'
+  AND ur.createdAt > NOW() - INTERVAL 1 DAY
+  AND ur.relatedId IS NULL;
+
+-- Keep the oldest pending request per sender before the following schema
+-- migration adds the unique pending-ANBU constraint. This only affects rows
+-- that could otherwise race multiple squad leaders during rollout.
+UPDATE `UserRequest` AS ur
+INNER JOIN (
+  SELECT id
+  FROM (
+    SELECT
+      id,
+      ROW_NUMBER() OVER (
+        PARTITION BY senderId
+        ORDER BY createdAt ASC, id ASC
+      ) AS request_rank
+    FROM `UserRequest`
+    WHERE type = 'ANBU'
+      AND status = 'PENDING'
+  ) AS ranked_requests
+  WHERE request_rank > 1
+) AS duplicate_requests ON duplicate_requests.id = ur.id
+SET ur.status = 'CANCELLED';
--- a/app/src/app/anbu/[anbuid]/page.tsx
+++ b/app/src/app/anbu/[anbuid]/page.tsx
@@ -102,9 +102,11 @@ export default function ANBUDetails(props: { params: Promise<{ anbuid: string }>
   if (userData.isOutlaw) return <Loader explanation="Unlikely to find outlaw ANBU" />;
   if (userData.isBanned) return <BanInfo />;
 
-  // Derived
-  const isKage = userData.userId === userData.village?.kageId;
-  const isElder = userData.rank === "ELDER";
+  // Derived — village-gate kage/elder so cross-village staff (canStaffEdit) do not
+  // inherit village-scoped moderation UI for a foreign squad.
+  const sameVillage = userData.villageId === squad.villageId;
+  const isKage = sameVillage && userData.userId === userData.village?.kageId;
+  const isElder = sameVillage && userData.rank === "ELDER";
   const isLeader = userData.userId === squad.leaderId;
   const inSquad = userData.anbuId === squadId;
   const canStaffEdit = canEditClans(userData.role);
@@ -149,7 +151,7 @@ export default function ANBUDetails(props: { params: Promise<{ anbuid: string }>
         userId={userData.userId}
         userRank={userData.rank}
         userAnbu={userData.anbuId}
-        sameVillage={userData.villageId === squad.villageId}
+        sameVillage={sameVillage}
         canStaffEdit={canStaffEdit}
       />
       {/* ANBU QUESTS - Only show if user is in the squad */}
@@ -802,10 +804,10 @@ const AnbuRequests: React.FC<AnbuRequestsProps> = (props) => {
   if (!requests) return <Loader explanation="Loading requests" />;
 
   // Derived
-  const canManageRequests = isLeader || isKage || canStaffEdit;
+  const canManageRequests = isLeader || isKage || isElder || canStaffEdit;
   // Kage/elder cannot join (blocked server-side); leaders are always in a squad.
-  // Cross-village viewers (e.g. staff) also cannot join, since the server rejects
-  // join requests for a squad in a different village.
+  // Cross-village viewers (e.g. staff) also cannot join. Leaderless squads stay
+  // joinable — kage/elder/staff can accept via relatedId.
   const canJoinSquad = !isKage && !isElder && sameVillage;
   const hasPending = requests?.some((req) => req.status === "PENDING");
   const showRequestSystem =
--- a/app/src/app/anbu/[anbuid]/page.tsx
+++ b/app/src/app/anbu/[anbuid]/page.tsx
@@ -72,6 +72,7 @@ import type { AnbuRouter } from "@/routers/anbu";
 import type { UserWithRelations } from "@/routers/profile";
 import type { BaseServerResponse } from "@/server/api/trpc";
 import { parseHtml } from "@/utils/parse";
+import { canEditClans } from "@/utils/permissions";
 import type { ArrayElement } from "@/utils/typeutils";
 import { useRequireInVillage } from "@/utils/UserContext";
 import { UploadButton } from "@/utils/uploadthing";
@@ -106,6 +107,7 @@ export default function ANBUDetails(props: { params: Promise<{ anbuid: string }>
   const isElder = userData.rank === "ELDER";
   const isLeader = userData.userId === squad.leaderId;
   const inSquad = userData.anbuId === squadId;
+  const canStaffEdit = canEditClans(userData.role);
 
   return (
     <>
@@ -115,6 +117,7 @@ export default function ANBUDetails(props: { params: Promise<{ anbuid: string }>
         isElder={isElder}
         isLeader={isLeader}
         inSquad={inSquad}
+        canStaffEdit={canStaffEdit}
         userId={userData.userId}
         squadId={squadId}
         squad={squad}
@@ -141,9 +144,13 @@ export default function ANBUDetails(props: { params: Promise<{ anbuid: string }>
       <AnbuRequests
         squadId={squadId}
         isLeader={isLeader}
+        isKage={isKage}
+        isElder={isElder}
         userId={userData.userId}
         userRank={userData.rank}
         userAnbu={userData.anbuId}
+        sameVillage={userData.villageId === squad.villageId}
+        canStaffEdit={canStaffEdit}
       />
       {/* ANBU QUESTS - Only show if user is in the squad */}
       {inSquad && (
@@ -163,6 +170,7 @@ interface AnbuMembersProps {
   isElder: boolean;
   isLeader: boolean;
   inSquad: boolean;
+  canStaffEdit: boolean;
   userId: string;
   squadId: string;
   squad: NonNullable<AnbuRouter["get"]>;
@@ -171,8 +179,17 @@ interface AnbuMembersProps {
 
 const AnbuMembers: React.FC<AnbuMembersProps> = (props) => {
   // Destructure
-  const { userId, squad, squadId, isKage, isElder, isLeader, inSquad, userData } =
-    props;
+  const {
+    userId,
+    squad,
+    squadId,
+    isKage,
+    isElder,
+    isLeader,
+    inSquad,
+    canStaffEdit,
+    userData,
+  } = props;
 
   // Get router
   const router = useRouter();
@@ -261,7 +278,7 @@ const AnbuMembers: React.FC<AnbuMembersProps> = (props) => {
             Confirm that you want to kick this member from the squad.
           </Confirm2>
         )}
-        {(isKage || isElder) && (
+        {(isKage || isElder || canStaffEdit) && (
           <Confirm2
             title="Promote Member"
             proceed_label="Submit"
@@ -288,7 +305,7 @@ const AnbuMembers: React.FC<AnbuMembersProps> = (props) => {
     { key: "rank", header: "Rank", type: "capitalized" },
     { key: "pvpActivity", header: "PVP Activity", type: "string" },
   ];
-  if (isLeader || isKage || isElder) {
+  if (isLeader || isKage || isElder || canStaffEdit) {
     columns.push({ key: "kickBtn", header: "Action", type: "jsx" });
   }
 
@@ -724,23 +741,40 @@ const AnbuOrders: React.FC<AnbuOrdersProps> = (props) => {
 interface AnbuRequestsProps {
   squadId: string;
   isLeader: boolean;
+  isKage: boolean;
+  isElder: boolean;
   userId: string;
   userRank: UserRank;
   userAnbu: string | null;
+  sameVillage: boolean;
+  canStaffEdit: boolean;
 }
 
 const AnbuRequests: React.FC<AnbuRequestsProps> = (props) => {
   // Destructure
-  const { squadId, isLeader, userId, userRank, userAnbu } = props;
+  const {
+    squadId,
+    isLeader,
+    isKage,
+    isElder,
+    userId,
+    userRank,
+    userAnbu,
+    sameVillage,
+    canStaffEdit,
+  } = props;
 
   // Get utils
   const utils = api.useUtils();
 
   // Query
-  const { data: requests } = api.anbu.getRequests.useQuery(undefined, {
-    enabled: !!squadId,
-    staleTime: 5000,
-  });
+  const { data: requests } = api.anbu.getRequests.useQuery(
+    { squadId },
+    {
+      enabled: !!squadId,
+      staleTime: 5000,
+    },
+  );
 
   // How to deal with success responses
   const onSuccess = async (data: BaseServerResponse) => {
@@ -768,9 +802,17 @@ const AnbuRequests: React.FC<AnbuRequestsProps> = (props) => {
   if (!requests) return <Loader explanation="Loading requests" />;
 
   // Derived
+  const canManageRequests = isLeader || isKage || canStaffEdit;
+  // Kage/elder cannot join (blocked server-side); leaders are always in a squad.
+  // Cross-village viewers (e.g. staff) also cannot join, since the server rejects
+  // join requests for a squad in a different village.
+  const canJoinSquad = !isKage && !isElder && sameVillage;
   const hasPending = requests?.some((req) => req.status === "PENDING");
-  const showRequestSystem = (isLeader && requests.length > 0) || !userAnbu;
-  const shownRequests = requests.filter((r) => !isLeader || r.status === "PENDING");
+  const showRequestSystem =
+    (canManageRequests && requests.length > 0) || (!userAnbu && canJoinSquad);
+  const shownRequests = requests.filter(
+    (r) => !canManageRequests || r.status === "PENDING",
+  );
   const sufficientRank = hasRequiredRank(userRank, ANBU_MEMBER_RANK_REQUIREMENT);
 
   // Do not show?
@@ -785,7 +827,7 @@ const AnbuRequests: React.FC<AnbuRequestsProps> = (props) => {
       padding={false}
     >
       {/* FOR THOSE WHO CAN SEND REQUESTS */}
-      {sufficientRank && !userAnbu && !hasPending && (
+      {sufficientRank && !userAnbu && !hasPending && canJoinSquad && (
         <div className="p-2">
           <p>Send a request to join this squad</p>
           <Button id="send" className="mt-2 w-full" onClick={() => create({ squadId })}>
@@ -804,6 +846,7 @@ const AnbuRequests: React.FC<AnbuRequestsProps> = (props) => {
           onReject={reject}
           onCancel={cancel}
           isLoading={isCreating || isAccepting || isRejecting || isCancelling}
+          canModerateRequests={canManageRequests}
         />
       )}
     </ContentBox>
--- a/app/src/app/anbu/[anbuid]/page.tsx
+++ b/app/src/app/anbu/[anbuid]/page.tsx
@@ -769,46 +769,50 @@ const AnbuRequests: React.FC<AnbuRequestsProps> = (props) => {
   // Get utils
   const utils = api.useUtils();
 
+  const canManageRequests = isLeader || isKage || isElder || canStaffEdit;
+  // Kage/elder cannot join (blocked server-side). Cross-village viewers cannot
+  // join either, while leaderless squads remain joinable by eligible villagers.
+  const canJoinSquad = !isKage && !isElder && sameVillage;
+  const shouldFetch = canManageRequests || (!userAnbu && canJoinSquad);
+
   // Query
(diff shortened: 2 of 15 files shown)


Phrosfire:
@coderabbitai review

Phrosfire:
@coderabbitai review

app/src/server/api/routers/anbu.ts
```diff
@@ -142,7 +173,21 @@ export const anbuRouter = createTRPCRouter({
     .output(baseServerResponse)
     .mutation(async ({ ctx, input }) => {
       const request = await fetchRequest(ctx.drizzle, input.id, "ANBU");
-      if (request.receiverId !== ctx.userId) {
+      const [updatedUser, squad] = await Promise.all([
```
MathiasGruber: `fetchRequest` and `fetchUpdatedUser` are independent — `fetchUpdatedUser` only needs `ctx.userId`, not any field from `request`. As written, they serialize into two round-trips before we can hit `fetchSquadByLeader`. Move `fetchUpdatedUser` into a `Promise.all` with the initial `fetchRequest` so the sequence becomes `[request, updatedUser]` in parallel, then `fetchSquadByLeader(request.receiverId)` — saves one DB round-trip on every reject.

app/src/server/api/routers/anbu.ts
```diff
@@ -174,34 +219,79 @@ export const anbuRouter = createTRPCRouter({
       // Fetch
       const request = await fetchRequest(ctx.drizzle, input.id, "ANBU");
       // Secondary fetches
-      const [squad, requester, leader] = await Promise.all([
+      const [updatedUser, squad, requester, leader] = await Promise.all([
```
MathiasGruber: Same pattern as `rejectRequest`: `fetchUpdatedUser` was added to the secondary `Promise.all`, but it doesn't depend on `request` at all — only `ctx.userId`. It should be pulled up into the same batch as the initial `fetchRequest` (i.e. `const [request, updatedUser] = await Promise.all([...])`), and only `fetchSquadByLeader` / `fetchUser(senderId)` / `fetchUser(receiverId)` need to wait for `request`. Currently this hot mutation is 3 sequential round-trips (fetchRequest → parallel batch → mutation) when it can be 2.

app/src/server/api/routers/anbu.ts
```diff
@@ -101,8 +106,34 @@ export const anbuRouter = createTRPCRouter({
...
+        ]);
+        const { user } = updatedUser;
+        if (!user || !squad?.leaderId) {
+          return [];
+        }
+        const isLeader = user.userId === squad.leaderId;
+        const isKageOfVillage =
+          user.userId === user.village?.kageId && user.villageId === squad.villageId;
+        if (isLeader || isKageOfVillage) {
+          return fetchRequests(ctx.drizzle, ["ANBU"], 3600 * 12, squad.leaderId);
```
MathiasGruber: This path serializes into three DB round-trips: `Promise.all([fetchUpdatedUser, fetchSquad])` → `fetchRequests`. Since `fetchRequests(..., ctx.userId)` only needs `ctx.userId`, it can join the initial parallel batch. In the leader case `squad.leaderId === ctx.userId`, so that same result satisfies the leader branch; only the kage-not-leader branch would still need a follow-up fetch for `squad.leaderId`. This cuts a round-trip in both the common leader path and the applicant path.

MathiasGruber (commented):
Performance review: found a few spots where independent fetches are serialized and could run in parallel.

app/src/app/anbu/[anbuid]/page.tsx
```diff
@@ -785,7 +811,7 @@ const AnbuRequests: React.FC<AnbuRequestsProps> = (props) => {
       padding={false}
     >
       {/* FOR THOSE WHO CAN SEND REQUESTS */}
-      {sufficientRank && !userAnbu && !hasPending && (
+      {sufficientRank && !userAnbu && !hasPending && canJoinSquad && (
```
Comment (MathiasGruber) [inline app/src/app/anbu/[anbuid]/page.tsx:836]: `canJoinSquad = !isKage && !isElder` (line 793) doesn't account for cross-village staff viewers. A `canStaffEdit` user opening a foreign ANBU squad passes `!isKage && !isElder`, and if they aren't already in a squad (`!userAnbu`) with sufficient rank they'll see the `Send Request` button here. Server-side `createRequest` then rejects with `Wrong village` (anbu.ts line 158), so the UI offers an action that always fails.

The browser test (`08-cross-village-staff-request-bug.png`) confirms this regression: a Tsukimori staff account viewing a Shirohana squad sees the CTA and the click errors out. Either gate the button on `userData.villageId === squad.villageId` (needs squad village passed into `AnbuRequests`) or exclude `canStaffEdit` viewers via an explicit flag similar to `canManageRequests`.

MathiasGruber (commented):
Following up on the tnr-review-now browser test finding about a dead-end `Send Request` CTA for cross-village staff viewers.

MathiasGruber (commented):
Two items — a possible feature gap and a small DRY note.

app/src/server/api/routers/anbu.ts
```diff
@@ -174,34 +219,79 @@ export const anbuRouter = createTRPCRouter({
...
+        !!user &&
+        !!squad &&
+        user.userId === user.village?.kageId &&
+        user.villageId === squad.villageId;
       // Guards
       if (!squad) return errorResponse("Squad not found");
+      if (!user) return errorResponse("User not found");
       if (!requester) return errorResponse("Requester not found");
       if (!leader) return errorResponse("Leader not found");
+      if (!isReceiver && !isKageOfVillage) return errorResponse("Not allowed");
```
MathiasGruber: Staff (`canEditClans`) can now promote/kick members (see `promoteMember`/`kickMember`) but not accept/reject join requests — `acceptRequest`, `rejectRequest`, and `getRequests` only recognize the receiver/leader or the village kage. If the intent of this PR (broader staff moderation across villages) is for admins to fully moderate ANBU membership, these three procedures should also allow `canEditClans(user.role)`, and the frontend `canManageRequests = isLeader || isKage` in `page.tsx` should include `canStaffEdit`. Otherwise a Content Admin can kick a member but cannot reject that same user's pending request.

app/src/server/api/routers/anbu.ts
```diff
@@ -174,34 +219,79 @@ export const anbuRouter = createTRPCRouter({
...
       ]);
       // Derived
+      const { user } = updatedUser;
       const nMembers = squad?.members.length || 0;
+      const isReceiver = ctx.userId === request.receiverId;
+      const isKageOfVillage =
+        !!user &&
+        !!squad &&
+        user.userId === user.village?.kageId &&
+        user.villageId === squad.villageId;
```
MathiasGruber: The `user.userId === user.village?.kageId && user.villageId === squad.villageId` check (kage of *this squad's* village, distinct from `getConvenienceStatus.isKage` which only checks user vs. their own village) is now inlined in three places: here, `rejectRequest`, and `getRequests`. Consider lifting it into `getConvenienceStatus` (e.g. as `isKageOfSquadVillage`) so the two `isKage` semantics don't get conflated in a future call site.

MathiasGruber:

**Conflicts resolved (migration metadata only):**

- `app/drizzle/migrations/meta/_journal.json` and `meta/0022_snapshot.json` conflicted because `main` had advanced to `0024_military_punisher` while this branch still claimed slot `0022`.
- Kept `main`'s versions of both meta files, and renumbered this PR's migration `0022_tearful_wilson_fisk.sql` → `0025_tearful_wilson_fisk.sql`. The SQL body (the `UserRequest.relatedId` ANBU backfill) is unchanged.
- Added a matching `meta/0025_snapshot.json` and journal entry. Since this is a data-only migration with no `schema.ts` changes, the snapshot is schema-identical to `0024` — confirmed by `bun makemigrations` reporting "No schema changes, nothing to migrate".

No conflicts in application code; `anbu.ts`, `page.tsx`, `UserRequestSystem.tsx`, `sparring.ts` and the tests replayed cleanly.

**Verification after rebase:** typecheck ✅, lint ✅, unit tests ✅ (549 passed).

app/src/server/api/routers/anbu.ts
```diff
@@ -101,8 +106,57 @@ export const anbuRouter = createTRPCRouter({
...
+            fetchRequests(ctx.drizzle, ["ANBU"], 3600 * 12, undefined, squadId),
+            squad.leaderId
+              ? fetchRequests(ctx.drizzle, ["ANBU"], 3600 * 12, squad.leaderId)
+              : Promise.resolve([]),
+          ]);
+          const seen = new Set(byRelatedId.map((r) => r.id));
+          const legacy = byLeader.filter(
+            (r) => !seen.has(r.id) && isAnbuRequestForSquad(r, squadId, squad.leaderId),
+          );
+          return [...byRelatedId, ...legacy];
```
MathiasGruber: This adds a permanent second DB round-trip on every kage/elder/staff view of a squad page, and it exists only to serve the legacy `relatedId IS NULL` fallback.

That fallback can't survive more than 24h:
- `createRequest` now always writes `relatedId` (`insertRequest(..., squad.id)`).
- The migration in this PR backfills `relatedId` for exactly the rows the fallback matches (`a.leaderId = ur.receiverId`).
- The cleaner hard-deletes every `UserRequest` older than 24h (`app/src/app/api/cleaner/route.ts`, step 28), so any row that escaped the backfill is gone within a day.

Meanwhile `fetchRequests(..., squadId)` only needs `squadId`, which is known before the first `Promise.all` — so it can be hoisted and the whole second batch dropped, making every path a single round-trip:

```ts
const [updatedUser, squad, ownRequests, squadRequests] = await Promise.all([
  fetchUpdatedUser({ client: ctx.drizzle, userId: ctx.userId }),
  fetchSquad(ctx.drizzle, squadId),
  fetchRequests(ctx.drizzle, ["ANBU"], 3600 * 12, ctx.userId),
  fetchRequests(ctx.drizzle, ["ANBU"], 3600 * 12, undefined, squadId),
]);
// ...
if (isKageOfSquadVillage || isElderOfSquadVillage || canStaffEdit) {
  return squadRequests;
}
```

If you want belt-and-braces for the deploy window, keep the `isAnbuRequestForSquad` JS fallback over `ownRequests` (free, already fetched) rather than paying for an extra query forever.

app/src/server/api/routers/anbu.ts
```diff
@@ -163,45 +244,116 @@ export const anbuRouter = createTRPCRouter({
       if (request.status !== "PENDING") {
         return errorResponse("You can only cancel pending requests");
       }
-      void pusher.trigger(request.receiverId, "event", { type: "anbu" });
+      const squad = await fetchSquadForAnbuRequest(ctx.drizzle, request);
+      const notifyUserId = squad?.leaderId ?? request.receiverId;
+      void pusher.trigger(notifyUserId, "event", { type: "anbu" });
```
MathiasGruber: `updateRequestState` on the next line does not consume `squad` — the squad is fetched purely to pick the pusher target. So this serialises a full extra round-trip in front of the write on every cancel.

Also worth noting `fetchSquadForAnbuRequest` -> `fetchSquad` is the heavy relational query (all members with avatars, plus `kageOrder` and `leaderOrder`) and we use exactly one field from it.

```ts
const [squad, result] = await Promise.all([
  fetchSquadForAnbuRequest(ctx.drizzle, request),
  updateRequestState(ctx.drizzle, input.id, "CANCELLED", "ANBU"),
]);
void pusher.trigger(squad?.leaderId ?? request.receiverId, "event", { type: "anbu" });
return result;
```

Better still, when `request.relatedId` is set, a `columns: { id: true, leaderId: true }` lookup on `anbuSquad` would do instead of the full squad hydration.

app/src/server/api/routers/anbu.ts
```diff
@@ -163,45 +244,116 @@ export const anbuRouter = createTRPCRouter({
...
-          .set({ anbuId: squad.id })
-          .where(eq(userData.userId, requester.userId)),
-      ]);
+          .set({ anbuId: null })
+          .where(
+            and(eq(userData.userId, requester.userId), eq(userData.anbuId, squad.id)),
+          );
+        return errorResponse("Squad is full");
+      }
+      const claimRequest = await ctx.drizzle
```
MathiasGruber: The mutation section went from one `Promise.all` of two writes to a 3-round-trip chain (join -> count -> claim), plus a 4th on the leaderless-promotion path. The join legitimately has to come first, but the count and the request claim are independent of each other — neither reads the other's result, and both only require the join to have succeeded:

```ts
const [[memberCountRow], claimRequest] = await Promise.all([
  ctx.drizzle
    .select({ count: sql<number>`COUNT(*)`.mapWith(Number) })
    .from(userData)
    .where(eq(userData.anbuId, squad.id)),
  ctx.drizzle
    .update(userRequest)
    .set({ status: "ACCEPTED" })
    .where(and(
      eq(userRequest.id, input.id),
      eq(userRequest.type, "ANBU"),
      eq(userRequest.status, "PENDING"),
    )),
]);
```

The rollback then handles both together: if `claimRequest.rowsAffected === 0`, null out `anbuId` as today; if the count exceeds the cap, null out `anbuId` and put the request back to `PENDING` (themselves in a single `Promise.all`). That drops the happy path from 3 sequential round-trips to 2 without weakening any of the CAS guarantees.

app/src/app/anbu/[anbuid]/page.tsx
```diff
@@ -724,23 +743,40 @@ const AnbuOrders: React.FC<AnbuOrdersProps> = (props) => {
...
   // Query
-  const { data: requests } = api.anbu.getRequests.useQuery(undefined, {
-    enabled: !!squadId,
-    staleTime: 5000,
-  });
+  const { data: requests } = api.anbu.getRequests.useQuery(
+    { squadId },
+    {
+      enabled: !!squadId,
+      staleTime: 5000,
```
Comment (MathiasGruber) [inline app/src/app/anbu/[anbuid]/page.tsx:783]: Two things here:

**1. The query is ungated for viewers who can never get a non-empty result.** For an ordinary squad member (has `userAnbu`, not leader/kage/elder/staff) and for cross-village viewers, the server runs `fetchUpdatedUser` + the full `fetchSquad` relational query + `fetchRequests` and then returns `[]`. Note `fetchSquad` there duplicates work the page already did via `api.anbu.get.useQuery`. The gating conditions are pure prop derivations, so they can be computed above the hook:

```ts
const canManageRequests = isLeader || isKage || isElder || canStaffEdit;
const canJoinSquad = !isKage && !isElder && sameVillage;
const shouldFetch = canManageRequests || (!userAnbu && canJoinSquad);

const { data: requests } = api.anbu.getRequests.useQuery(
  { squadId },
  { enabled: !!squadId && shouldFetch, staleTime: 5000 },
);
```

This mirrors the server's own authorisation exactly, so nothing is lost. Move the `canManageRequests` / `canJoinSquad` definitions from the derived block up to the top and drop the duplicates, and change the loader guard to `if (!shouldFetch) return null;` before `if (!requests) return <Loader ... />;` so a disabled query doesn't render a permanent spinner.

**2. `onSuccess` (lines 785-786) awaits the two invalidations sequentially.** `getRequests.invalidate()` does not depend on `get.invalidate()`, so this serialises two refetch round-trips before the UI settles:

```ts
await Promise.all([
  utils.anbu.get.invalidate(),
  utils.anbu.getRequests.invalidate(),
]);
```

And since reject/cancel don't change squad membership, the `anbu.get` invalidation (the heavy members + orders payload) is wasted work on those two paths — a per-mutation `onSuccess` that only invalidates `getRequests` would avoid it.

app/src/server/api/routers/anbu.ts
```diff
@@ -123,27 +177,54 @@ export const anbuRouter = createTRPCRouter({
...
-      // Create
-      return { success: true, message: "User assigned to squad" };
+      // Leaderless squads stay joinable — route the request to the village kage so
+      // kage/elder/staff can accept (relatedId still identifies the squad).
+      const receiverId = squad.leaderId ?? user.village?.kageId;
+      if (!receiverId) {
+        return errorResponse("No leader or kage available to receive request");
+      }
+      // Mutate — persist squad id on relatedId so requests survive leader changes
+      await insertRequest(
```
MathiasGruber: `createRequest` has no guard against a user having several pending ANBU requests at once, and this PR removes the only thing that was preventing it.

Before, `getRequests` returned *all* of the caller's ANBU requests, so `hasPending` in `anbu/[anbuid]/page.tsx` hid the "Send Request" button on every squad page as soon as one request was pending anywhere. Now that `getRequests` is scoped by `squadId`, an applicant's pending request to squad A is filtered out on squad B's page (`isAnbuRequestForSquad` returns false for a different `relatedId`), so the button reappears and the user can queue a pending request against every squad in the village. Multiple leaders then each see a request they can accept; only the `isNull(userData.anbuId)` CAS in `acceptRequest` decides the winner, and the losers get a confusing "Requester already in a squad" error.

Suggest adding a server-side guard before inserting: fetch the caller's pending ANBU requests (`fetchRequests(ctx.drizzle, ["ANBU"], 3600 * 12, user.userId)` can go into the existing `Promise.all` above) and reject when one is still `PENDING`. That also closes the double-click duplicate-insert case.

app/src/server/api/routers/anbu.ts
MathiasGruber: The leader branch and the kage/elder/staff branch resolve the same set of requests by two different keys, and the leader one is the weaker of the two.

The leader path returns `ownSquadRequests`, derived from `fetchRequests(..., ctx.userId)` — rows where the leader is sender *or receiver*. Everything else in this PR treats `relatedId` as the authoritative squad link and `receiverId` as a best-effort pointer that gets repointed on promotion. Any time `receiverId` drifts from the current leader (see the leaderless-recovery block in `acceptRequest`, where the reassign runs even if the leader CAS loses), the leader silently sees zero pending requests for their own squad while the kage sees them fine.

Suggest having the leader branch use the same `fetchRequests(..., undefined, squadId)` query as the kage branch — the two branches then collapse into one `canManage` check, and `ownSquadRequests` / the `isAnbuRequestForSquad` fallback is only needed for the applicant path.

Separately: `ownRequests` (line 123) is fetched unconditionally but discarded in this kage/elder/staff branch, and lines 141-146 add two more request queries on top — 5 round-trips for that branch. `byRelatedId` and `byLeader` can be a single query with an `or(eq(relatedId, squadId), eq(receiverId, leaderId))` predicate.

app/src/server/api/routers/anbu.ts
```diff
@@ -163,45 +244,116 @@ export const anbuRouter = createTRPCRouter({
...
           .update(userData)
-          .set({ anbuId: squad.id })
-          .where(eq(userData.userId, requester.userId)),
-      ]);
+          .set({ anbuId: null })
+          .where(
+            and(eq(userData.userId, requester.userId), eq(userData.anbuId, squad.id)),
+          );
+        return errorResponse("Squad is full");
+      }
```
MathiasGruber: The join-then-count-then-rollback sequence makes concurrent accepts abort *each other* instead of picking a winner. With `ANBU_MAX_MEMBERS = 4`, take a squad at 3 members and two managers accepting different requests at the same time: both pass the `nMembers >= ANBU_MAX_MEMBERS` pre-check, both CAS-join (count is now 5), both then read `count = 5 > 4` and both roll themselves back. The squad drops back to 3, both callers get "Squad is full", and a free seat is left unused.

A deterministic tie-break avoids the mutual abort — e.g. after joining, select the squad's member ids in a stable order and only roll back when this requester falls outside the first `ANBU_MAX_MEMBERS` positions. At minimum, re-read the count after rolling back and retry the join once if it dropped back below the cap.

app/src/server/api/routers/anbu.ts
```diff
@@ -163,45 +244,116 @@ export const anbuRouter = createTRPCRouter({
...
+            .update(anbuSquad)
+            .set({ leaderId: requester.userId })
+            .where(and(eq(anbuSquad.id, squad.id), isNull(anbuSquad.leaderId))),
+          reassignPendingAnbuRequestsOnPromotion(
+            ctx.drizzle,
+            squad.id,
+            null, // squad is leaderless; avoid touching other squads' relatedId=null requests
+            requester.userId,
+          ),
+        ]);
```
MathiasGruber: The leader write is CAS-guarded (`isNull(anbuSquad.leaderId)`) but `reassignPendingAnbuRequestsOnPromotion` runs in parallel with it unconditionally, so it fires even when the CAS matched 0 rows.

If two accepts race on a leaderless squad, the loser's CAS is a no-op but its reassign still rewrites `receiverId` on every pending request for the squad to point at the *loser*. The actual new leader then can't see those requests (the `isLeader` branch of `getRequests` looks them up by sender/receiver), while the loser — a plain member — becomes the receiver of requests they have no permission to act on.

Await the `anbuSquad` update first and only call the reassign when `rowsAffected === 1`.

app/src/server/api/routers/anbu.ts
```diff
@@ -722,18 +894,33 @@ export const removeFromSquad = async (
...
+    mutations.push(
+      client
+        .update(anbuSquad)
+        .set({ leaderId: nextLeaderId })
+        .where(eq(anbuSquad.id, squad.id)),
+      // Always anchor relatedId to this squad (even with no successor) so legacy
+      // rows cannot later match a different squad via receiverId === leaderId.
+      reassignPendingAnbuRequestsOnPromotion(client, squad.id, userId, nextLeaderId),
+    );
+  }
```
MathiasGruber: The leadership transfer has no compare-and-swap on the outgoing leader, and the successor comes from the `squad.members` snapshot without re-verifying they are still a member.

Concretely: leader L is kicked while member M (the chosen successor) calls `leaveSquad` concurrently. M's `removeFromSquad` clears `M.anbuId`; L's call then writes `leaderId = M` unconditionally. The squad ends up with a `leaderId` pointing at a user whose `anbuId` is null. That user now passes `isLeader` in `getConvenienceStatus`, so they can `acceptRequest` / `rejectRequest` / `upsertNotice(LEADER)` for a squad they are not in (`editSquad` and the espionage upgrades happen to block them via `user.anbuId !== squad.id`; the request endpoints do not).

Suggest guarding the update with `and(eq(anbuSquad.id, squad.id), eq(anbuSquad.leaderId, userId))` so a concurrent promotion isn't clobbered, and verifying the successor still has `anbuId === squad.id` before installing them (falling back to `null` otherwise). The same unguarded `.where(eq(anbuSquad.id, squad.id))` is used in `promoteMember` around line 554.

app/drizzle/migrations/0027_backfill_anbu_request_related_id.sql
```diff
@@ -0,0 +1,16 @@
...
+-- current anbuId — a demoted/kicked leader who later joined a different squad
+-- would incorrectly re-link the original squad's requests. Remaining unmatched
+-- rows (relatedId still null) are covered by getRequests' legacy receiverId ===
+-- leaderId fallback while that leader still leads the squad.
+
+UPDATE `UserRequest` AS ur
+INNER JOIN `AnbuSquad` AS a ON a.leaderId = ur.receiverId
+SET ur.relatedId = a.id
+WHERE ur.type = 'ANBU'
+  AND ur.relatedId IS NULL;
```
MathiasGruber, line 18: This backfill isn't restricted to the rows that need it, so it permanently rewrites history on a heuristic.

The `WHERE` only filters `type = 'ANBU' AND relatedId IS NULL`, so every ACCEPTED / REJECTED / CANCELLED row ever created also gets stamped — including rows belonging to a squad the receiver used to lead before joining or being promoted in a different one, which then get mis-attributed to their *current* squad. The runtime code only consults `relatedId` for pending requests inside a 12h window, so those older rows gain nothing from the update and can only be wrong.

Add `AND ur.status = 'PENDING'` and, given `getRequests` caps at 12h, `AND ur.createdAt > NOW() - INTERVAL 1 DAY`. That also keeps the write small — `UserRequest` accumulates every spar/clan/anbu challenge ever created, and this is an unbatched multi-table UPDATE against the whole table.

app/tests/server/api/anbu.test.ts
```diff
@@ -0,0 +1,261 @@
...
+  if ("value" in obj && !("queryChunks" in obj) && !Array.isArray(obj.value)) {
+    return `param(${JSON.stringify(obj.value)})`;
+  }
+  if ("queryChunks" in obj && Array.isArray(obj.queryChunks)) {
+    return (obj.queryChunks as unknown[]).map(describeSql).join("");
+  }
+  return Object.prototype.toString.call(node);
+}
+
+describe("ANBU request squad identity", () => {
```
MathiasGruber, line 94: This suite only covers three pure helpers (`isAnbuRequestForSquad`, `fetchSquadForAnbuRequest`, `reassignPendingAnbuRequestsOnPromotion`). None of the behaviour the PR is actually about is exercised:

- the new authorization branches on `acceptRequest` / `rejectRequest` (leader vs. kage-of-squad-village vs. elder vs. `canEditClans`), in particular that a kage of a *different* village is rejected;
- the double-accept CAS on `userRequest.status` and the `isNull(userData.anbuId)` join CAS, including both rollback paths;
- the member-cap rollback;
- the `getRequests` permission branches (a plain squad member must get `[]`; an applicant must only get their own rows).

Those are the paths where a regression is costly. The routers are already fully mocked here, so driving `anbuRouter.acceptRequest` with a stubbed drizzle client and asserting on the `rowsAffected` handling should be a small extension of the existing setup.

MathiasGruber (approved):

Issue #664, opened 2025-08-15
Sage Mode


theeneon:
Sage Mode is similar to Bloodlines but the only caveat is that Sage Mode needs to be activated by the player in combat for it to be active. Once active the effects of sage mode can't be cleared. 

- [ ] Sage Mode will be created through the content panel on the info page. Will need the option to set rounds for how long sage mode is active for. Sage Mode becomes active the same turn it's used. Sage Mode will need a special feature after activation. 
- [ ] Sage Mode will have two Levels. (This is to allow players to play the game complete quest to unlock their full powers for Sage Mode instead of being given everything from the get go. 
- [ ] Add Sage Mastery  to player profile as a stat
- [ ] Add Sage Mastery experience to Event creation 
- [ ] Add Required Sage Mode to Event Creation. 
- [ ] After Activation: This is basically a set of tags that will take effect after the sage mode expires and how long they will last for. 

- [ ] Once a player acquires Sage Mode, a new basic action option will show up for them called Activation. The image of the activation will be taken from the Icon image from when the Sage Mode is created. 

- [ ] Sage Mode effects can't be cleared or copied or mirrored. Similar to Bloodlines.

MathiasGruber:
@claude please create an implementation plan for this feature, and be sure to post the full implementation plan into this issue, including any potential follow up questions needed for clarity.

MathiasGruber:
@theeneon does this plan look reasonable to you? 

Also, can you help clarify wrt. to the 7 follow-up questions?

theeneon:
> [@theeneon](https://github.com/theeneon) does this plan look reasonable to you?
> 
> Also, can you help clarify wrt. to the 7 follow-up questions?


1. Activation cost will be Chakra and Stamina
2. Sage mode is a combination of given through Random roll, similar to bloodlines, can be attached to an item to be given to the player, and achievement-based unlocks. Automatically when reaching the mastery Threshold, the 2nd level is unlocked. 
3. Sage Mode can be used in PVP,PVE,Quest,Arena but not in Rank. 
4. Activation is once per battle. 
5. Can only have one Sage mode at a time, they are not able to collect and store multiple. 
6. Different visuals 
7. After effects refers to debuffs like exhaustion after using sage mode. 

Other than the plan looks good.

theeneon:
@claude
 Activation cost will be Chakra and Stamina
[...]

PR #1402, opened 2026-07-19
Feature/sage mode 664

ifroshty:
# Sage Mode (#664)

This PR introduces **Sage Mode**, a new combat-transformation system with its own staff-editable content catalog, a per-user **Sage Mastery** progression stat, and full acquisition/removal, quest, and admin integration.

An earlier iteration of this branch included a pity system, a D/C/B/A/S catalog rank, and player-facing swapping. Those were all **removed** (commit `feat: remove pity, ranks, and swap`) and are no longer part of the feature — see Breaking changes below.

### Core mechanics

- **Sage Mode is a combat transformation** a player activates once per battle. Activation is offered inline with the basic actions via an injected activation jutsu (`SAGE_MODE_ACTIVATION_JUTSU_ID`) that is built into battle state at effect-injection time — it is not a seeded DB row. It targets SELF and carries a single `activatesagemode` effect.
- **Active window + after-effects.** On activation the mode applies its configured `effects` for `activationRounds` (default 5). When that window elapses the round pipeline transitions to the mode's `afterEffects` for `afterEffectRounds` (default 3). Both durations are per-mode, staff-editable columns.
- **Costs.** Activation charges a flat chakra + stamina cost derived from the mode's `chakraCostPerc` / `staminaCostPerc` against the user's max pools (`getSageModeActivationCost`); an unaffordable activation is never offered. The mode's AP/action cost rides on the action's `actionCostPerc` (per-mode, default `SAGE_MODE_DEFAULT_ACTION_COST_PERC` = 80) and is gated by the normal action pipeline.
- **Once per battle.** Activation is blocked after `sageModeUsedThisBattle` is set; the flag is set on activation, re-asserted on the after-effect transition, and never cleared until the battle ends. The activation/expiry bookkeeping fields (`sageModeActivated`, `sageModeActivatedRound`, `sageModeExpiresRound`, `sageModeUsedThisBattle`) are transient battle-state fields, not persisted schema columns.
- **Two levels.** `SAGE_MODE_MAX_LEVEL` = 2. Level 2 is not a stored value — it auto-unlocks for a mode when the user's `sageMasteryExperience` meets that mode's `requiredSageMastery` threshold (`getActiveSageLevel`). `requiredSageMastery = 0` disables level 2 for that mode entirely. At level 2 the mode's `level2Effects` are merged with its base effects.
- **Protected effects.** Sage effects (`fromType` `sageMode` / `sageModeAfter`) are excluded from clear/cleanse (`persistentEffectSourceTypes`) and from both `copy` and `mirror` (`excludedFromTypes`), so they cannot be stripped or stolen.
- **Battle-type scoping.** Sage Mode is disabled in `RANKED_PVP` and `RANKED_SPARRING` (`SAGE_MODE_DISABLED_BATTLES`). All other battle types (including non-ranked sparring and arena) permit it. Enforced both at the offer gate and the activation guard.

### Sage Mastery (progression stat)

- A first-class user stat backed by `userData.sageMasteryExperience` (int, default 0), shown on the profile page (`Sage: <rank>`) and in `StrengthWeaknesses`.
- Rank ladder `SAGE_MASTERY_RANKS` = `NONE / INITIATE / ADEPT / MASTER / LEGENDARY`, with thresholds `SAGE_MASTERY_REQUIRED_EXP` (ADEPT 150000 / MASTER 250000 / LEGENDARY 450000) and cap `SAGE_MASTERY_EXP_CAP` = 450000.
- **Daily activation cap scales with mastery.** `getSageDailyCap` maps rank → `SAGE_MASTERY_DAILY_ACTIVATIONS` (NONE/INITIATE 10, ADEPT 12, MASTER 15, LEGENDARY 20). Enforced via the `userData.dailySageActivations` counter (incremented on activation, reset to 0 by the daily-counters cron).
- Mastery gates the per-mode level-2 unlock and quest availability (see below).

### Acquisition & removal

- **Item roll (`ITEM`).** The `rollsagemode` item effect performs an independent probabilistic roll (`roll < effect.power`). On success it inserts a `SageModeRolls` row of type `ITEM` and equips a randomly chosen eligible mode. Blocked if the user already has a sage mode. A failed roll records nothing.
- **Quest reward (`QUEST`).** `reward_sage_modes` rolls and records a not-yet-owned mode only when the player has no equipped sage mode. Already-equipped players skip the grant so a mode is not burned into history they can never wear. `COALESCE` still protects an empty-slot race.
- **Staff direct-assign.** Staff with `canEditBloodline` can set a user's `sageModeId` through the user-edit flow (no roll-history row written).
- **Removal is a flat reputation cost.** `removeSageMode` requires `reputationPoints >= REMOVAL_COST` (5) for every removal, guarded by a compare-and-swap on `sageModeId` + the reputation predicate. There is no free-removal path and no roll-type inspection.
- There is **no player purchase path, no swap procedure, and no pity/natural-roll path** for sage mode. `SAGE_MODE_ROLL_TYPES` = `ITEM | QUEST` only.

### Quest integration

- Rewards: `reward_sage_modes` (string array) and `reward_sage_mastery_experience` (number), aggregated/scaled in `libs/quest.ts`.
- Gating: quest fields `requiredSageModeId` and `requiredSageRank` (enum over `SAGE_MASTERY_RANKS`), enforced in quest availability via `sageModeCheck` (exact mode) and `sageRankCheck` (minimum mastery).
- Quest/event editor exposes **Required Sage Mode** and **Required Sage Rank** fields plus the two reward fields.

### Admin / content management

- New content type with a manual list page and editor. Editable fields: name, activation message (`battleDescription`, `%user` templated), image, roll-pool `level`, **Tier 2 Unlock** (a mastery-experience threshold dropdown mapping to `requiredSageMastery`), active/after-effect durations, chakra/stamina/AP cost percentages, `hidden`, `villageId`, and description. The three effect arrays (`effects`, `afterEffects`, `level2Effects`) are edited in separate effect panels.
- The editor exposes **no catalog rank field** — sage-mode content has no letter rank; gating is done entirely through the integer `requiredSageMastery` threshold and the roll-pool `level`.

## Why

Sage Mode adds a high-impact, staff-tunable combat transformation with a long-tail progression hook (Sage Mastery) that gates both power (Tier 2) and frequency (daily cap). Making mastery a first-class stat and driving level-2 off a per-mode experience threshold lets staff tune each mode's ceiling independently, while the mastery-scaled daily cap and the once-per-battle rule keep it from dominating. Excluding sage effects from clear/copy/mirror keeps a transformation from being trivially neutralized.

## Implementation notes

- The after-effect transition (`applySageModeAfterRoundTransition`) is invoked from the combat router at the three `progressRound` sites rather than from a client mutation, deliberately, to avoid a `util` ↔ `process` circular import.
- Sage concurrency safety is the `sageModeId` compare-and-swap in `updateSageMode` (verifies `rowsAffected`), not any pity accounting — the pity/reserve machinery in `concurrency.ts` is used exclusively by bloodlines.
- The bloodline pity system (and its `BloodlineRolls.pityRolls` / `NATURAL` roll type) is a separate, still-live system and is untouched by sage code.
- This round's fixes: **G1** — the staff "All Visibility" content filter now returns hidden and visible modes together (previously it excluded one set); **CR3** — `getAll` cursor/limit inputs are hardened to non-negative integers.

## Files changed

The large insertion count is dominated by a single generated file, `app/drizzle/migrations/meta/0031_snapshot.json`; hand-written sage code is a few thousand lines. Grouped:

- **DB / Drizzle (5):** `schema.ts`, `constants.ts`, the single sage migration `0031_sour_dakota_north.sql` + its `meta/0031_snapshot.json`, and `_journal.json`.
- **Server / API (10):** new `routers/sageMode.ts`; edits to `bloodline.ts`, `quests.ts`, `item.ts`, `combat.ts`, `profile.ts`, `root.ts`, `server/utils/concurrency.ts`, and the two cron routes (`daily-counters`, `monthly-mcp-reset`).
- **libs (12):** new `libs/sageMode.ts` and `libs/combat/sageModeActivationJutsu.ts`; edits to `combat/process.ts`, `combat/actions.ts`, `combat/{constants,database,tags,types,util}.ts`, `libs/item.ts`, `libs/objectives.ts`, `libs/quest.ts`.
- **validators (6):** new `validators/sageMode.ts`; edits to `validators/{combat,objectives,quest,rewards,user}.ts`.
- **UI (20):** new manual pages `manual/sageMode/page.tsx` and `manual/sageMode/edit/[sagemodeid]/page.tsx`; new hook `hooks/sageMode.ts`; new layout components `layout/SageMode.tsx` and `layout/SageModeFiltering.tsx`; plus edits across profile/blackmarket/science/manual pages, `hooks/{profile,quest}.ts`, and several layout components.
- **tests (12):** see Testing.
- **assets (1):** `assets/manual/Sage.webp`.

## Database changes

**New tables**

- `SageMode` — the content catalog. Key columns: `name` (unique), `image`, `description`, nullable `battleDescription`, JSON `effects` / `afterEffects` / `level2Effects` (default `[]`), `activationRounds` (default 5), `afterEffectRounds` (default 3), `chakraCostPerc` / `staminaCostPerc` (default 20), `actionCostPerc` (default 80), `level` (default 1), `requiredSageMastery` (int, default 0), `hidden` (default false), nullable `villageId`, timestamps. Indexes on `level`, `villageId`, `hidden`. **No catalog rank column** — gating is the integer `requiredSageMastery`.
- `SageModeRolls` — roll provenance. Columns: `id`, `userId`, nullable `sageModeId`, `type` (enum `ITEM | QUEST`), timestamps. Indexes on `userId`, `sageModeId`. No pity/natural/rank columns.

**New `userData` columns**

- `sageModeId` (nullable, with index and a `sageMode` relation), `sageMasteryExperience` (int, default 0), `dailySageActivations` (smallint unsigned, default 0).

**Quest columns**

- `requiredSageModeId` (nullable) and `requiredSageRank` (enum over `SAGE_MASTERY_RANKS`).

**Migration**

- A single new migration on this branch: `0031_sour_dakota_north.sql` (+ its snapshot). It creates the two tables, adds the `userData` and `Quest` columns, and adds the associated indexes. It does **not** change `BloodlineRolls.pityRolls` (that column stays `tinyint`).

## Breaking changes

- **Pity, catalog ranks, and swapping were removed.** There is no sage pity system, no D/C/B/A/S catalog rank on sage content, and no player-facing swap procedure. To change modes a player must `removeSageMode` (flat 5 reputation) and re-acquire via an item roll or quest reward. (The mastery ladder is a separate, retained user stat.)
- No changes to existing combat data are required; battles already in flight are safe (sage bookkeeping is transient battle state).

## Testing

- Sage-specific suites (all new): `tests/libs/combat/sage_mode.test.ts`, `sage_daily_cap.test.ts`, `sage_activation_ap_cost.test.ts`, `tests/libs/sageMode.test.ts`, `tests/libs/quest_available_sage_mode.test.ts`, `tests/validators/sageMode.test.ts`.
- Adjacent suites this branch adds/touches: `tests/server/utils/pityReservation.test.ts`, `tests/server/utils/bloodlineEconomy.test.ts`, `tests/libs/objectives_reward.test.ts`, `tests/validators/objectives.test.ts`, `tests/validators/objective_tag_types.test.ts`, and the shared `tests/libs/combat/helpers/battleScenario.ts` harness.
- Coverage includes activation gating (battle type, once-per-battle, daily cap, affordability), the active→after-effect transition, the mastery daily cap by rank, level-2 unlock by experience threshold, quest availability gating, and validator shape. This round additionally verifies the **G1** all-visibility content filter and the **CR3** `getAll` cursor/limit hardening.

## License

## Summary by CodeRabbit

* **New Features**
  * Added Sage Mode progression, mastery ranks, activation limits, and combat effects (including exhaustion/after-effects and daily caps).
  * Added Sage Mode rolls, quest gating/rewards, and new profile/stat displays.
  * Added manual management for Sage Modes (list/create/edit/delete) plus filtering.
* **Bug Fixes**
  * Secured monthly reset endpoints with cron authorization and now reset Sage Mode state correctly.
  * Daily Sage Mode activation counters reset properly; improved combat UI/consumption flows to prevent incorrect/duplicate Sage Mode grants.


app/src/server/api/routers/sageMode.ts
```diff
@@ -0,0 +1,321 @@
...
+  repCost: number,
+  logMsg: string,
+) => {
+  const updateResult = await client
+    .update(userData)
+    .set({
+      sageModeId: mode?.id || null,
+      reputationPoints: sql`${userData.reputationPoints} - ${repCost}`,
+      // Advance the whole-user version so a concurrent claimUserSnapshot CAS detects this write.
+      updatedAt: new Date(),
```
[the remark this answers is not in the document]
ifroshty: The `update` mutation's `sageMode` update and `actionLog` insert now run in `Promise.all`, matching
the sibling `delete` mutation. Left the `updateSageMode` **helper** sequential deliberately: its
`actionLog` insert is gated behind the CAS `rowsAffected` throw, so parallelising it would log a
change that didn't happen.

app/drizzle/migrations/0031_sour_dakota_north.sql
```diff
@@ -0,0 +1,47 @@
+CREATE TABLE `SageMode` (
```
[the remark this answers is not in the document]
ifroshty: The collision is real and must be resolved before merge, but the literal "rename to 0020, keep the
timestamp" is **incomplete and premature** for two independent reasons:

1. **Timestamp gating.** This repo's drizzle migrator gates on the journal `when` (`created_at <
   folderMillis`), not filename/idx/hash (verified in `drizzle-orm/mysql-core/dialect.js`). This
   branch's `0019_striped_runaways` has `when 1784365621925`; main's `0019_even_vision` has
   `when 1784395621841` (~8h later). Renaming to `0020` while keeping the old `when` sorts it
   *before* even_vision, so on any DB that already applied even_vision the migrator treats it as
   already-applied and **skips it → sage tables never created.**
2. **Snapshot lineage.** even_vision isn't in this branch yet, and both `0019_snapshot.json` files
   share `prevId → 0018` (a genuine fork). A hand-rename leaves `prevId` pointing at 0018 instead of
   main's real 0019, corrupting drizzle-kit's schema-diff chain for future `generate`.

Correct resolution (at rebase): rebase onto main (even_vision becomes 0019), then **regenerate** the
sage migration as 0020 via `make makemigrations` so it gets a fresh `when` (> even_vision) and a
correct `prevId`. Note: preview/dev DBs that already ran the old 0019 will already have the sage
tables, so the regenerated 0020 must be idempotent (or those DBs reset). No CI target runs
`drizzle-kit migrate` (only `dbpush`), so there's no automatic-break risk, but the chain must still
be correct for anyone running `migrate` manually or future `makemigrations` diffs.

MathiasGruber:
## Rebased on latest `main`

This branch had merge conflicts against `main` and has been rebased (force-pushed). Conflicts were limited — the bulk of the branch applied cleanly.

### How the conflicts were resolved

**`app/src/server/api/root.ts`** — both sides added routers to the same import block and router map. Took the union: `main`'s new routers plus `sageModeRouter`.

### Migrations

Both this branch and `main` had claimed `0019`. Per the usual process the branch migration was deleted and regenerated on top of `main`: `main`'s `0019_even_vision` is restored untouched, and this branch's schema now lands as `0020_fixed_chimera` (`SageMode` / `SageModeRolls` tables, sage columns on `UserData` and `Quest`).

One manual correction to the generated file: `drizzle-kit` emitted

```sql
ALTER TABLE `BloodlineRolls` MODIFY COLUMN `pityRolls` smallint NOT NULL;
```

dropping the `DEFAULT 0`, even though both `schema.ts` and the generated `0020_snapshot.json` declare `.default(0)`. In MySQL a `MODIFY COLUMN` without a `DEFAULT` clause drops the existing column default, so this would have silently diverged from the schema. Restored to `... smallint NOT NULL DEFAULT 0`, matching what the branch's original migration emitted.

### Verification

`make typecheck`, `make lint` and `make test` all pass locally (503 tests, 0 failures).

app/src/validators/combat.ts
```diff
@@ -1138,12 +1158,13 @@ const SuperRefineJutsu = (
...
   const hasBloodlineRoll = data.effects.find((e) => e.type === "rollbloodline");
+  const hasSageModeRoll = data.effects.find((e) => e.type === "rollsagemode");
   const hasRemoveBloodline = data.effects.find((e) => e.type === "removebloodline");
   const hasNonCombatConsumeReward = data.effects.find(
     (e) => e.type === "noncombatconsumereward",
   );
-  if (hasBloodlineRoll || hasRemoveBloodline) {
-    addIssue(ctx, "Cannot have bloodline add/remove effects on jutsu");
+  if (hasBloodlineRoll || hasSageModeRoll || hasRemoveBloodline) {
+    addIssue(ctx, "Cannot have bloodline/sage add/remove effects on jutsu");
```
MathiasGruber, line 1205: `SuperRefineJutsu` blocks `rollbloodline`, `rollsagemode`, and `removebloodline` on jutsu, but does not block `activatesagemode`. In-battle sage activation is meant to happen only via the injected `SAGE_MODE_ACTIVATION_JUTSU_FALLBACK` (its `actionCostPerc` is overridden to the equipped mode's cost at `app/src/libs/combat/actions.ts:207-216`, keyed strictly on `SAGE_MODE_ACTIVATION_JUTSU_ID`). Any other jutsu carrying the `activatesagemode` effect would still succeed through `applyActivateSageMode` (`app/src/libs/combat/process.ts:1791-1858` — no check that the action is the fallback), and its own arbitrary `actionCostPerc` would apply instead of the sage mode's, giving staff a way to accidentally publish a cheaper activation route. Add `activatesagemode` to this guard (and mirror it in `SuperRefineItem` at line 1120 for items) so only the fallback carries this effect. Flagged by greptile in the initial review summary.

MathiasGruber (commented):
Re-opening previously raised feedback that still applies to the current code.

app/src/server/api/routers/sageMode.ts
```diff
@@ -0,0 +1,320 @@
...
+    });
+  }),
+
+  /** All sage modes (including hidden) for staff editing user data — same permission as bloodline edit. */
+  getAllNamesForEdit: protectedProcedure.query(async ({ ctx }) => {
+    const user = await fetchUser(ctx.drizzle, ctx.userId);
+    if (!user || !canEditBloodline(user.role)) {
+      throw serverError("FORBIDDEN", "Not allowed to list sage modes for editing");
+    }
+    return await ctx.drizzle.query.sageMode.findMany({
```
MathiasGruber: Sequential `await fetchUser` (line 41) followed by `await ...sageMode.findMany` here — the `findMany` has no filter, join, or condition that depends on `user`, so it should run in parallel with `fetchUser`:

```ts
const [user, results] = await Promise.all([
  ctx.drizzle.query.sageMode.findMany({
    columns: { id: true, name: true, image: true },
    orderBy: (table, { asc }) => [asc(table.name)],
  }),
]);
if (!user || !canEditBloodline(user.role)) throw serverError("FORBIDDEN", "Not allowed to list sage modes for editing");
return results;
```

Per CLAUDE.md: "Latency matters more than bandwidth" and "prefer parallel queries even if it means fetching slightly more data". The wasted findMany on a rejected request is a small content table; the common caller here is staff editing, so parallelizing removes a full round-trip from the normal path.

MathiasGruber (commented):
Automated performance findings — please address.

app/src/validators/combat.ts
ifroshty: Added `activatesagemode` to both content guards — `SuperRefineJutsu` (jutsu) and `SuperRefineItem`
(items) now reject it on save with a dedicated message ("…injected automatically in battle").

Verified the premise before applying: the cost override at `actions.ts:207-216` is keyed strictly on
`SAGE_MODE_ACTIVATION_JUTSU_ID`, while `applyActivateSageMode` (`process.ts:1182`) runs on any
`activatesagemode` effect with no fallback check — so a non-fallback carrier would activate with its
own arbitrary `actionCostPerc`. Also confirmed `SAGE_MODE_ACTIVATION_JUTSU_FALLBACK`
(`sageModeActivationJutsu.ts`) is a plain `Jutsu` object constant, never parsed through
`JutsuValidator`, so the new guards block staff-authored content only and don't touch the injected
action.

app/src/server/api/routers/sageMode.ts
ifroshty: Parallelized. `fetchUser` and `sageMode.findMany` now run in one `Promise.all`; the
`canEditBloodline` guard is applied after and `results` returned. Removes the sequential round-trip
from the normal staff path (the wasted findMany on a rejected request is the accepted tradeoff you
noted).

app/src/libs/combat/process.ts
```diff
@@ -1757,3 +1784,228 @@ export const applyDamageModifierPipelineToConsequences = ({
...
+  const activeLevel = getActiveSageLevel(newTarget.sageMasteryExperience, sageMode);
+  // At level 2, level2Effects apply IN ADDITION to base effects; sort together so
+  // ordering-sensitive tags (shields before damage, etc.) interleave correctly.
+  const activeEffects =
+    activeLevel >= SAGE_MODE_MAX_LEVEL
+      ? [...sageMode.effects, ...(sageMode.level2Effects ?? [])]
+      : sageMode.effects;
+  const sorted = [...activeEffects].sort((a, b) =>
+    sortEffects(a as UserEffect, b as UserEffect),
+  );
```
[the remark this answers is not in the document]
ifroshty: Not reproducible — the counter is persisted atomically at battle end, not just in-memory. The claim
missed the battle-end save path in `database.ts`.

The atomic increment already exists in `updateUser` (`database.ts:1140`, the only battle-end save
path, called from `combat.ts:327` and `:750`):

```ts
dailySageActivations: sql`dailySageActivations + ${user.sageModeUsedThisBattle ? 1 : 0}`, // database.ts:1456
```

`user` is the same in-memory battle-state object (`curBattle.usersState.find(...)`) that set
`sageModeUsedThisBattle = true` on activation. Sage mode is once-per-battle (guard in
`applyActivateSageMode`, `process.ts`), so the single in-memory `+1` maps to exactly one DB `+1`.

The described exploit ("end the battle, start a fresh one, activate again") is structurally
impossible, not merely usually-handled:

- That increment lives in the **same** `UPDATE userData` statement that sets `battleId = null` and
  `status = 'AWAKE'/'HOSPITALIZED'` (`database.ts:1456-1493`).
- `initiateBattle` (`combat.ts`) CAS-guards claiming a new battle on
  `status IN ('AWAKE','QUEUED','KAGE_QUEUED')` and rolls back if `rowsAffected` doesn't match.
- Therefore you cannot reach a new battle without that single statement — counter increment
  included — having already committed. Flee/abandon paths flush too (`calcBattleResult` returns
  non-null on `!stillInBattle`, not only on win/loss).

Both `sageModeUsedThisBattle` and `dailySageActivations` are in `publicState`
(`constants.ts:59-61`), so they round-trip through the battle state from initiation (loaded from
`userData`) to the save. Reset is handled by the daily-counters cron
(`daily-counters/route.ts:39`) and the monthly reset.

Optional FYI (not this issue, not a fix): the whole reward-flush shares a systemic partial-failure
race — `updateBattle` deletes the battle row before the `Promise.all` that contains `updateUser`, so
a sibling-promise throw could drop the batch on retry. It's not sage-specific, fails safe
(under-counts, no exploit), and is a known consequence of PlanetScale's no-transactions constraint.

app/drizzle/constants.ts
```diff
@@ -1977,6 +2020,8 @@ export const IMG_MANUAL_TRAVEL =
   "https://uploadthing.b-cdn.net/f/Hzww9EQvYURJu1h1uHCyJLoOFkrcn4gxSwCfEQ9eMNXZlG8b.webp";
 export const IMG_MANUAL_BLOODLINE =
   "https://uploadthing.b-cdn.net/f/Hzww9EQvYURJaCMo8gYYfKMcJ2B5EmWt6VsNgqxpG8OSXAQk.webp";
+export const IMG_MANUAL_SAGE_MODE =
+  "https://uploadthing.b-cdn.net/f/Hzww9EQvYURJunUUX2bCyJLoOFkrcn4gxSwCfEQ9eMNXZlG8";
 export const IMG_MANUAL_JUTSU =
   "https://uploadthing.b-cdn.net/f/Hzww9EQvYURJMI7fE4tsO4cexqW2RDgkE3zZbNXSFGitmnar.webp";
```
MathiasGruber, line 2096: The greptile summary flagged this and it's still unaddressed: `IMG_MANUAL_SAGE_MODE` is the only `IMG_MANUAL_*` constant missing its `.webp` extension. Since commit d93062816 ("Serve all uploads through extension-bearing URLs for CDN optimization"), the Bunny CDN optimizer only engages when the served URL path carries a file extension — so this image is served at full size on every manual page load, ignoring any `?width=` param. Please re-upload the asset with a `.webp` customId and update the URL to match the pattern of the surrounding constants.

app/src/libs/combat/process.ts
```diff
@@ -1757,3 +1784,228 @@ export const applyDamageModifierPipelineToConsequences = ({
...
+      battle,
+    });
+    realized.longitude = newTarget.longitude;
+    realized.latitude = newTarget.latitude;
+    realized.fromType = "sageMode";
+    realized.targetId = newTarget.userId;
+    // Ignore per-effect `rounds` in the JSON for duration; use only `activationRounds` above.
+    // Exception: `rounds === 0` means an instant / one-shot tag — leave unchanged.
+    if (realized.rounds !== 0) {
+      realized.rounds = activeDurationRounds;
```
MathiasGruber: The sage mode router deletes `rounds` from `effects`/`afterEffects`/`level2Effects` before storing them (sageMode.ts:193/198/203), so when the stored tag reaches `realizeTag` here (line 1872) the `"rounds" in tag` check is false and `tag.timeTracker` is never initialized. Then this line sets `realized.rounds = activeDurationRounds` after the fact, leaving the effect with `rounds > 0` but `timeTracker === undefined`. The per-round dedup guard in `getEffectApplyRatio` (util.ts:1011) is `effect.timeTracker` gated, so it is silently skipped for every sage effect — any tag type not in the `alwaysApply` list will apply multiply per round with no dedup. The same pattern exists in `applySageModeAfterRoundTransition` at line 1982. Suggest initializing `realized.timeTracker = {}` alongside `realized.rounds = activeDurationRounds` in both branches (or having the sage router preserve `rounds` through storage so `realizeTag`'s own init path fires).

MathiasGruber (commented):
Verifying previously raised feedback that appears not to be fully addressed.

app/src/server/api/routers/bloodline.ts
```diff
@@ -914,32 +916,54 @@ export const bloodlineRouter = createTRPCRouter({
...
+        );
+      } catch (error) {
+        if (error instanceof BloodlineGrantRejectedError) {
+          await refundPityCredit({
+            client: ctx.drizzle,
+            table: bloodlineRolls,
+            rollId: prevRoll.id,
+          });
+        }
+        throw error;
```
MathiasGruber, line 967: If `refundPityCredit` throws (e.g. transient DB error on the decrement), the exception propagates out of the catch block before line 956 executes, so the original `BloodlineGrantRejectedError` is masked and the caller sees the refund's error instead. That's a downgrade in observability: the real failure was the grant CAS, not the refund. Consider wrapping the refund call in its own try/catch that logs and swallows, then re-throws `error`, so the client always gets the meaningful message.

```ts
try {
  await refundPityCredit({...});
} catch (refundErr) {
  console.error("Pity credit refund failed", refundErr);
}
throw error;
```

app/src/server/api/routers/combat.ts
```diff
@@ -1093,6 +1115,9 @@ export const combatRouter = createTRPCRouter({
       const bloodline = user.bloodlineId
         ? userBattle.extraState.bloodlines?.[user.bloodlineId]
         : null;
+      const sageMode = user.sageModeId
```
MathiasGruber: This `const sageMode` shadows the `sageMode` schema table imported at line 82. Today it's harmless because the schema import is only referenced inside `initiateBattle` (a different function scope), but a future edit that wants to reference the table here would silently pick up the local `SageMode | null` and produce a confusing type error. Rename either the local (e.g. `equippedSageMode`, mirroring names used elsewhere in this file) or the import alias to avoid the ambiguity.

app/src/server/utils/concurrency.ts
```diff
@@ -196,3 +196,54 @@ export const removeBloodlineFromPoolAtomically = async ({
...
+ * so `pityRolls` returns to the count the caller read without underflowing.
+ */
+export const refundPityCredit = async ({
+  client,
+  table,
+  rollId,
+}: RefundPityCreditParams) => {
+  await client
+    .update(table)
+    .set({ pityRolls: sql`${table.pityRolls} - 1`, updatedAt: new Date() })
```
MathiasGruber, line 337: `refundPityCredit` decrements unconditionally with no CAS/guard. The docstring contract ("invoke only when the grant CAS rejected") makes this safe today, but a future caller that misuses it — or a retry loop that fires it twice for the same reservation — will push `pityRolls` below zero without any error. Since `pityRolls` was widened from `tinyint` to `smallint` (signed) in this PR, negatives now silently store rather than error out. A `WHERE pityRolls > 0` guard would make the second call a no-op instead of silently corrupting the counter.

MathiasGruber (commented):
Logic / readability review of the sage mode PR. Overall the change is thoroughly commented and well-tested; a few small observations inline.

app/drizzle/constants.ts
ifroshty: Confirmed and unresolved — flagging rather than silently applying the literal edit, because appending
`.webp` to the current URL breaks the image outright:

```
GET .../Hzww9EQvYURJunUUX2bCyJLoOFkrcn4gxSwCfEQ9eMNXZlG8        → 200 image/webp (15932 bytes)
GET .../Hzww9EQvYURJunUUX2bCyJLoOFkrcn4gxSwCfEQ9eMNXZlG8.webp   → 404 text/html
```

The stored uploadthing key genuinely has no extension, so the fix is the re-upload half of your
comment, not the constant edit — the constant can only be updated once a new extension-bearing key
exists. I don't have upload credentials for the uploadthing app, so this needs an asset owner to
re-upload and hand back the new key. Happy to land the one-line constant change the moment there is
one. Everything else in this round is fixed below.

app/src/libs/combat/process.ts
ifroshty: Confirmed end-to-end and fixed at both sites. Traced the full chain:

- `sageMode.ts:193/198/203` — `delete e.rounds` on `effects` / `afterEffects` / `level2Effects`.
- `tags.ts:67` — `realizeTag` seeds `tag.timeTracker = {}` only under `if ("rounds" in tag)`, so it
  never fires for a stored sage tag.
- `util.ts:1011` — the dedup in `calcApplyRatio` (the function is named `calcApplyRatio`, not
  `getEffectApplyRatio`) requires `effect.timeTracker` truthy; otherwise it falls through to the
  default `ratio = 1` with no gating.

The re-application is reachable, not just theoretical: `applyEffects` runs once per action
(`actions.ts:1371`) and again on the idle/poll path (`combat.ts:277`), so one round can make multiple
passes over the same effect.

Took your first option — `realized.timeTracker = {}` alongside the `realized.rounds` assignment — at
`applyActivateSageMode` and `applySageModeAfterRoundTransition`, rather than preserving `rounds`
through storage, since the stripping is what enforces "duration comes from `activationRounds` /
`afterEffectRounds`, never per-tag `rounds`".

Checked for regressions and other sites: each tag is `structuredClone`d per raw effect
(`process.ts:1870`, `1965`) and again inside `realizeTag`, and `timeTracker` is keyed by `targetId`,
so there's no shared-state risk across targets; `alwaysApply` types return before the tracker check
(`util.ts:1003`) so they're untouched. Those two are the only sage call sites that assign `.rounds`
after `realizeTag`.

**Separate pre-existing issue this surfaced (not fixed here, flagging for a follow-up):** because
`delete e.rounds` is unconditional, it strips `rounds` even when authored as `0`. So post-load
`realized.rounds` is always `undefined`, `undefined !== 0` is always true, and the `rounds === 0`
"instant tag" exception at `process.ts:1887`/`1986` is dead — every stored sage tag becomes a
backdated duration effect. Instant damage/heal tags gate on `castThisRound && rounds === 0`
(`tags.ts:1446`, `1625`), which a stored sage tag can now never satisfy. That guard predates this
round; want me to open a ticket, or is instant one-shot sage authoring not a real use case?

app/src/server/api/routers/bloodline.ts
ifroshty: Fixed as suggested — the refund is now wrapped in its own try/catch that logs and swallows, so the
grant rejection always reaches the client. Comment records why swallowing is the right tradeoff here:
a dropped refund leaves the credit reserved (user is short one pity roll) rather than double-spent,
so failing loudly buys nothing and costs the meaningful error message.

One caveat worth knowing: `refundPityCredit` discards `rowsAffected` and never throws on a zero-row
match (asserted by `tests/server/utils/pityReservation.test.ts:58`), so the new `console.error` fires
only on genuine driver/network errors, never on a no-op refund. If you want that distinguishable
operationally, it'd need `rowsAffected` surfaced as a metric — say the word and I'll add it.

app/src/server/api/routers/combat.ts
ifroshty: Renamed the local to `equippedSageMode` (kept the import alias as-is, since the table name is
referenced by `combat.ts:1781-1782` in `processUsersForBattle`). Only reference was the
`sageMode: equippedSageMode ?? null` property at line 1131; every other `sageMode` identifier in the
file is in a different function scope and resolves to the table or an unrelated property.

app/src/server/utils/concurrency.ts
ifroshty: Added `and(eq(table.id, rollId), gt(table.pityRolls, 0))`, and extended the docstring to say the
predicate is a floor rather than the correctness mechanism (the reserve CAS remains that).

Verified it can't suppress a legitimate refund: `reservePityCredit`'s CAS only lands when the row
still reads the expected value, so a successful reservation always leaves the counter at least one
above zero at refund time; the refund fires at most once per reservation; and no other writer
decrements `pityRolls` (the `pityRolls: 0` writes in `bloodline.ts:848/869/972/1049`,
`item.ts:1391`, `quests.ts:1951`, `register.ts:198` are all inserts on fresh rows). So the guard is a
pure no-op on every correct path and only bites a hypothetical duplicate refund — which, as you note,
now matters more since `pityRolls` widened to signed `smallint`.

app/drizzle/constants.ts
ifroshty: Disregard adjusted

MathiasGruber:

**Conflicts resolved (migrations only):**

`main` has since added `0021_relocate_world_landmarks`, `0022_water_walkability`, `0023_shallow_blackheart` and `0024_military_punisher`, which collided with this branch's hand-numbered `0021_perpetual_wasp` (add/add on `meta/0021_snapshot.json`, content conflict on `meta/_journal.json`).

Resolution: dropped the branch's `0021_perpetual_wasp.sql` and its snapshot, took `main`'s migration metadata, and regenerated with `make makemigrations` → **`0025_tense_crusher_hogan.sql`**. The regenerated SQL is byte-identical to the old `0021` apart from the file number, with one manual fix: drizzle-kit emitted `ALTER TABLE BloodlineRolls MODIFY COLUMN pityRolls smallint NOT NULL` and dropped the `DEFAULT 0`, even though the schema and the generated snapshot both still declare `.default(0)`. `DEFAULT 0` was restored on that statement so the database matches the snapshot.

No source-file conflicts — the rest of the branch replayed cleanly.

Verified after rebase: typecheck clean, lint clean, 558/558 unit tests pass.

app/src/server/api/routers/sageMode.ts
```diff
@@ -0,0 +1,323 @@
...
+
+  // Remove a sage mode from session user
+  removeSageMode: protectedProcedure
+    .output(baseServerResponse)
+    .mutation(async ({ ctx }) => {
+      const user = await fetchUser(ctx.drizzle, ctx.userId);
+      if (!user.sageModeId) {
+        throw serverError("PRECONDITION_FAILED", "You do not have a sage mode");
+      }
+      if (user.reputationPoints < REMOVAL_COST) {
```
MathiasGruber, line 288: **Missing battle-status guard on `removeSageMode`** (from the 2026-07-22 `/tnr-review-now` run, reproduced against current code).

A user whose `status` is `BATTLE` can open `/profile/edit` and remove their equipped Sage Mode mid-fight. Neither this mutation nor `CurrentSageMode` in `app/src/layout/SageMode.tsx` checks `user.status`, while most other user-mutating endpoints in the codebase do gate on `user.status !== "AWAKE"` (e.g. `item.ts:867`, `item.ts:1300`, `item.ts:1604`, `jutsu.ts:538`, `clan.ts:973`).

Scoping this before you act on it, since the QA report speculated about "state edge cases around transient battle Sage fields" and I could not find any:

- The battle carries its own `sageModeId` snapshot on `BattleUserState` plus the definition in `extraState.sageModes`; `performAction` never re-reads `userData` for sage, and the in-battle loadout refresh in `combat.ts` rebuilds from the battle-state user, not the DB row.
- The battle-end flush in `libs/combat/database.ts` (`dailySageActivations + ${user.sageModeUsedThisBattle ? 1 : 0}`) has no CAS on `updatedAt`, so the removal write does not cause the reward flush to be rejected, and the activation still gets counted.
- `sageModeUsedThisBattle` lives only on battle state, so removing and re-acquiring cannot reset the once-per-battle gate.

Also worth noting that `bloodline.removeBloodline` (`app/src/server/api/routers/bloodline.ts:980`) has the identical gap today, so this is a consistency question rather than a regression introduced here.

Net: no correctness or exploit risk found. If the intent is that persistent profile state should not change mid-fight, the guard belongs here and should be mirrored in `CurrentSageMode` so the button is disabled rather than failing after the click.

MathiasGruber (commented):
Triage of the `/tnr-review-now` browser-QA output (runs of 2026-07-19 and 2026-07-22). One finding is still outstanding and reproducible in the current code; the rest were either test-harness limitations or pre-existing behaviour unrelated to this PR.

MathiasGruber (commented):
Audit of previously-raised feedback on this PR. All 13 inline review threads are genuinely fixed in code (verified against the rebased branch) — the migration is now `0025_tense_crusher_hogan`, `IMG_MANUAL_SAGE_MODE` points at a live `.webp` asset, `timeTracker` is seeded at both sage sites, `SuperRefineItem`/`SuperRefineJutsu` both reject `activatesagemode`, the `COALESCE` grant now falls back to a request-local `grantedSageModeId`, `refundPityCredit` has its `pityRolls > 0` floor and is wrapped in its own try/catch, and both parallelisation notes are applied.

What remains are three earlier findings that were raised at PR level (Greptile's review summary and the automated browser-QA run) and never answered or changed in code. Details inline.

app/src/libs/sageMode.ts
```diff
@@ -0,0 +1,119 @@
...
+  sageMode.requiredSageMastery > 0 && exp >= sageMode.requiredSageMastery
+    ? SAGE_MODE_MAX_LEVEL
+    : 1;
+
+/**
+ * Fetch item-based sage mode rolls (used to dedup previously-rolled modes from the pool).
+ */
+export const fetchItemSageModeRolls = async (client: DrizzleClient, userId: string) => {
+  return await client.query.sageModeRolls.findMany({
+    where: and(eq(sageModeRolls.userId, userId), eq(sageModeRolls.type, "ITEM")),
```
MathiasGruber: **Unaddressed from the Greptile review summary** ("Files Needing Attention: `app/src/libs/sageMode.ts` (asymmetric ITEM-only dedup in `filterRollableSageModes`) worth a final author confirmation"). No reply and no code change.

The dedup feeding `filterRollableSageModes` is filtered to `type = "ITEM"` here, while the quest-side dedup in `app/src/server/api/routers/quests.ts` LEFT JOINs `sageModeRolls` on `(sageModeId, userId)` with **no** type filter, so it excludes every previously-rolled mode regardless of source.

Consequence: a mode acquired through a `reward_sage_modes` quest is recorded as `type = "QUEST"` and is therefore invisible to this query, so a player who obtains mode X from a quest, removes it for `REMOVAL_COST` reps, and then consumes a `rollsagemode` item can roll X again. The reverse is not possible — the quest side already blocks re-acquisition of item-rolled modes.

Please either drop the `eq(sageModeRolls.type, "ITEM")` predicate so both acquisition paths dedup against the same history, or add a comment stating that item re-rolls of quest-granted modes are intentional. Also worth renaming the helper if the ITEM scoping is kept, since `filterRollableSageModes` reads as a global dedup.

app/src/server/api/routers/item.ts
```diff
@@ -1341,7 +1360,12 @@ export const itemRouter = createTRPCRouter({
           });
           data.push(bloodlinePool);
           const randomBloodline = getRandomElement(bloodlinePool);
-          if (!randomBloodline) throw serverError("NOT_FOUND", "No bloodline found");
+          if (!randomBloodline) {
```
MathiasGruber, line 1380: **Unaddressed from the Greptile review summary** ("Files Needing Attention: `app/src/server/api/routers/item.ts` (rollbloodline soft-fail behavioral change) worth a final author confirmation"). No reply and no code change.

This replaces the pre-existing `throw serverError("NOT_FOUND", "No bloodline found")` with a soft `messages.push(...); return;` — a behavioural change to the **bloodline** path that is unrelated to Sage Mode and is bundled into this feature PR.

It matters because `consumeUserItemAtomically` has already committed the quantity decrement by the time this runs, so the empty-pool case now silently consumes the item and returns `success: true` with a flavour string. Previously the user got a hard error they could report, which is what surfaced exhausted bloodline pools operationally.

Please confirm this is intentional and, if so, say so in a comment here (and consider whether the item should be refunded rather than silently burned when the pool is empty). If it is incidental, revert it and land it separately from Sage Mode.

app/src/server/api/routers/sageMode.ts
```diff
@@ -0,0 +1,323 @@
...
+      } else {
+        return { success: false, message: `Not allowed to edit sage mode` };
+      }
+    }),
+
+  // Remove a sage mode from session user
+  removeSageMode: protectedProcedure
+    .output(baseServerResponse)
+    .mutation(async ({ ctx }) => {
+      const user = await fetchUser(ctx.drizzle, ctx.userId);
```
MathiasGruber, line 279: **Unaddressed from the automated browser-QA run on this PR** (2026-07-22, severity `Medium`, recommendation "Needs follow-up ... mid-battle removal should be reviewed before approval"). The finding was never replied to and no guard was added on either side.

Reproduction from that run: activate Sage Mode in the Arena, stay in `Status: BATTLE`, open `/profile/edit`, and `removeSageMode` succeeds — 5 reputation is charged and `sageModeId` is cleared while the sage effects, `sageModeActivated`, and `sageModeExpiresRound` are still live in the battle state. There is no status check here and none in `app/src/layout/SageMode.tsx`.

Battle state is loaded from `extraState.sageModes` at `initiateBattle`, so the in-flight battle keeps running on the removed mode and `dailySageActivations` is still incremented at battle end — the player pays the removal cost yet keeps the buff for the remainder of the fight.

Please do one of the following:
- add a guard alongside the existing checks, e.g. `if (user.status !== "AWAKE") return errorResponse("Cannot remove sage mode while " + user.status.toLowerCase());`, and hide/disable the button in `CurrentSageMode` when `userData.status !== "AWAKE"`; or
- reply explaining that this deliberately mirrors `bloodline.removeBloodline` (which is also unguarded) and that the transient sage battle fields are safe, so the QA finding can be closed.

app/src/server/api/routers/sageMode.ts
```diff
@@ -0,0 +1,323 @@
...
+    .input(
+      sageModeFilteringSchema.extend({
+        cursor: z.number().int().nonnegative().nullish(),
+        limit: z.number().int().min(1).max(500),
+      }),
+    )
+    .query(async ({ ctx, input }) => {
+      const currentCursor = input.cursor ? input.cursor : 0;
+      const skip = currentCursor * input.limit;
+      const user = ctx.userId ? await fetchUser(ctx.drizzle, ctx.userId) : null;
```
MathiasGruber: `getAll` is a public, infinite-scroll listing endpoint, and this `await fetchUser(...)` runs sequentially before the list query — so every page of scroll costs two serial round-trips instead of one. `fetchUser` also selects the entire `userData` row (~200 columns incl. JSON blobs like `questData`) purely to read `user.role`.

The `hidden` filter genuinely depends on the role, so it can't be folded into a `Promise.all` as-is, but the read itself can be made cheap:

```ts
const user = ctx.userId
  ? await ctx.drizzle.query.userData.findFirst({
      where: eq(userData.userId, ctx.userId),
      columns: { role: true },
    })
  : null;
```

Compare `get` and `getAllNamesForEdit` just below, which correctly parallelize their user fetch.

MathiasGruber (commented):
Performance pass over the Sage Mode PR. Overall the hot paths look good: no new DB queries were introduced into `performAction` (all sage state is preloaded into battle state during `initiateBattle` and persisted via the existing end-of-battle `updateUser` SQL increment), the new reward query in `updateRewards` and the new item-consume fetches were added inside the existing `Promise.all` batches, and the bloodline router's newly-serialized writes are all correctness-ordered (CAS reserve -> grant -> record) rather than accidental. Two comments below on the new sageMode router.

app/src/server/api/routers/sageMode.ts
```diff
@@ -0,0 +1,323 @@
...
+
+  // Delete a sage mode
+  delete: protectedProcedure
+    .input(z.object({ id: z.string() }))
+    .output(baseServerResponse)
+    .mutation(async ({ ctx, input }) => {
+      const [user, entry, usersWithSageMode] = await Promise.all([
+        fetchUser(ctx.drizzle, ctx.userId),
+        fetchSageMode(ctx.drizzle, input.id),
+        ctx.drizzle.query.userData.findMany({
```
MathiasGruber, line 176: This `findMany` has no `columns` restriction, so it pulls every column of every matching `userData` row (~200 columns including JSON blobs) when only `username` is used on the next line. There is no `limit` either, so a sage mode equipped by a large number of players loads the whole set into memory just to build an error string (which would itself be enormous).

Suggest narrowing and capping:

```ts
ctx.drizzle.query.userData.findMany({
  columns: { username: true },
  where: and(eq(userData.sageModeId, input.id), eq(userData.isAi, false)),
  limit: 10,
}),
```

(The existing `bloodline.delete` has the same shape, so this is inherited rather than introduced — but worth fixing in the new file.)

app/src/libs/combat/process.ts
```diff
@@ -1757,3 +1784,237 @@ export const applyDamageModifierPipelineToConsequences = ({
...
+      round: battle.round,
+      battle,
+    });
+    realized.longitude = newTarget.longitude;
+    realized.latitude = newTarget.latitude;
+    realized.fromType = "sageMode";
+    realized.targetId = newTarget.userId;
+    // Ignore per-effect `rounds` in the JSON for duration; use only `activationRounds` above.
+    // Exception: `rounds === 0` means an instant / one-shot tag — leave unchanged.
+    if (realized.rounds !== 0) {
```
MathiasGruber: The documented `rounds === 0` "instant / one-shot tag" exception is unreachable, so **every** sage effect becomes a duration effect.

`sageMode.update` strips the field from every stored tag before writing:

```ts
effects: input.data.effects.map((e) => { delete e.rounds; delete e.friendlyFire; return e; }),
```

(and the sage editor pushes `rounds` onto `ignore`, so it is never re-added). The JSON column is read back raw — no Zod re-parse — so `battle.extraState.sageModes[id].effects` never has a `rounds` key. `realizeTag` then skips its own `if ("rounds" in tag)` init and `realized.rounds` is `undefined`, and `undefined !== 0` is always true.

Consequences:
- A content designer cannot author a one-shot sage tag (e.g. an instant heal on entry, or a burst of damage on exhaustion) — it will tick for `activationRounds` / `afterEffectRounds` instead.
- The comments here and at the after-effect copy (line ~1987), the `afterEffects` doc on `SageModeValidator` ("instant damage/heal/drain resolve immediately"), and the `sage_mode.test.ts` case titled "…when the active effects were all instant" all describe a configuration that cannot exist.

Either drop the branch and the comments, or give the editor a real way to mark a tag instant (e.g. keep `rounds: 0` through `update` instead of deleting it) so the branch actually does something.

app/src/libs/combat/process.ts
```diff
@@ -1757,3 +1784,237 @@ export const applyDamageModifierPipelineToConsequences = ({
...
+      );
+      if (activeSageFx.some((e) => isEffectActive(e))) return;
+    }
+
+    const sageModeId = u.sageModeId;
+    const sageMode: SageMode | undefined = sageModeId
+      ? (battle.extraState.sageModes ?? {})[sageModeId]
+      : undefined;
+
+    const afterRounds = sageMode?.afterEffectRounds ?? 0;
```
MathiasGruber, line 2100: `afterEffectRounds` is validated as `min(0)`, so a mode can legitimately have `0` while still carrying `afterEffects`. In that case the block below still realizes and pushes every after-effect with `rounds = 0`, and the result splits by tag type inside `calcApplyRatio`:

- tags in the `alwaysApply` list short-circuit on `effect.rounds === 0` and return ratio `0` — they never fire;
- every other tag (`increasemaxpools`, `decreasemaxpools`, `elementalseal`, `immunity`, `finalstand`, …) falls through to `ratio = 1` and is dispatched exactly once before `isEffectActive` drops it at the tail of `applySingleEffect`.

So a `0`-round exhaustion phase silently applies some of its tags and not others. Suggest guarding the whole queueing block on `afterRounds > 0` (or rejecting `afterEffectRounds === 0` in `SageModeValidator` when `afterEffects` is non-empty).

app/src/libs/combat/process.ts
```diff
@@ -1757,3 +1784,237 @@ export const applyDamageModifierPipelineToConsequences = ({
...
+
+    if (sageMode?.afterEffects?.length) {
+      const sorted = [...sageMode.afterEffects].sort((a, b) =>
```
MathiasGruber: DRY: this realization loop is a near-verbatim copy of the one in `applyActivateSageMode` (~lines 1866-1904). Both do: sort with `sortEffects` → `structuredClone` → `realizeTag` → copy `longitude`/`latitude` → set `fromType`/`targetId` → override `rounds` → seed `timeTracker` → backdate `createdRound` → `isNew = false` / `castThisRound = false`.

Only four things differ: `fromType` (`sageMode` vs `sageModeAfter`), the duration source (`activationRounds` vs `afterEffectRounds`), `actionId` (`effect.actionId` vs `"sageModeAfter"`), and the sink (recursive `applySingleEffect` vs `battle.usersEffects.push`). A shared `realizeSageTags({ tags, user, fromType, rounds, actionId, level, battle })` returning the realized array would collapse ~50 lines to one helper and, more importantly, keep the two paths from drifting — the `timeTracker` / `createdRound` backdating subtleties are exactly the kind of thing that gets fixed in one copy only.

app/src/server/api/routers/quests.ts
```diff
@@ -1699,6 +1739,25 @@ export const updateRewards = async (info: {
     hasWarRewards && user.villageId
       ? fetchActiveWars(client, user.villageId)
       : undefined,
+    // Fetch not-yet-owned candidate sage modes for reward_sage_modes (dedup at fetch)
+    (rewards.reward_sage_modes?.length ?? 0) > 0 && !user.sageModeId
```
MathiasGruber: The `!user.sageModeId` gate contradicts the two comments that describe this feature.

Below you write *"Roll ONE not-yet-owned candidate sage mode (if any). It is recorded into the player's history below; **additionally** auto-equip it only when the player currently has no sage mode"*, and at the `sageModeRolls` insert *"Record the not-yet-owned candidate into history (dedup for future quest grants); it auto-equips below only when the player currently has no sage mode"*. Both read as: always roll + record, conditionally equip.

But because the fetch itself is gated on `!user.sageModeId`, a player who already has a sage mode gets `sageModes = []` → `rolledSageMode === undefined` → no `sageModeId` write, **no `sageModeRolls` history row**, and no message. The `reward_sage_modes` payload is silently dropped, and the `COALESCE(...)` guard's stated purpose (protecting against a concurrent grant) is moot since it can only ever run when the column was already read as null.

Pick one: drop the `!user.sageModeId` gate so the history row is written (matching the comments and keeping the COALESCE meaningful), or keep the gate and fix both comments to say the reward is skipped outright for players who already have a mode.

app/src/server/api/routers/quests.ts
```diff
@@ -2058,6 +2138,19 @@ export const fetchUncompletedQuests = async (
           isNull(quest.requiredBloodlineId),
           eq(quest.requiredBloodlineId, user.bloodlineId ?? ""),
         ),
+        or(
```
MathiasGruber: DRY: this 14-line sage availability predicate is now pasted three times in this file — here, and in both branches of the `getQuestsAvailable` ternary (~416-433 and ~470-487) — all byte-identical:

```ts
or(isNull(quest.requiredSageModeId), eq(quest.requiredSageModeId, user.sageModeId ?? "")),
or(isNull(quest.requiredSageRank), inArray(quest.requiredSageRank, sageRanksAtOrBelow(getSageMasteryDisplayRank(user.sageMasteryExperience, !!user.sageModeId)))),
```

Worth extracting a `sageQuestFilters(user)` helper next to `sageRanksAtOrBelow` and spreading it (`...sageQuestFilters(user)`) at the three call sites. As written, a change to the rank semantics has to be made in three places, and `isAvailableUserQuests` in `libs/quest.ts` is a fourth expression of the same rule that has to stay in sync.

app/src/server/api/routers/combat.ts
```diff
@@ -2489,6 +2541,10 @@ export const processUsersForBattle = async (
       damageDealt: 0,
       initiative: 0,
       basicActions: [],
+      sageModeActivated: false,
```
MathiasGruber: These four hard resets sit **after** `...inputUser` in the object literal, so they unconditionally clobber whatever the incoming user carried. That is correct for `initiateBattle`, but `processUsersForBattle` is also re-run mid-battle by `updateCombatLoadout` (~line 1140) on a user rebuilt from live `BattleUserState`.

That endpoint's only staleness guard is `if (user.iAmHere) return errorResponse(...)`, and `iAmHere` is *never* set for a player who does not press "I'm Ready" before the lobby elapses — the `iAmHere` mutation early-returns on `new Date() > userBattle.roundStartAt` without setting the flag. So a player who skips the Ready button can call `updateCombatLoadout` at any point in a COMBAT battle, and it will:

- reset `sageModeActivated` / `sageModeExpiresRound`, so `applySageModeAfterRoundTransition` never fires and the `afterEffects` exhaustion phase is skipped entirely;
- reset `sageModeUsedThisBattle` to `false`, which both re-offers the Activation action (`handleInjectedJutsus` gates on it) and makes `updateUser`'s `dailySageActivations + ${user.sageModeUsedThisBattle ? 1 : 0}` charge nothing at battle end — i.e. the once-per-battle rule and the daily cap are both bypassed.

(The same call also replaces the user's own `usersEffects`, dropping the active `fromType: "sageMode"` buffs, which nothing re-creates.)

`updateCombatLoadout` already explicitly restores `initiative` and `direction` after this call for exactly this reason — the sage fields should be preserved the same way (or seeded from `inputUser.sageModeActivated ?? false` etc. here).

MathiasGruber (commented):
Logic / readability / DRY pass over the Sage Mode feature. Six items: two correctness issues around sage effect duration handling, one reachable path that wipes the in-battle sage state (bypassing the once-per-battle rule and the daily cap), a code/comment mismatch in the quest sage-mode reward grant, and two duplication cleanups. Everything else in the combat activation, daily-cap and CAS/concurrency paths checked out.

app/src/server/api/routers/sageMode.ts
MathiasGruber: Fixed: `removeSageMode` now rejects non-AWAKE users, and `CurrentSageMode` disables the remove button while in battle.

app/src/libs/sageMode.ts
MathiasGruber: Fixed: item rolls now dedup against the full sage-mode history (item and quest), so a quest-granted mode cannot be rolled again after removal.

app/src/server/api/routers/item.ts
MathiasGruber: Reverted: empty bloodline pools throw `NOT_FOUND` again. The soft-fail was incidental to Sage Mode and should not silently burn the item.

app/src/server/api/routers/sageMode.ts
MathiasGruber: Fixed: same AWAKE guard as above, on both the mutation and the profile-edit button.

app/src/server/api/routers/sageMode.ts
MathiasGruber: Fixed: `getAll` now reads only `userData.role` when a user is present, instead of fetching the full row.

app/src/server/api/routers/sageMode.ts
MathiasGruber: Fixed: the delete-in-use lookup now selects only `username` and is capped at 10 rows.

app/src/libs/combat/process.ts
MathiasGruber: Fixed: `update` keeps `rounds: 0` so one-shot tags stay instant, and the sage editor exposes that field as Instant (0) or mode duration.

app/src/libs/combat/process.ts
MathiasGruber: Fixed: after-effects are queued only when `afterEffectRounds > 0`, so a 0-round exhaustion phase no longer applies a subset of tags.

app/src/libs/combat/process.ts
MathiasGruber: Fixed: activation and exhaustion now share `realizeSageTags` so duration, timeTracker, and createdRound backdating stay in one place.

app/src/server/api/routers/quests.ts
MathiasGruber: Fixed: quest sage-mode rewards always roll and record history; auto-equip still uses COALESCE only when the player has no mode.

app/src/server/api/routers/quests.ts
MathiasGruber: Fixed: extracted `sageQuestFilters(user)` and reused it at the remaining SQL call sites.

app/src/server/api/routers/combat.ts
MathiasGruber: Fixed: `processUsersForBattle` now preserves incoming sage battle fields, so a mid-fight loadout refresh cannot reset the once-per-battle gate or skip exhaustion.

MathiasGruber:
Rebased onto latest `main` and addressed the open review threads.

**Rebase:** `main` had added migrations `0025`–`0030`, so the Sage Mode migration was regenerated as `0031_sour_dakota_north`. Source conflicts were merged to keep both sides (quest hall unification, copy/mirror transfer helpers, daily overworld rolls, pity/NPC concurrency helpers). `pityRolls` `DEFAULT 0` was restored on the generated `ALTER` so it matches the snapshot.

**Review fixes:**
- Block mid-battle Sage Mode removal (server + profile button)
- Dedup item rolls against the full roll history
- Revert the incidental bloodline empty-pool soft-fail
- Cheap `getAll` role read; cap delete-in-use usernames
- Keep `rounds: 0` instant tags; queue after-effects only when `afterEffectRounds > 0`
- Share `realizeSageTags`; extract `sageQuestFilters`
- Always record quest sage-mode history; auto-equip only when unequipped
- Preserve live sage battle fields across loadout refresh

app/src/server/api/routers/bloodline.ts
```diff
@@ -1005,24 +1036,22 @@ export const bloodlineRouter = createTRPCRouter({
...
-        }),
-      ]);
+      await updateBloodline(
+        ctx.drizzle,
+        user,
+        line,
+        BLOODLINE_COST[line.rank],
+        `Bloodline Purchased: ${line.name}`,
+      );
+      await ctx.drizzle.insert(bloodlineRolls).values({
```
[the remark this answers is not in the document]
MathiasGruber: Leaving these sequential on purpose.

`updateBloodline` is a CAS grant: it updates `userData` and throws `BloodlineGrantRejectedError` when `rowsAffected === 0`. The `bloodlineRolls` insert is a side-effect that should only run after that grant wins. Parallelizing them would write a DIRECT purchase history row even when the reputation/bloodline CAS loses the race (double-click, concurrent grant, stale snapshot).

That ordering is the same contract as the rest of this helper — side-effect writes stay after the CAS, and `bloodlineEconomy.test.ts` asserts they do not run on a failed grant.

MathiasGruber (approved):

PR #1406, opened 2026-07-20
Inventory improvements

Phrosfire:
# Pull Request

Changes:
- Inventory sorts alphabetically
- Home storage sorts alphabetically
- added merge option to home storage
- Added ryo payment for repairs to the inventory page for crafters

## License

## Summary by CodeRabbit

* **New Features**
  * Added “Merge all stacks” for items stored at home and in carried inventory.
  * Added “Repair All Items” using repair kits or Ryo, with cost and affordability details.
  * Added per-item repair options for damaged equipped and backpack items.

* **Improvements**
  * Repair actions now show availability, costs, and progress states.
  * Home, backpack, and character inventories sort items by name.
  * Repair actions exclude ineligible or auctioned items.
  * Stacked leveling equipment now displays quantity and level badges.
  * Confirmation dialogs support optional extra footer actions.

### DIFF
--- a/app/src/app/home/page.tsx
+++ b/app/src/app/home/page.tsx
@@ -32,8 +32,10 @@ import ContentBox from "@/layout/ContentBox";
 import Image from "@/layout/Image";
 import ItemWithEffects from "@/layout/ItemWithEffects";
 import Loader from "@/layout/Loader";
+import { MergeAllStacksButton } from "@/layout/MergeAllStacksButton";
 import Modal2 from "@/layout/Modal2";
 import {
+  byItemName,
   calcMaxHouseMaterials,
   showsItemLevelBadge,
   userItemActionBadges,
@@ -109,25 +111,31 @@ export default function HomePage() {
   );
   // Filter normal items (non-materials)
   const storedItems =
-    filteredItems?.filter(
-      (useritem) => useritem.storedAtHome && useritem.item.itemType !== "MATERIAL",
-    ) ?? [];
+    filteredItems
+      ?.filter(
+        (useritem) => useritem.storedAtHome && useritem.item.itemType !== "MATERIAL",
+      )
+      .sort(byItemName) ?? [];
   const nonStoredItems =
     filteredItems
       ?.filter((useritem) => !useritem.storedAtHome)
       .filter((useritem) => useritem.equipped === "NONE")
-      .filter((useritem) => useritem.item.itemType !== "MATERIAL") ?? [];
+      .filter((useritem) => useritem.item.itemType !== "MATERIAL")
+      .sort(byItemName) ?? [];
 
   // Filter materials separately
   const storedMaterials =
-    filteredItems?.filter(
-      (useritem) => useritem.storedAtHome && useritem.item.itemType === "MATERIAL",
-    ) ?? [];
+    filteredItems
+      ?.filter(
+        (useritem) => useritem.storedAtHome && useritem.item.itemType === "MATERIAL",
+      )
+      .sort(byItemName) ?? [];
   const nonStoredMaterials =
     filteredItems
       ?.filter((useritem) => !useritem.storedAtHome)
       .filter((useritem) => useritem.equipped === "NONE")
-      .filter((useritem) => useritem.item.itemType === "MATERIAL") ?? [];
+      .filter((useritem) => useritem.item.itemType === "MATERIAL")
+      .sort(byItemName) ?? [];
 
   const canStoreMoreItems = storedItems.length < homeStorage;
   const canStoreMoreMaterials =
@@ -377,6 +385,17 @@ export default function HomePage() {
             title="Item Storage"
             subtitle={`Items in home (${totalStoredItems}/${homeStorage} slots used) | Materials in home (${storedMaterials.length}/${userData && homeData ? calcMaxHouseMaterials(userData, homeData.storage) : 0} slots used)`}
             initialBreak={true}
+            topRightContent={
+              homeData?.homeType !== "NONE" ? (
+                <MergeAllStacksButton
+                  storedAtHome={true}
+                  onMerged={() => {
+                    setSelectedItem(undefined);
+                    setIsModalOpen(false);
+                  }}
+                />
+              ) : undefined
+            }
           >
             {isHomeLoading || isItemsLoading ? (
               <Loader explanation="Loading item storage data" />
--- a/app/src/app/home/page.tsx
+++ b/app/src/app/home/page.tsx
@@ -386,7 +386,7 @@ export default function HomePage() {
             subtitle={`Items in home (${totalStoredItems}/${homeStorage} slots used) | Materials in home (${storedMaterials.length}/${userData && homeData ? calcMaxHouseMaterials(userData, homeData.storage) : 0} slots used)`}
             initialBreak={true}
             topRightContent={
-              homeData?.homeType !== "NONE" ? (
+              homeData && homeData.homeType !== "NONE" ? (
                 <MergeAllStacksButton
                   storedAtHome={true}
                   onMerged={() => {
--- a/app/src/app/items/page.tsx
+++ b/app/src/app/items/page.tsx
@@ -849,6 +849,8 @@ const Backpack: React.FC<BackpackProps> = (props) => {
     (userItem: UserItemWithRelations) =>
       userItem.item?.effects?.some((e: { type: string }) => e.type === "repair") &&
       userItem.quantity > 0 &&
+      !userItem.storedAtHome &&
+      !userItem.isInAuction &&
       (!userItem.craftingFinishedAt || userItem.craftingFinishedAt < new Date()),
   );
 
@@ -1269,6 +1271,8 @@ const Character: React.FC<CharacterProps> = (props) => {
     (userItem: UserItemWithRelations) =>
       userItem.item?.effects?.some((e: { type: string }) => e.type === "repair") &&
       userItem.quantity > 0 &&
+      !userItem.storedAtHome &&
+      !userItem.isInAuction &&
       (!userItem.craftingFinishedAt || userItem.craftingFinishedAt < new Date()),
   );
 
--- a/app/src/app/items/page.tsx
+++ b/app/src/app/items/page.tsx
@@ -314,6 +314,7 @@ export default function MyItems() {
                 canRepairWithKits && !isRepairAllPending ? "Repair with Kits" : null
               }
               isValid={canRepairWithKits && !isRepairAllPending}
+              disabled={isRepairAllPending}
               button={
                 <Button disabled={isRepairAllPending} variant="outline">
                   <Wrench className="mr-2 h-4 w-4" />
--- a/app/src/app/items/page.tsx
+++ b/app/src/app/items/page.tsx
@@ -47,11 +47,14 @@ import Image from "@/layout/Image";
 import ItemLoadoutSelector from "@/layout/ItemLoadoutSelector";
 import ItemWithEffects from "@/layout/ItemWithEffects";
 import Loader from "@/layout/Loader";
+import { MergeAllStacksButton } from "@/layout/MergeAllStacksButton";
 import Modal2 from "@/layout/Modal2";
 import NavTabs from "@/layout/NavTabs";
 import { applyActiveVariant } from "@/libs/combat/util";
 import { meetsEvolutionStatRequirements } from "@/libs/evolution";
 import {
+  byItemName,
+  calcItemRepairCost,
   calcItemSellingPrice,
   calcMaxEventItems,
   calcMaxItems,
@@ -111,24 +114,24 @@ export default function MyItems() {
       onSuccess: async (data) => {
         showMutationToast(data);
         if (data.success) {
-          await utils.item.getUserItemsWithVariants.invalidate();
-          await utils.profile.getUser.invalidate();
+          await Promise.all([
+            utils.item.getUserItemsWithVariants.invalidate(),
+            utils.profile.getUser.invalidate(),
+          ]);
         }
       },
     });
 
-  const { mutate: mergeAllStacks, isPending: isMergingAllStacks } =
-    api.item.mergeAllStacks.useMutation({
+  const { mutate: mutateRepairAllRyo, isPending: isRepairingAllRyo } =
+    api.item.repairAll.useMutation({
       onSuccess: async (data) => {
         showMutationToast(data);
-        await utils.item.getUserItemsWithVariants.invalidate();
-      },
-      onError: (error) => {
-        showMutationToast({
-          success: false,
-          message: error.message,
-          variant: "destructive",
-        });
+        if (data.success) {
+          await Promise.all([
+            utils.item.getUserItemsWithVariants.invalidate(),
+            utils.profile.getUser.invalidate(),
+          ]);
+        }
       },
     });
 
@@ -190,6 +193,13 @@ export default function MyItems() {
       useritem.item.maxDurability > 0,
   );
   const repairKits = getRepairKits(userItems);
+  const isCrafter = userData.occupation === "CRAFTING";
+  const totalRyoRepairCost = itemsNeedingRepair.reduce(
+    (total, useritem) => total + calcItemRepairCost(useritem),
+    0,
+  );
+  const canAffordRyoRepair = userData.money >= totalRyoRepairCost;
+  const isRepairAllPending = isRepairingAll || isRepairingAllRyo;
 
   // Calculate which kits will be used (shared with item.useRepairAll)
   const repairKitCalculation = calculateKitsToUse(
@@ -199,6 +209,7 @@ export default function MyItems() {
   );
 
   const repairAllInfo = repairKitCalculation;
+  const canRepairWithKits = repairKits.length > 0 && repairAllInfo.canRepairAll;
 
   return (
     <>
@@ -252,38 +263,13 @@ export default function MyItems() {
             <div className="w-full basis-1/2 p-3">
               <h2 className="font-bold text-2xl text-foreground">Equipped</h2>
               <div className="relative">
-                <Character useritems={userItems} />
+                <Character useritems={userItems} userData={userData} />
               </div>
             </div>
             <div className="max-h-full basis-1/2 overflow-y-scroll border-t-2 border-dashed bg-poppopover p-3 sm:max-h-[600px] sm:border-t-0 sm:border-l-2">
               <div className="mb-2 flex flex-row flex-wrap items-center justify-between gap-2">
                 <h2 className="font-bold text-2xl text-foreground">Backpack</h2>
-                <Confirm2
-                  title="Merge all stacks"
-                  proceed_label={isMergingAllStacks ? undefined : "Merge all stacks"}
-                  isValid={!isMergingAllStacks}
-                  button={
-                    <Button
-                      type="button"
-                      variant="outline"
-                      size="sm"
-                      disabled={isMergingAllStacks}
-                    >
-                      <Merge className="mr-2 h-4 w-4" />
-                      {isMergingAllStacks ? "Merging..." : "Merge all stacks"}
-                    </Button>
-                  }
-                  onAccept={(e) => {
-                    e.preventDefault();
-                    mergeAllStacks();
-                  }}
-                >
-                  <p>
-                    Consolidate stackable items in your carried inventory (backpack and
-                    equipped slots) into the fewest possible stacks. Home storage is not
-                    affected. This cannot be automatically undone.
-                  </p>
-                </Confirm2>
+                <MergeAllStacksButton storedAtHome={false} />
               </div>
               <Backpack
                 userData={userData}
@@ -325,64 +311,106 @@ export default function MyItems() {
           </Confirm2>
         </div>
         <div className="flex gap-2">
-          {itemsNeedingRepair.length > 0 && repairKits.length > 0 && (
+          {itemsNeedingRepair.length > 0 && (repairKits.length > 0 || isCrafter) && (
             <Confirm2
               title="Repair All Items"
               proceed_label={
-                repairAllInfo.canRepairAll
-                  ? isRepairingAll
-                    ? undefined
-                    : "Repair All"
-                  : undefined
+                canRepairWithKits && !isRepairAllPending ? "Repair with Kits" : null
               }
-              isValid={repairAllInfo.canRepairAll && !isRepairingAll}
+              isValid={canRepairWithKits && !isRepairAllPending}
               button={
-                <Button disabled={isRepairingAll} variant="outline">
+                <Button disabled={isRepairAllPending} variant="outline">
                   <Wrench className="mr-2 h-4 w-4" />
-                  {isRepairingAll ? "Repairing..." : "Repair All"}
+                  {isRepairAllPending ? "Repairing..." : "Repair All"}
                 </Button>
               }
               onAccept={(e) => {
                 e.preventDefault();
-                if (repairAllInfo.canRepairAll) {
+                if (canRepairWithKits) {
                   mutateRepairAll();
                 }
               }}
+              footerExtra={
+                isCrafter
+                  ? ({ close }) => (
+                      <Button
+                        variant="default"
+                        disabled={!canAffordRyoRepair || isRepairAllPending}
+                        onClick={(e) => {
+                          e.preventDefault();
+                          e.stopPropagation();
+                          mutateRepairAllRyo();
+                          close();
+                        }}
+                      >
+                        <CircleDollarSign className="mr-2 h-4 w-4" />
+                        Repair with Ryo ({totalRyoRepairCost.toLocaleString()})
+                      </Button>
+                    )
+                  : undefined
+              }
             >
               <div className="space-y-3">
-                {repairAllInfo.canRepairAll ? (
-                  <>
-                    <p>
-                      You are about to repair all {itemsNeedingRepair.length} item
-                      {itemsNeedingRepair.length !== 1 ? "s" : ""} using repair kits.
-                    </p>
-                    {repairAllInfo.kitsToUse.length > 0 && (
-                      <div>
-                        <p className="mb-2 font-semibold">Repair kits to be used:</p>
-                        <div className="space-y-1">
-                          {repairAllInfo.kitsToUse.map((kit) => (
-                            <div
-                              key={kit.repairItemId}
-                              className="flex items-center justify-between text-sm"
-                            >
-                              <span>{kit.repairItemName}</span>
-                              <span className="font-medium">x{kit.quantityUsed}</span>
-                            </div>
-                          ))}
+                {repairKits.length > 0 ? (
+                  canRepairWithKits ? (
+                    <>
+                      <p>
+                        You are about to repair all {itemsNeedingRepair.length} item
+                        {itemsNeedingRepair.length !== 1 ? "s" : ""} using repair kits.
+                      </p>
+                      {repairAllInfo.kitsToUse.length > 0 && (
+                        <div>
+                          <p className="mb-2 font-semibold">Repair kits to be used:</p>
+                          <div className="space-y-1">
+                            {repairAllInfo.kitsToUse.map((kit) => (
+                              <div
+                                key={kit.repairItemId}
+                                className="flex items-center justify-between text-sm"
+                              >
+                                <span>{kit.repairItemName}</span>
+                                <span className="font-medium">x{kit.quantityUsed}</span>
+                              </div>
+                            ))}
+                          </div>
                         </div>
-                      </div>
-                    )}
-                  </>
+                      )}
+                    </>
+                  ) : (
+                    <div>
+                      <p className="mb-2 font-semibold text-red-600">
+                        Not enough repair kits
+                      </p>
+                      <p className="text-muted-foreground text-sm">
+                        You have {itemsNeedingRepair.length} damaged item
+                        {itemsNeedingRepair.length !== 1 ? "s" : ""} that need{" "}
+                        {repairAllInfo.totalDurabilityNeeded} total durability, but you
+                        don&apos;t have enough repair kits to repair all of them.
+                      </p>
+                    </div>
+                  )
                 ) : (
-                  <div>
-                    <p className="mb-2 font-semibold text-red-600">
-                      Not enough repair kits
-                    </p>
-                    <p className="text-muted-foreground text-sm">
-                      You have {itemsNeedingRepair.length} damaged item
-                      {itemsNeedingRepair.length !== 1 ? "s" : ""} that need{" "}
-                      {repairAllInfo.totalDurabilityNeeded} total durability, but you
-                      don&apos;t have enough repair kits to repair all of them.
+                  <p className="text-muted-foreground text-sm">
+                    You don&apos;t have any repair kits. As a crafter, you can repair
+                    with ryo instead.
+                  </p>
+                )}
+                {isCrafter && (
+                  <div className="rounded-lg border bg-muted/50 p-3 text-sm">
+                    <p className="font-semibold">Repair with ryo</p>
+                    <p className="text-muted-foreground">
+                      Total cost:{" "}
+                      <span
+                        className={
+                          canAffordRyoRepair ? "text-green-600" : "text-red-600"
+                        }
+                      >
+                        {totalRyoRepairCost.toLocaleString()} ryo
+                      </span>
+                      {!canAffordRyoRepair && (
+                        <span className="ml-2">
+                          (You have {userData.money.toLocaleString()} ryo)
+                        </span>
+                      )}
                     </p>
                   </div>
                 )}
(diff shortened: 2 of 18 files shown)


Phrosfire:
@coderabbitai review

app/src/app/items/page.tsx
```diff
@@ -107,6 +108,17 @@ export default function MyItems() {
       },
     });
 
+  const { mutate: mutateRepairAllRyo, isPending: isRepairingAllRyo } =
+    api.item.repairAll.useMutation({
+      onSuccess: async (data) => {
+        showMutationToast(data);
+        if (data.success) {
+          await utils.item.getUserItemsWithVariants.invalidate();
+          await utils.profile.getUser.invalidate();
```
MathiasGruber: These two invalidations are independent and can run in parallel. Per the CLAUDE.md guideline ("prefer reducing the number of DB roundtrips … running queries in parallel with `Promise.all()` is faster than sequential"), wrap them in `Promise.all`:

```ts
await Promise.all([
  utils.item.getUserItemsWithVariants.invalidate(),
  utils.profile.getUser.invalidate(),
]);
```

This matches the existing convention in the same file (see `unequipAllItems.onSuccess`, which already uses `Promise.all`).

app/src/app/items/page.tsx
```diff
@@ -665,6 +734,18 @@ const Backpack: React.FC<BackpackProps> = (props) => {
       },
     });
 
+  const { mutate: mutateRepairWithRyo, isPending: isRepairingWithRyo } =
+    api.item.repair.useMutation({
+      onSuccess: async (data) => {
+        showMutationToast(data);
+        if (data.success) {
+          await utils.item.getUserItemsWithVariants.invalidate();
+          await utils.profile.getUser.invalidate();
```
MathiasGruber: Same performance issue as in `mutateRepairAllRyo` above — these two invalidations run sequentially and hold the UI in a pending state longer than needed. Parallelize with `Promise.all`:

```ts
await Promise.all([
  utils.item.getUserItemsWithVariants.invalidate(),
  utils.profile.getUser.invalidate(),
]);
onSettled();
```

MathiasGruber (commented):
Performance review — parallelize cache invalidations after mutations.

app/src/app/items/page.tsx
```diff
@@ -986,8 +1114,10 @@ const Character: React.FC<CharacterProps> = (props) => {
         showMutationToast(data);
         if (data.success) {
           setIsRepairModalOpen(false);
-          await utils.item.getUserItemsWithVariants.invalidate();
-          await utils.profile.getUser.invalidate();
+          await Promise.all([
+            utils.item.getUserItemsWithVariants.invalidate(),
+            utils.profile.getUser.invalidate(),
+          ]);
```
MathiasGruber, line 1320: The per-item **Repair with Ryo** action was only added to `Backpack`'s item-details modal, not to `Character`'s. Equipped items are exactly the ones that take durability damage in combat, so a crafter who opens a damaged *equipped* item here only sees `Use Repair Item` (disabled when they own no kits) and has no way to repair that single item with ryo — even though `Repair All` already offers `Repair with Ryo` and includes equipped items in its cost.

This is the gap the automated preview review reported (damaged equipped `Shuriken`, crafter account, modal showed `Use Repair Item` but no `Repair with Ryo`).

Please mirror the Backpack implementation here: add `isCrafter` / `calcItemRepairCost(useritem)` / affordability derivations in `Character`, wire an `api.item.repair` mutation (with the same `Promise.all` invalidation of `getUserItemsWithVariants` + `profile.getUser`), and render the `Confirm2` "Repair with Ryo" button next to `Use Repair Item` under the same `durability < maxDurability && maxDurability > 0` condition. Consider extracting the shared block so the two modals cannot drift again.

app/src/app/home/page.tsx
```diff
@@ -86,6 +88,17 @@ export default function HomePage() {
       },
     });
 
+  const { mutate: mergeAllHomeStacks, isPending: isMergingAllHomeStacks } =
+    api.item.mergeAllStacks.useMutation({
+      onSuccess: (data) => {
+        showMutationToast(data);
+        void userItemsRefetch();
+        void userHomeRefetch();
```
MathiasGruber: `userHomeRefetch()` is a wasted round-trip here. `api.home.getUserHome` returns only `{ homeType, regen, storage }`, all derived from `user.homeType` — merging item stacks cannot change any of them. It is also not a cheap query: `getUserHome` goes through `fetchUpdatedUser`, which fans out into ~7 parallel queries (user with bloodline/clan/village+structures+relationships+sectors, achievements, game settings, polls, wars, shrine battles, raids) plus the regen update path. So every "Merge all stacks" click pays for the heaviest user query on the page for data that is guaranteed unchanged.

Also note the handler has no `if (data.success)` guard, unlike `upgradeHome` and `toggleStoreItem` just above — so both refetches fire even on a failed merge or the common "Nothing to merge" response.

Suggest:

```ts
onSuccess: (data) => {
  showMutationToast(data);
  if (data.success) {
    void userItemsRefetch();
  }
  setSelectedItem(undefined);
  setIsModalOpen(false);
},
```

app/src/app/items/page.tsx
```diff
@@ -659,8 +732,24 @@ const Backpack: React.FC<BackpackProps> = (props) => {
...
+  const { mutate: mutateRepairWithRyo, isPending: isRepairingWithRyo } =
+    api.item.repair.useMutation({
+      onSuccess: async (data) => {
+        showMutationToast(data);
+        if (data.success) {
+          await Promise.all([
+            utils.item.getUserItemsWithVariants.invalidate(),
+            utils.profile.getUser.invalidate(),
+          ]);
+          onSettled();
```
MathiasGruber: `onSettled()` runs only after both `invalidate()` calls resolve, so the item modal stays open (and `document.body.style.cursor` stays busy) for the full duration of the `getUserItemsWithVariants` + `profile.getUser` refetches — `profile.getUser` in particular is one of the heaviest queries in the app.

Every sibling mutation in this component (`merge`, `consume`, `sell`, `equip`, `splitStack`) passes `onSettled` as a mutation *option* so the UI closes as soon as the mutation returns, and `mutateRepairItem` directly above closes its modal (`setIsRepairModalOpen(false)`) before awaiting. Moving the close ahead of the await keeps this consistent and removes the perceived stall:

```ts
onSuccess: async (data) => {
  showMutationToast(data);
  if (data.success) {
    onSettled();
    await Promise.all([
      utils.item.getUserItemsWithVariants.invalidate(),
      utils.profile.getUser.invalidate(),
    ]);
  }
},
```

(As a side effect this also fixes the case where the mutation fails — currently `onSettled()` is never called at all, leaving the modal open and the cursor stuck.)

app/src/app/items/page.tsx
```diff
@@ -781,14 +876,47 @@ const Backpack: React.FC<BackpackProps> = (props) => {
...
+                  <>
+                    <Button
+                      variant="outline"
+                      onClick={() => setIsRepairModalOpen(true)}
+                      disabled={repairItems.length === 0}
+                    >
+                      <Wrench className="mr-2 h-5 w-5" />
+                      Use Repair Item
+                    </Button>
+                    {isCrafter && (
```
MathiasGruber: The enclosing action row is `<div className="flex flex-row gap-1">` — no `flex-wrap`. Every `Button` carries `whitespace-nowrap` from `buttonVariants`, and `DialogContent` is `max-w-[calc(100%-2rem)]` below `md`, so the row can neither wrap nor shrink. For a damaged stackable item this row is now Equip + Merge Stacks + Split Stack + Use Repair Item + **Repair with Ryo** + Sell Item, which overflows a ~343px mobile dialog by a wide margin; the trailing buttons end up off-screen behind a horizontal scroll rather than visible. Adding `flex-wrap` to the parent row keeps the new button reachable on mobile (repo is mobile-first per CLAUDE.md).

app/src/app/items/page.tsx
```diff
@@ -986,8 +1114,10 @@ const Character: React.FC<CharacterProps> = (props) => {
         showMutationToast(data);
         if (data.success) {
           setIsRepairModalOpen(false);
-          await utils.item.getUserItemsWithVariants.invalidate();
-          await utils.profile.getUser.invalidate();
+          await Promise.all([
```
MathiasGruber, line 1317: The per-item **Repair with Ryo** action was only added to `Backpack`; the `Character` item-details modal (the `Unequip` / `Use Repair Item` row further down in this component) still offers `Use Repair Item` only. Durability is consumed in combat, which only affects *equipped* gear, so the common case — a crafter opening a damaged equipped weapon — shows no ryo repair option even though the top-level `Repair All` dialog offers `Repair with Ryo` and `item.repair` accepts any owned `userItemId`. This is very likely what the browser review run flagged as "per-item Ryo repair is missing in item details". Suggest mirroring the `isCrafter` + `calcItemRepairCost` + `api.item.repair` block here, or extracting it into a small shared component used by both modals so the two paths cannot drift again.

app/src/app/items/page.tsx
```diff
@@ -673,18 +762,24 @@ const Backpack: React.FC<BackpackProps> = (props) => {
...
-    ...useritem,
-  }));
+    isUsingRepairItem ||
+    isRepairingWithRyo;
+  const items = useritems
+    ?.map((useritem) => ({
+      ...applyActiveVariant(useritem as UserItemWithVariants),
+      ...useritem,
+    }))
+    .sort((a, b) => a.item.name.localeCompare(b.item.name));
```
MathiasGruber: The sort key doesn't match what the list actually renders. `applyActiveVariant()` returns the item fields with `name` overridden by the active variant's name, and because `...useritem` (a `UserItem` row) has no `name` field, the top-level `name` on each element is the **variant** name — that's what `ActionSelector`/`ActionOption` destructures and displays (`CombatActions.tsx`, `const { ..., name, ... } = item`). Sorting by `a.item.name` uses the **base** item name instead, so any item whose active variant defines its own `name` renders visibly out of alphabetical order.

```suggestion
    .sort((a, b) => a.name.localeCompare(b.name));
```

Two related points while here:
- The same comparator now exists inline here and as `byItemName` in `app/src/app/home/page.tsx`. Worth a single exported helper (e.g. in `@/libs/item.ts`) rather than two copies that can drift.
- `Character`'s `items` (same file, `const items = useritems?.map(...)` around L1068) is still unsorted, so the "Select Item to Equip" picker keeps DB order while the backpack is now alphabetical — inconsistent within the same page.

app/src/app/items/page.tsx
```diff
@@ -315,21 +338,25 @@ export default function MyItems() {
...
+          {itemsNeedingRepair.length > 0 && (repairKits.length > 0 || isCrafter) && (
             <Confirm2
               title="Repair All Items"
               proceed_label={
-                repairAllInfo.canRepairAll
-                  ? isRepairingAll
-                    ? undefined
-                    : "Repair All"
-                  : undefined
+                repairKits.length > 0 && repairAllInfo.canRepairAll
```
MathiasGruber: `repairKits.length > 0 && repairAllInfo.canRepairAll` is now spelled out three times in this block (`proceed_label`, `isValid` right below, and again as the branch condition of the body ternary at L389/L390), and `proceed_label` is a triple-nested ternary that collapses to a single condition. Hoisting it next to `isRepairAllPending` (~L205) makes the intent readable and keeps the three uses from drifting apart:

```ts
const canRepairWithKits = repairKits.length > 0 && repairAllInfo.canRepairAll;
```

then `proceed_label={canRepairWithKits && !isRepairAllPending ? "Repair with Kits" : null}` and `isValid={canRepairWithKits && !isRepairAllPending}`.

app/src/app/items/page.tsx
```diff
@@ -673,18 +762,24 @@ const Backpack: React.FC<BackpackProps> = (props) => {
     isSelling ||
     isEquipping ||
     isSplitting ||
-    isUsingRepairItem;
-  const items = useritems?.map((useritem) => ({
-    ...applyActiveVariant(useritem as UserItemWithVariants),
-    ...useritem,
-  }));
+    isUsingRepairItem ||
+    isRepairingWithRyo;
```
MathiasGruber, line 840: `isRepairingWithRyo` is folded into `isLoading`, and `isLoading` gates the entire action-button row (`{!isLoading && (...)}` at L834-L957) — but unlike `isMerging`/`isConsuming`/`isSelling`/`isEquipping` there is no matching `<Loader>` at L958-L961. So after confirming "Repair for X ryo" the Item Details modal stays open showing the item with no buttons and no spinner until the mutation resolves. Add the missing line alongside the others:

```tsx
{isRepairingWithRyo && <Loader explanation={`Repairing ${useritem.item.name}`} />}
```

app/src/app/home/page.tsx
```diff
@@ -366,6 +387,38 @@ export default function HomePage() {
             title="Item Storage"
             subtitle={`Items in home (${totalStoredItems}/${homeStorage} slots used) | Materials in home (${storedMaterials.length}/${userData && homeData ? calcMaxHouseMaterials(userData, homeData.storage) : 0} slots used)`}
             initialBreak={true}
+            topRightContent={
+              homeData?.homeType !== "NONE" ? (
+                <Confirm2
```
MathiasGruber: This whole `Confirm2` block is a near-verbatim copy of the one in `app/src/app/items/page.tsx` (L274-L299): same title, same `proceed_label`/`isValid`/`disabled` wiring, same `Merge` icon with the "Merging..."/"Merge all stacks" label, and the same body copy with only "home storage"/"carried inventory" swapped. Since `mergeAllStacks` now takes `{ storedAtHome }`, this is a good candidate for one shared component in `@/layout/` (e.g. `MergeAllStacksButton({ storedAtHome, onMerged })`) rendered by both pages — otherwise any future change to the confirm copy or pending-state handling has to be made twice.

Also note the two call sites handle cache invalidation differently: `items/page.tsx` invalidates `item.getUserItemsWithVariants`, this one refetches `item.getUserItems` + `home.getUserHome`. With `staleTime: Infinity` (`app/src/app/_trpc/Provider.tsx`), whichever query the *other* page uses keeps serving pre-merge rows until gc. A shared component would be the natural place to invalidate both.

app/src/server/api/routers/item.ts
```diff
@@ -3411,59 +3704,100 @@ async function executeMergeStacksForItemBucket(
...
+    await Promise.all(
+      sortedItems.map((item) =>
+        drizzle
+          .update(userItem)
+          .set({ quantity: item.quantity, updatedAt: item.updatedAt })
+          .where(
+            and(
+              eq(userItem.id, item.id),
+              eq(userItem.userId, userId),
+              eq(userItem.quantity, -item.quantity),
```
MathiasGruber: This rollback can't actually recover a partially-applied publish, and that makes the whole protocol unsound under interference. The publish UPDATE has per-row guards, so a concurrent write that unmatches a strict subset of claimed rows (auction listing flipping `isInAuction`, a quest `remove_item` delete, the battle-end writeback, the zero-grace reaper — see my review comment) makes it apply to n−1 rows: the keeper gets the **absolute CASE target that already includes the unmatched row's quantity**, and `rowsAffected !== n` fires. But this restore only matches `eq(quantity, -item.quantity)` — published rows now hold non-negative CASE targets and can never match it. Net result: the unmatched row is restored to +q while its quantity also lives in the keeper → permanent duplication. Mirror case (keeper unmatched): sources publish to their small/zero targets while the keeper restores to its original → permanent loss. Either way the user just sees "inventory changed, please try again" and the corruption is silent.

I don't think this is patchable with more guards — any writer that touches a guard column mid-window recreates it. It needs claim state the publish/rollback can target exactly (dedicated claim column/token), or partial publish must be treated as corruption to repair from claim records, not as a benign CAS conflict.

app/src/server/utils/concurrency.ts
```diff
@@ -118,6 +131,116 @@ export const consumeUserItemAtomically = async ({
...
+
+  const result = await refund();
+  if (result.rowsAffected > 0) return;
+
+  // A no-op duplicate update means the row is held by a merge claim. Recover every row in that
+  // merge before retrying so its pending publish cannot overwrite the refunded quantity.
+  await restoreStaleUserItemMergeClaims({
+    client,
+    userId: itemSnapshot.userId,
+    staleBefore: new Date(),
```
MathiasGruber: `staleBefore: new Date()` gives zero grace — this reaps **every** negative row of the user, including claims an active in-flight merge created microseconds ago (`lte(updatedAt, staleBefore)` is always true for them), and it can't distinguish active from abandoned claims. Since this runs from the `useRepairAll` kit-consume rollback, an ordinary "Repair All" racing "Merge all stacks" from two tabs flips the merge's claims positive mid-protocol. If the flip lands between the merge's parallel claim statements and its publish, a strict subset of rows unmatches → the unrecoverable partial publish flagged at the rollback in item.ts → duplication/loss with no external writer involved at all. The PR races itself here.

At minimum this should pass the same 30s staleness window as everywhere else; with claim ownership (a claim id/token column) it could restore only the abandoned claim for the specific row it needs instead of a user-wide sweep.

app/src/server/utils/concurrency.ts
```diff
@@ -118,6 +131,116 @@ export const consumeUserItemAtomically = async ({
...
+  });
+  await refund();
```
MathiasGruber: This retry is fire-and-forget: no `rowsAffected` check, no throw, no logging. If the row is negative again at retry time (a new merge re-claimed it in the reap→retry gap — e.g. the user re-clicks "Merge all stacks" after the conflict toast), the `IF(quantity >= 0, ...)` upsert no-ops a second time and the refund is silently dropped — the kits are permanently gone while `useRepairAll` returns "Could not consume repair kits — please try again", implying nothing was consumed. Narrow three-way race, but every participant is an ordinary user mutation. The retry result needs to be checked and surfaced (retry again / log / compensate), not discarded.

app/src/server/api/routers/item.ts
```diff
@@ -2275,35 +2356,112 @@ export const itemRouter = createTRPCRouter({
...
+        );
+        return errorResponse(
+          "Could not repair all items — one or more were stored, auctioned, or changed. No kits were consumed.",
+        );
+      }
+
+      // Consume repair kits only after every target write succeeded
+      const kitConsumeResults = await Promise.all(
+        kitsToUse.map(async ({ repairItemId, quantityUsed }) => {
+          const repairKitRow = useritems.find((ui) => ui.id === repairItemId);
```
MathiasGruber, line 2415: Two problems with the kit side of this path:

1. **Scope**: `getRepairKits` never excludes `storedAtHome`/`isInAuction` stacks, and these consume CASes guard only id+userId+quantity — so bulk repair silently spends kits parked in home storage and can delete/decrement a kit stack backing a **live auction listing** (quantity is unchanged, so the guard passes). `completeAuctionInternal` transfers by bare `where(eq(id))` with no `rowsAffected` check, so the winner is charged and receives nothing or a short stack. The single-kit `useRepairItem` path rejects exactly these kits ("Fetch the repair item from home storage first" / auction guard) — the bulk path needs the same exclusions in `getRepairKits` or in these WHERE clauses.

2. **Reuse**: this delete-if-exact-quantity / decrement-with-guard pair re-implements `updateUserItemQuantityAtomically` — the helper this exact code called before the PR and occupation.ts still uses. `updateUserItemQuantityAtomically({ client, userId, userItemId: repairItemId, expectedQuantity: repairKitRow.quantity, nextQuantity: repairKitRow.quantity - quantityUsed })` is a drop-in (with `quantityConsumed: quantityUsed`; `calculateKitsToUse` clamps so `quantityUsed <= quantity` always).

app/src/server/api/routers/item.ts
```diff
@@ -3411,59 +3704,100 @@ async function executeMergeStacksForItemBucket(
...
-      return [];
-    }
-    return [
+  const conflictMessage = `Failed to merge stacks of ${itemName} — inventory changed, please try again`;
+
+  const claimedAt = new Date();
+
+  // Phase 1: claim every row (qty → -qty) so no subset of later writes can commit alone while
+  // retaining enough information for a later request to recover an abandoned claim.
+  const claimResults = await Promise.all(
```
MathiasGruber: Three issues with the claim phase:

1. **No try/catch**: a single rejected claim UPDATE (transient PlanetScale error, deadlock victim, timeout) propagates out of this `Promise.all`, skipping both the rollback below and publish — already-claimed rows stay negative.

2. **No self-healing**: when that happens (or the process dies mid-protocol), the claimed stacks are invisible on every normal read — `getUserItems`/`getUserItemsWithVariants`/`fetchUpdatedUser` filter `gt(quantity, 0)` and never call the reaper, and there's no cron (`/api/cleaner` only handles deleted-user orphans). Recovery requires the player to press merge again or staff to open `getPublicUserItems`. The reaper belongs at a chokepoint every flow passes through (inside `fetchUserItems`/`fetchUserItemsWithVariants`, cheap on the userId index) or in the vercel.json cron set — otherwise this is a "my items vanished" support ticket generator.

3. **Statement count**: this issues K parallel UPDATEs per bucket (plus up to K per-row statements in each rollback path), when the publish below already demonstrates the single-statement CASE+OR pattern. One UPDATE with `SET quantity = CASE id … THEN -q … ELSE quantity END, updatedAt = claimedAt WHERE OR(mergeStackRowGuard(...))` and `rowsAffected === K` as the success check claims the whole bucket atomically; partial-claim rollback can then target `eq(quantity, -q) AND eq(updatedAt, claimedAt)` — strictly safer than the current restore guard (which omits the claimedAt stamp) and drops a 40-stack bucket from 42 statements to 3.

app/src/layout/MergeAllStacksButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const MergeAllStacksButton: React.FC<MergeAllStacksButtonProps> = ({
+  storedAtHome,
+  onMerged,
+}) => {
+  const utils = api.useUtils();
+
+  const { mutate: mergeAllStacks, isPending } = api.item.mergeAllStacks.useMutation({
+    onSuccess: async (data) => {
+      showMutationToast(data);
+      if (data.success) {
```
MathiasGruber, line 30: `mergeAllStacks` returns `success: false` for **partial** merges that did mutate the database ("Merged N item types, but M others did not complete (inventory may already show partial merges — try merge again)"), so gating invalidation on `data.success` leaves the UI stale in exactly the case where the server just told the user their inventory changed. The old inline code invalidated unconditionally. Every action on the stale rows (sell, split, equip, store) then fails with "Inventory changed, please refresh" until an unrelated refetch. Invalidate unconditionally (keep `onMerged` gated on success if you like), or have the server return a flag for "anything mutated" and gate on that.

app/src/app/items/page.tsx
```diff
@@ -1135,10 +1257,12 @@ const Character: React.FC<CharacterProps> = (props) => {
...
-  }));
+  const items = useritems
+    ?.map((useritem) => ({
+      ...applyActiveVariant(useritem as UserItemWithVariants),
+      ...useritem,
+    }))
+    .sort(byItemName);
   const itemBadges = userItemActionBadges(useritems);
   const equipped = items?.find((item) => item.equipped === slot);
   const repairItems = (useritems || []).filter(
```
MathiasGruber, line 1271: This filter still counts kits that are `storedAtHome` or `isInAuction`, but the server's `useRepairItem` now always rejects those ("Fetch the repair item from home storage first" / auction guard — new in this PR). So for a player whose only kits are at home, the equipped-item details modal shows an enabled "Use Repair Item" button and a picker full of kits that can only fail — a dead-end that worked before this PR. Needs the same `!storedAtHome && !isInAuction` exclusion here (the Backpack side is fine — its rows already pass `isEquippableUserItem`, which excludes both).

app/src/server/api/routers/item.ts
```diff
@@ -3318,8 +3538,20 @@ export const splitItemStack = async (
...
+      .update(userItem)
+      .set({ quantity: currentUserItem.quantity })
+      .where(
+        and(
+          eq(userItem.id, userItemId),
+          eq(userItem.userId, userId),
+          eq(userItem.quantity, quantityToKeep),
+        ),
+      );
+    return { success: false, message: "Could not create the split stack" };
```
MathiasGruber: This compensation treats any thrown insert as not-committed. On an ambiguous commit+throw (PlanetScale applies the INSERT but the HTTP response is lost), the catch restores the source to full quantity while the split row also exists — e.g. split 4 off a stack of 10 → user holds 14. The old code had no compensation, so this duplication path is new to the PR. Also, the restore's `eq(quantity, quantityToKeep)` guard is unchecked — if a concurrent claim changed the source in the window, a genuine insert failure silently leaves the user short instead. Low likelihood, but worth a comment acknowledging the two-generals residual, and the restore's `rowsAffected` should at least be checked/logged so the two outcomes are distinguishable.

MathiasGruber (commented):
I went through this in depth — the inventory UX improvements are good, but the new merge claim/publish protocol has a structural problem I think we need to solve before this merges, plus a handful of smaller issues. Inline comments below cover everything in the diff; this comment covers the protocol-level concern and the issues that live in files this PR doesn't touch.

## The core concern: claim state lives in the sign of `quantity`

Encoding "claimed" as a negated quantity (and tombstone identity in `updatedAt`) turns two business columns into distributed-protocol state that **every** current and future `userItem` query/mutation must remember to guard with `gt(quantity, 0)`. This PR patches ~23 sites in item.ts/home.ts/profile.ts, but the invariant is enforced by convention, and there are already five unpatched writers in the codebase (below) — each one is an item-duplication or item-loss trigger via the partial-publish problem flagged inline at the rollback in `executeMergeStacksForItemBucket`.

I'd rather we keep `quantity` always positive and move claim state to a dedicated column (e.g. `mergeClaimedAt`, or a claims side-table). That makes claims invisible to business logic by default, lets publish/rollback target rows by claim token exactly, makes the reaper's WHERE unambiguous, and kills the whole partial-publish class instead of chasing guard coverage forever.

## Unguarded `userItem` writers outside this diff

These all mutate rows the merge protocol may have claimed (negative quantity) or tombstoned (zero quantity), and none carry a `gt(quantity, 0)` guard:

1. **`auction.ts:312` / `auction.ts:415`** — `createAuctionListing`'s eligibility fetch and its `isInAuction=true` CAS have no quantity guard. Listing a claimed row flips `isInAuction` mid-protocol → the merge publish partially applies → the restored row's quantity also lives in the keeper, and the duplicate sits in an **active, sellable listing**. Worse: `quantity` is optional in `createAuctionListingSchema`, so a zero-tombstone can be listed with no quantity check; the flip bumps `updatedAt` (defeating the merge's tombstone-delete guard `eq(updatedAt, claimedAt)`), and ≥30s later the reaper's zero-row delete (which has no `isInAuction` exclusion) hard-deletes the listed row — `completeAuctionInternal` then transfers by bare `where(eq(id))` with no `rowsAffected` check, so the buyer pays for nothing. Both endpoints are MCP-enabled and status-unguarded, so one player can script this.

2. **`quests.ts:2730`** — `remove_item` consequences delete rows by bare id (`inArray`) from a pre-fetched snapshot. Deleting a claimed row mid-protocol → partial publish → the keeper still absorbs the deleted row's quantity, i.e. quest-confiscated items reappear in the keeper (or, roles reversed, the bucket loses the keeper's quantity).

3. **`libs/combat/database.ts:1432-1445`** (and the kage path at 461-477) — battle-end persistence writes absolute snapshot quantities with bare `where(eq(id))` and deletes unconditionally. `mergeStacks`/`mergeAllStacks` have no user-status guard (and are MCP-enabled), so a mid-battle merge is reachable: unequip mid-battle, merge, and the battle-end write later resurrects a tombstone whose quantity already lives in the keeper (duplication) or stomps a merged keeper back to its stale pre-merge quantity (loss). No tight race needed — the merge can complete minutes before the battle ends.

4. **`occupation.ts:471`** — the imbue mutation's write-time guard (`set({updatedAt: now}) where id/userId/isInAuction=false`) has no quantity predicate, so it matches claimed rows and tombstones: the crystal is consumed and the imbuement row attached to a tombstone the reaper will hard-delete within ~30s — crystal, imbuement, and crafting time silently lost. The `updatedAt` bump also postpones stale-claim recovery and defeats the tombstone-delete guard. This endpoint needs the same `gt(quantity, 0)` treatment toggleEquip/toggleStore/adjustUserItem got.

5. **`staff.ts:327`** — `unequipAllGear` bulk-writes `equipped='NONE'` fleet-wide with no quantity guard, mutating the `equipped` publish-guard column of any claimed row. Narrow window (needs a merge in flight during the rare staff sweep), but the missing guard is real and the failure mode is the same partial-publish corruption.

## Also worth knowing

A crashed merge currently has **no self-healing path** (details inline at the claim phase): the reaper never runs from `getUserItems`/`getUserItemsWithVariants`/`fetchUpdatedUser` or any cron, so claims abandoned by a transient error leave stacks invisible on every normal read until the player happens to press merge again or staff opens their item editor.

app/src/server/api/routers/item.ts
MathiasGruber: Addressed in f1dde58d5. The claim phase is now a single CASE+OR statement, and both failure restores go through one claim-scoped UPDATE that matches a row at either `-q` (still claimed) or its publish target, gated on `updatedAt = claimedAt` — so a partially applied publish is fully undone, including the published keeper. If a row has left both recoverable states (a concurrent writer moved it mid-window), the mismatch is logged via console.error for reconciliation instead of silently returning a retry toast. Combined with the quantity guards added to every remaining userItem writer (auction, quests, imbue, staff unequip, battle-end persistence), the concrete duplication/loss triggers are closed.

app/src/server/utils/concurrency.ts
MathiasGruber: Fixed in f1dde58d5 by removing the reap-from-refund path entirely: `refundUserItemQuantityAtomically` no longer calls the reaper at all. When the upsert no-ops because the row is merge-claimed, the refund is inserted as a brand-new row (fresh id, equipped NONE) that the in-flight merge cannot touch — the next merge simply folds it back into the stack. Active merges are never disturbed, and the zero-grace sweep is gone.

app/src/server/utils/concurrency.ts
MathiasGruber: Fixed in f1dde58d5 — the retry is gone along with the reaper call. The fallback path is now a plain insert of a new row: it cannot no-op against a claim, and if it fails it throws to the caller instead of silently dropping the refund.

app/src/server/api/routers/item.ts
MathiasGruber: Both fixed in f1dde58d5. `getRepairKits` now excludes `storedAtHome`/`isInAuction` stacks, which fixes the server selection and the inventory preview together (they share the helper). The consume block is back on `updateUserItemQuantityAtomically` — the delete-at-zero branch covers the full-stack case since `calculateKitsToUse` clamps `quantityUsed` to the stack quantity.

app/src/server/api/routers/item.ts
MathiasGruber: All three addressed in f1dde58d5. (1) The claim volley is one CASE+OR UPDATE (no partial claims possible), and the claim/publish phases sit in a try/catch whose handler runs the claim-scoped restore before returning a conflict. (2) `fetchUserItems`/`fetchUserItemsWithVariants` now self-heal: rows are fetched without the quantity filter, and only when a stale non-positive row is actually present does the reaper run followed by one refetch — the common case pays no extra round trip, and a crashed merge recovers on the player's next inventory load. Both merge endpoints also require AWAKE now, closing the mid-battle merge window. (3) The single-statement claim drops a K-stack bucket to 3 statements on the happy path.

app/src/layout/MergeAllStacksButton.tsx
MathiasGruber: Fixed in f1dde58d5 — the invalidations now run unconditionally in onSuccess (a partial merge mutates the database while reporting success: false), and only the `onMerged` callback stays gated on success.

app/src/app/items/page.tsx
MathiasGruber: Fixed in f1dde58d5 — both the Backpack and Character `repairItems` filters now exclude `storedAtHome`/`isInAuction` kits (Backpack's was already vacuously safe via `isEquippableUserItem`, but now states it explicitly), and `getRepairKits` carries the same exclusions.

app/src/server/api/routers/item.ts
MathiasGruber: Addressed in f1dde58d5 — the catch now resolves the ambiguity instead of assuming: it reads back the new row id, and only compensates when the insert genuinely did not land (a committed-but-thrown insert returns success, so the split can no longer duplicate). The restore's rowsAffected is also checked and logged when it cannot re-apply.

MathiasGruber:
Follow-up on the remaining points from my review, addressed in f1dde58d5:

**Unguarded writers** — all five now carry a `gt(quantity, 0)` guard so they can neither touch a merge-claimed row nor resurrect a tombstone: `createAuctionListing` (both the eligibility fetch and the `markInAuction` CAS), the quest `remove_item` delete, the imbue write-time guard in occupation.ts, staff `unequipAllGear`, and the battle-end item persistence (updates and deletes, both call sites). Both merge endpoints additionally require AWAKE status now, which closes the mid-battle merge window at the source rather than only at the writeback.

**Self-healing** — `fetchUserItems`/`fetchUserItemsWithVariants` detect stale claims/tombstones in the rows they already fetch and only then run the reaper plus one refetch, so a crashed merge recovers on the player's next inventory load with zero extra round trips in the common case. `getPublicUserItems` rides on that instead of carrying its own reaper call.

**On the dedicated claim column** — I'm deliberately not doing the schema migration in this PR. With the claim phase collapsed to one statement, the restore able to recover published rows, the refund path no longer un-claiming active merges, and every writer guarded, the concrete duplication/loss interleavings are closed without changing the data model. The sign encoding is still not pretty, and if we grow more userItem writers I'd revisit a `mergeClaimedAt` column (or claims table) so the invariant is structural rather than per-writer — but that's follow-up material, not a blocker for this PR.

MathiasGruber (approved):

Issue #80, opened 2024-03-19
Occupations(Anbu, Med-Nin, Sensei)


theeneon:
Their will be 3 main occupations in the game, Anbu, Medical Ninja and Sensei.

> **### Anbu:**
> To gain access to Anbu you must be Chuunin, and have the required village prestige. Anbu squads are created by the Kage or Chief of Anbu. A Village can have a maximum of 10 Anbu squads, each containing 20 persons. Each squad will have a name, a leader and a co leader. 
> 
> When players enter the Anbu hall, they will have a general Anbu Orders(Kage and Cheif are the only ones that can post these orders)  and each squad will have their personalize orders that he leader and co leader of the Anbu Squad can edit. 
> 
> Anbu Squads will be tracked by their kills and village funds and tokens gained. 
> 
> Anbu members receive an increase in tokens and village funds they also receive a 5% discount on ninja tools and hospital fees. 
> Only Anbu members, Kage, Village Elders can view the Anbu Hall and it’s orders 

> ### **Medical Ninja:**
> Med nin can be accessed at Genin, their 3 levels to it. Medical Ninjas will be able to heal persons who are in the hospital to full health and will be able to heal persons inside of combat. 
> 
> The amount a medical ninja can heal is dependent on their level and it will cost Chakra to do so. 
> Using basic heal, they will be to perform healing.  The chakra cost will scale with the amount of HP that is needed. 
> 
> 
> The healing will work as follow: 50 Chakra to heal 100 HP This would make it that it would take 2500 chakra to heal a fully capped user at 5000 HP. However the more experience the healer is their will be a percentage reduction. 
> 
> - Novice: No reduction. 
> - Apprentice: 10 reduction
> - Master 25 Reduction 
> 
> The experience requirement are as followed: 
> To reach Apprentice you need 5000 Exp
> To reach master you need 10000 Exp
> 
> For every time they heal someone they gain 10% of the total heal amount as exp. 
> The medical ninja will go to the hospital to heal someone, they will see a list of persons who are in the hospital the option to heal will be present, when they select it they can choose to heal 25,50,75 and 100% of the targets current health. 
> 
> To add more agency to the medical ninjas, hospital prices will go up and scale with HP at higher rate, when you pay to leave the hospital you will be fully healed . This is similar to core 2 to give the Medical Ninjas people to actually heal. 
> 
[...]

PR #1407, opened 2026-07-20
Quick travel buttons

Phrosfire:
# Pull Request

Add quick travel buttons for each village to the eye glass button

## License

## Summary by CodeRabbit

* **New Features**
  * Expanded Quick Travel with village, safezone, outlaw-hub, and War-Torn Battleground destinations.
  * Added alphabetical sorting, descriptions, readable color markers, and a scrollable destination list.
  * Added manual sector-ID travel with map-bound guidance.
  * Added PvP and sleep restriction details for War-Torn Battleground travel.

* **Improvements**
  * Prevented travel to the current sector and duplicate requests during active travel.
  * Improved map labels, highlights, and ownership indicators with automatic color contrast adjustments.
  * Improved map interaction reliability and handling of incomplete map data.


Phrosfire:
@coderabbitai review

Phrosfire:
@coderabbitai review

MathiasGruber:

**Conflicts resolved (1 file):**

- `app/src/app/travel/page.tsx` — `main` deleted the now-unused `currentSectorDimensions` / `onEdge` helpers (and the `isAtEdge` import) as part of the sector-window rework, while this branch had added `initiateTravelToSector` in the same block. Resolved by keeping only `initiateTravelToSector` and dropping the dead helpers, matching `main`.

Verified after rebase: typecheck clean.

app/src/app/travel/page.tsx
```diff
@@ -562,6 +571,16 @@ export default function Travel() {
     [currentStep, userData?.tutorialOn, startGlobalMove],
   );
 
+  // Open the world-travel confirm modal for a destination sector
+  const initiateTravelToSector = useCallback(
+    (sector: number) => {
+      if (sector === userData?.sector || isStartingTravel) return;
```
MathiasGruber: DRY: this helper wraps `setTargetSector` + `setShowModal`, but the map path still inlines the exact same two setters in `GlobalMap`'s `onTileClick` (`page.tsx` ~630-633), so it isn't the single entry point it looks like.

That duplication also leaves two different behaviours for the same intent: clicking your *own* tile on the globe still opens the confirm modal (the server then rejects with "You are already in that sector"), while the new list buttons silently no-op. Calling `initiateTravelToSector(sector)` from `onTileClick` collapses both.

Related: the `sector === userData?.sector || isStartingTravel` guard here is re-implemented verbatim in three `disabled` props below (lines 886-888, 907-910, 950-954). A small `canTravelTo(sector)` derived helper would let the guard live in one place.

app/src/app/travel/page.tsx
```diff
@@ -845,18 +864,67 @@ export default function Travel() {
...
                     <p className="pb-2 text-muted-foreground text-sm">
-                      Enter a sector ID to travel there directly.
+                      Travel to a village, or enter a sector ID.
                     </p>
+                    <div className="flex max-h-64 flex-col gap-1.5 overflow-y-auto pb-3">
+                      {travelDestinations.map((village) => {
+                        // Hyorin's stored color is near-white and invisible on the
+                        // popover background; show it as a light blue instead.
+                        const dotColor =
+                          village.name === "Hyorin" ? "#93c5fd" : village.hexColor;
```
MathiasGruber: Hardcoding a village *name* to patch its stored colour is fragile: rename Hyorin (it has already been renamed once - `Glacier` -> `Hyorin`, see `0009_violet_zemo.sql`) and the workaround silently stops applying, with no type error to catch it.

It also only fixes one row. `Village.hexColor` defaults to `#000000` (`schema.ts:3110`), so any village left on the default is equally invisible against a dark popover, and the same near-white colour is still used for Hyorin on the globe itself (`Map.tsx` highlights) where the stated readability problem also exists.

Either fix the stored `hexColor` in a migration, or derive the swatch generically (e.g. clamp luminance / add a neutral ring around the dot) so every village is handled by one rule.

app/src/app/travel/page.tsx
```diff
@@ -845,18 +864,67 @@ export default function Travel() {
...
+                              style={{ backgroundColor: dotColor }}
+                            />
+                            {village.mapName || village.name}
+                            <span className="ml-auto text-muted-foreground text-xs">
+                              {village.sector}
+                            </span>
+                          </Button>
+                        );
+                      })}
+                      <Button
```
MathiasGruber: This button is a hand-copied clone of the destination button above - same `type`/`size`/`variant`/`justify-start`, same colour-swatch + label + right-aligned sector number layout, same `=== userData?.sector || isStartingTravel` disable rule. Any future change to the row (tooltip, travel-time estimate, keyboard handling) now has to be made twice.

Cleaner to make it a synthetic entry appended to `travelDestinations` after the sort, e.g. `{ id: "war-torn", label: "War-Torn Battleground", sector: MAP_WAR_TORN_BATTLEGROUND_SECTOR, color: "#dc2626" }`, and render one list with a single `<Button>` template.

app/src/app/travel/page.tsx
```diff
@@ -866,7 +934,7 @@ export default function Travel() {
                               <FormControl>
                                 <Input
                                   className="w-full"
-                                  placeholder="Sector ID (0-491)"
+                                  placeholder="Sector ID"
```
MathiasGruber: Dropping the range removes the only hint about what's valid, and the input is still validated by `sectorIdSchema` (`MAP_SECTOR_ID_MIN`..`MAP_SECTOR_ID_MAX`), so an out-of-range entry now fails with no prior signal.

Per CLAUDE.md ("always import and use the actual constants"), prefer deriving it:
```tsx
placeholder={`Sector ID (${MAP_SECTOR_ID_MIN}-${MAP_SECTOR_ID_MAX})`}
```
Worth doing for the sibling "Find Sector" popover too (line 829): it still hardcodes `"Sector ID (0-491)"`, which is now wrong - `MAP_SECTOR_ID_MAX` is `MAP_WORLD_COLUMNS * MAP_WORLD_ROWS - 1` = 1943.

MathiasGruber (commented):
Review focused on logic correctness, readability and duplication.

No functional bugs found in the travel flow itself - the new destination list, the confirm-modal wiring and the outlaw/non-outlaw filtering all behave as intended. Four maintainability findings inline: the new `initiateTravelToSector` helper isn't used by the existing map-click path (leaving duplicated state updates and two different behaviours for the same action), the War-Torn button duplicates the destination-row markup, the Hyorin colour workaround hardcodes a village name for a data problem, and the sector-ID placeholder now hides a still-enforced validation range.

app/src/app/travel/page.tsx
```diff
@@ -609,8 +653,8 @@ export default function Travel() {
           focusSector={focusSector}
           focusSectorLabel="Target"
           onTileClick={(sector) => {
-            setTargetSector(sector);
-            setShowModal(true);
+            if (sector === null) return;
+            initiateTravelToSector(sector);
```
MathiasGruber, line 675: Routing the globe click through `initiateTravelToSector` is the right consolidation, but it puts state-dependent logic into a callback `GlobalMap` never refreshes, so the guard goes stale.

In `Map.tsx` the tile/label handlers (`onDblClick` ~line 404, `onTouchEnd` ~line 444, `onLabelPointerUp` ~line 501) live inside the big scene `useEffect` (starts ~line 146, dep array ~line 1112) and call `props.onTileClick?.(...)` off the `props` object captured when that effect last ran. The dep array is `[props.highlights, props.usersHighlighted, props.intersection, props.markerOnlyInteraction, props.focusSector, showOwnership, ownershipData, autoRotate]` — it does not include `onTileClick`, and none of those change when the player's sector changes (`villages` is memoized on `villageData`/`isOutlaw`, and react-query structural sharing keeps `trackedBounties` identity stable). So the scene keeps calling the *old* `initiateTravelToSector`, which closes over the *old* `canTravelTo`, which closes over the old `userData.sector` / `isStartingTravel`.

Concrete fallout after arriving in a new sector without a scene rebuild:
- Clicking the sector you just left silently does nothing (the stale guard still thinks it is your current sector), so you cannot travel back from the globe.
- Clicking your actual current sector still opens the confirm modal and gets rejected server-side — exactly the split behaviour this change was meant to remove.
- If the scene happens to rebuild while `isStartingTravel` is true, every tile click is dead until the next rebuild.

Adding `initiateTravelToSector` to the `MapComponent` memo deps (line 670) does not help: it produces a new element and a re-render, but the effect does not re-run, so the handler stays pinned to the old props.

Fix this in `Map.tsx` rather than here: keep the callback in a ref that is updated on every render and call it from the handlers, e.g.

```tsx
const onTileClickRef = useRef(props.onTileClick);
onTileClickRef.current = props.onTileClick;
```

then replace the three `props.onTileClick?.(sector, tile)` call sites with `onTileClickRef.current?.(sector, tile)` (the `if (props.intersection && props.onTileClick)` gate can stay). That keeps the globe path reading the current `canTravelTo` without rebuilding the scene.

app/src/app/travel/page.tsx
```diff
@@ -562,6 +590,22 @@ export default function Travel() {
...
+
+  // Open the world-travel confirm modal for a destination sector
+  const initiateTravelToSector = useCallback(
+    (sector: number) => {
+      if (!canTravelTo(sector)) return;
+      setTargetSector(sector);
+      setShowModal(true);
+    },
+    [canTravelTo],
+  );
```
MathiasGruber: `initiateTravelToSector` is an unstable identity (it depends on `canTravelTo`, which depends on `isStartingTravel` and `userData?.sector`), and it is now a dependency of the `MapComponent` memo (line 670). That memo exists specifically to keep the `GlobalMap` element stable — see its comment, "Memoized Map component to prevent re-renders during countdown updates" — so every time travel starts, finishes, or the player's sector changes, the memo is now discarded and `GlobalMap` re-renders.

Worse, the map never sees the fresher callback anyway. `GlobalMap`'s scene `useEffect` (Map.tsx, the effect starting at line 146) builds its `dblclick` / `touchend` / `pointerup` handlers around the `props` object captured at setup time, and its dependency array does not include `props.onTileClick`. So the globe keeps calling whatever `initiateTravelToSector` existed at the last scene rebuild — its `!isStartingTravel` check and its `sector !== userData?.sector` check both read stale values. The result is a memo invalidation that costs re-renders while delivering none of the guard it was added for (and can wrongly block travel to the sector you just left, or allow a duplicate start).

Suggest keeping the handler identity stable and reading the guard through a ref, e.g.:

```ts
const travelGuardRef = useRef({ sector: userData?.sector, isStartingTravel });
travelGuardRef.current = { sector: userData?.sector, isStartingTravel };

const initiateTravelToSector = useCallback((sector: number) => {
  const { sector: current, isStartingTravel: busy } = travelGuardRef.current;
  if (sector === current || busy) return;
  setTargetSector(sector);
  setShowModal(true);
}, []);
```

That gives an empty dep array, so `MapComponent` stays memoized, and the globe click path evaluates the guard against current state instead of a snapshot. The popover buttons can keep using `canTravelTo(...)` for their `disabled` state since those are re-rendered normally.

MathiasGruber: `initiateTravelToSector` returns silently when `canTravelTo` is false. That is fine for the popover buttons (they render disabled, so the user sees why), but the globe click path at line 655-657 now goes through the same guard, and there the failure is invisible: with `markerOnlyInteraction` the village label *is* the affordance, so a player standing in their own village who taps their own village marker gets literally no response. Previously that opened the confirm modal and the server answered with "You are already in that sector".

Suggest giving the no-op a reason, e.g.

```ts
const initiateTravelToSector = useCallback(
  (sector: number) => {
    if (isStartingTravel) return;
    if (sector === userData?.sector) {
      showMutationToast({ success: false, message: "You are already in that sector" });
      return;
    }
    setTargetSector(sector);
    setShowModal(true);
  },
  [userData?.sector, isStartingTravel],
);
```

app/src/app/travel/page.tsx
```diff
@@ -623,6 +667,7 @@ export default function Travel() {
     showOwnership,
     focusSector,
     userData?.tutorialOn,
+    initiateTravelToSector,
```
MathiasGruber: Adding `initiateTravelToSector` to this dep array does not do what it looks like it does. `GlobalMap` reads `props.onTileClick` only from inside its big scene `useEffect` (Map.tsx line 404 / 444 / 501), and that effect's dep array (Map.tsx ~1113-1122) lists `props.highlights`, `usersHighlighted`, `intersection`, `markerOnlyInteraction`, `focusSector`, `showOwnership`, `ownershipData`, `autoRotate` — not `onTileClick`. So the scene keeps whichever callback it captured on its last rebuild.

That did not matter before this PR, because the inline handler only called the stable `setTargetSector` / `setShowModal` setters. Now the handler closes over `userData?.sector` and `isStartingTravel`, so map clicks can be evaluated against a stale guard until something unrelated (a `village.getAll` refetch, an ownership toggle) happens to rebuild the scene. Meanwhile this new dep re-creates the `<GlobalMap>` element on every sector / pending change, which is exactly the churn this memo exists to avoid.

Cleanest fix is to keep the callback current inside `GlobalMap` instead of trying to rebuild the scene: hold it in a ref (`const onTileClickRef = useRef(props.onTileClick); onTileClickRef.current = props.onTileClick;`) and call `onTileClickRef.current?.(sector, tile)` from the three handlers, then drop `initiateTravelToSector` from this dep array.

app/src/app/travel/page.tsx
```diff
@@ -312,6 +316,30 @@ export default function Travel() {
...
+        color: v.hexColor,
+      }));
+    return [
+      ...villagesAsDestinations,
+      {
+        id: "war-torn",
+        label: "War-Torn Battleground",
+        sector: MAP_WAR_TORN_BATTLEGROUND_SECTOR,
+        color: "#dc2626",
+      },
```
MathiasGruber, line 341: The war-torn battleground is appended to the same list as villages and safezones and renders as an identical button, but it is not an equivalent destination. Sector 335 is a free-for-all: `combat.ts` (~line 1843) skips same-village, allied-village and XP-bracket restrictions for anyone standing there, and `home.ts` (~line 70) refuses to let you sleep there. The confirm modal only says "You are about to move from sector X to 335".

A low-level player who clicks the prettiest-sounding entry in a village list is one confirm away from a zone where anyone can attack them regardless of bracket, with no way to log out safely. Please differentiate it — at minimum a distinct label/tooltip ("free-for-all PvP, no sleeping") and an extra warning line in the confirm modal when `targetSector === MAP_WAR_TORN_BATTLEGROUND_SECTOR`.

Also, the `#dc2626` here is a third hardcoded red for this sector; `Map.tsx` already paints it with `warTweenColor = { r: 1, g: 0, b: 0 }`. Worth pulling into a constant next to `MAP_WAR_TORN_BATTLEGROUND_SECTOR` so the map highlight and the button agree.

app/src/layout/Map.tsx
```diff
@@ -639,8 +640,10 @@ const GlobalMap: React.FC<MapProps> = (props) => {
               canvas.height = h;
               const context = canvas.getContext("2d");
               if (context) {
+                const labelFill = getReadableVillageHexColor(highlight.hexColor);
+                const labelText = contrastingTextColor(labelFill);
                 context.globalAlpha = 0.9;
-                context.fillStyle = highlight.hexColor;
+                context.fillStyle = labelFill;
```
MathiasGruber, line 660: The label fill is now clamped, but the other two places in this same file that paint a village's colour are not, so the globe becomes internally inconsistent:

- line ~529-536: the ownership tile material uses `ownershipColorByVillageId.get(...)`, i.e. the raw `hexColor`
- line ~1131: `const hoveredOwnerColor = hoveredOwner?.hexColor ?? "#9ca3af";`

With the DB default of `#000000` for `village.hexColor`, a village that never picked a colour will now show a mid-grey label sitting on near-black territory, and the hover chip will use yet another shade. Since the point of ownership mode is "match the label to the territory", these should go through `getReadableVillageHexColor` too (or the clamp should be applied once where the colours are read, rather than per call site). Note line 1131's `"#9ca3af"` is also the exact fallback baked into `getReadableVillageHexColor`, so routing it through the helper removes that duplicated literal as well.

app/tests/utils/color.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  getReadableVillageHexColor,
+  parseHexColor,
+  relativeLuminance,
+} from "@/utils/color";
+
+test("parseHexColor accepts 3- and 6-digit hex", () => {
+  expect(parseHexColor("#fff")).toEqual({ r: 255, g: 255, b: 255 });
+  expect(parseHexColor("003366")).toEqual({ r: 0, g: 51, b: 102 });
+  expect(parseHexColor("not-a-color")).toBeNull();
+});
```
MathiasGruber, line 27: Two edges that matter in production aren't covered here:

1. **Accepted input formats.** `village.hexColor` is `varchar(191)` with no format constraint, and `parseHexColor` accepts a bare `fff` / `003366` (no `#`) while rejecting `#RRGGBBAA` and `rgb(...)`. Anything rejected silently becomes `#9ca3af`, so a stored 8-digit hex would quietly wipe out the village colour everywhere. Please pin the contract: `expect(parseHexColor("#ff000080")).toBeNull()` and `expect(parseHexColor("")).toBeNull()`.

2. **Hue preservation.** The doc comment on `getReadableVillageHexColor` promises the clamp keeps hue, but both clamping tests use `#FFFFFF` / `#000000`, which are achromatic — hue is meaningless there, so the HSL round-trip is never actually exercised. Add a saturated out-of-band case, e.g. `getReadableVillageHexColor("#000080")` (l = 0.25, below `MIN_LIGHTNESS`) should come back still blue (`b` clearly the dominant channel, `r === g === 0`), and something above `MAX_LIGHTNESS` like `#FFE0E0` should come back still red-dominant.

Asserting the resulting lightness lands on the `MIN_LIGHTNESS` / `MAX_LIGHTNESS` bounds rather than just "greater than black" would also make a future tweak of those constants fail loudly instead of silently.

MathiasGruber (approved):

PR #1409, opened 2026-07-22
Update banned word filter and comment failure handling

Phrosfire:
# Pull Request

Add and remove words from the banned word list.

## License

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved offensive-language detection by recognizing additional term variants and multi-word phrases.
  * Refined the moderation keyword list by removing duplicate or outdated entries and adding relevant variants for more consistent identification of prohibited content.
  * Improved comment reliability by preserving submitted text when validation fails and correctly updating or restoring comments after submission errors.

### DIFF
--- a/app/src/app/forum/[boardid]/[threadid]/page.tsx
+++ b/app/src/app/forum/[boardid]/[threadid]/page.tsx
@@ -66,8 +66,9 @@ export default function Thread(props: { params: Promise<{ threadid: string }> })
           const newPage = totalComments % limit === 0 ? totalPages : totalPages - 1;
           if (newPage !== page) {
             setPage(newPage);
+          } else {
+            await refetch();
           }
-          await refetch();
         }
       },
     });
--- a/app/src/app/forum/[boardid]/[threadid]/page.tsx
+++ b/app/src/app/forum/[boardid]/[threadid]/page.tsx
@@ -60,6 +60,7 @@ export default function Thread(props: { params: Promise<{ threadid: string }> })
     api.comments.createForumComment.useMutation({
       onSuccess: async (data) => {
         showMutationToast(data);
+        if (!data.success) return;
         reset();
         if (totalComments && totalPages && allComments) {
           const newPage = totalComments % limit === 0 ? totalPages : totalPages - 1;
--- a/app/src/layout/Conversation.tsx
+++ b/app/src/layout/Conversation.tsx
@@ -273,8 +273,8 @@ const Conversation: React.FC<ConversationProps> = (props) => {
     emoji: string,
     username: string,
   ) => {
-    await utils.comments.getConversationComments.cancel();
-    const cached = utils.comments.getConversationComments.getInfiniteData();
+    await utils.comments.getConversationComments.cancel(queryKey);
+    const cached = utils.comments.getConversationComments.getInfiniteData(queryKey);
     const comment = cached?.pages
       .flatMap((page) => page.data)
       .find((c) => c.id === commentId);
@@ -403,7 +403,7 @@ const Conversation: React.FC<ConversationProps> = (props) => {
     }
 
     // Optimistic update
-    await utils.comments.getConversationComments.cancel();
+    await utils.comments.getConversationComments.cancel(queryKey);
 
     const quoteText =
       quoteIds
--- a/app/src/layout/Conversation.tsx
+++ b/app/src/layout/Conversation.tsx
@@ -143,6 +143,29 @@ const Conversation: React.FC<ConversationProps> = (props) => {
       : composeRestriction;
   type ReturnedComment = ArrayElement<typeof allComments>;
 
+  const utils = api.useUtils();
+
+  /**
+   * Shared infinite-query cache writer for conversation comments. The
+   * `pages[0].convo` guard and ValidPage cast live here only — callers just
+   * map each page's `data` array.
+   */
+  const updateCachedComments = (
+    mapPageData: (data: ReturnedComment[], pageIndex: number) => ReturnedComment[],
+  ) => {
+    utils.comments.getConversationComments.setInfiniteData(queryKey, (old) => {
+      if (!old?.pages[0]?.convo) return old;
+      type ValidPage = Extract<(typeof old.pages)[number], { convo: object }>;
+      return {
+        pageParams: old.pageParams,
+        pages: (old.pages as ValidPage[]).map((page, i) => ({
+          ...page,
+          data: mapPageData(page.data, i),
+        })),
+      };
+    });
+  };
+
   // Search functionality
   const [isSearchOpen, setIsSearchOpen] = useState<boolean>(false);
 
@@ -250,43 +273,17 @@ const Conversation: React.FC<ConversationProps> = (props) => {
     emoji: string,
     username: string,
   ) => {
-    let old = utils.comments.getConversationComments.getInfiniteData();
     await utils.comments.getConversationComments.cancel();
-    utils.comments.getConversationComments.setInfiniteData(queryKey, (oldQueryData) => {
-      if (!oldQueryData) return undefined;
+    const cached = utils.comments.getConversationComments.getInfiniteData();
+    const comment = cached?.pages
+      .flatMap((page) => page.data)
+      .find((c) => c.id === commentId);
+    if (!comment) return;
 
-      // Guard: Don't update if there's no conversation
-      if (!oldQueryData.pages[0]?.convo) return oldQueryData;
-
-      // Find the comment looking at all pages in the oldQueryData
-      const comment = oldQueryData.pages
-        .flatMap((page) => page.data)
-        .find((c) => c.id === commentId);
-      if (!comment) return oldQueryData;
-
-      // Update the reactions
-      const newReactions = getNewReactions(comment.reactions, emoji, username);
-
-      // Type assertion: we know pages have valid convo after guard check
-      type ValidPage = Extract<(typeof oldQueryData.pages)[number], { convo: object }>;
-
-      old = {
-        pageParams: oldQueryData.pageParams,
-        pages: (oldQueryData.pages as ValidPage[]).map((page) => {
-          return {
-            ...page,
-            data: page.data.map((c) => {
-              if (c.id === commentId) {
-                return { ...c, reactions: newReactions };
-              }
-              return c;
-            }),
-          };
-        }),
-      };
-      return old;
-    });
-    return { old };
+    const newReactions = getNewReactions(comment.reactions, emoji, username);
+    updateCachedComments((data) =>
+      data.map((c) => (c.id === commentId ? { ...c, reactions: newReactions } : c)),
+    );
   };
 
   // Mutation for reactions
@@ -306,9 +303,6 @@ const Conversation: React.FC<ConversationProps> = (props) => {
   // Mutation for typing indicator
   const { mutate: sendTypingIndicator } =
     api.comments.sendTypingIndicator.useMutation();
-
-  // tRPC utils
-  const utils = api.useUtils();
 
   // infinite pagination
   useInfinitePagination({ fetchNextPage, hasNextPage, lastElement });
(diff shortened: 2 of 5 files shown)


app/src/utils/profanity.ts
```diff
@@ -274,8 +273,11 @@ export const OFFENSIVE_WORDS = new Set([
   "faeces",
   "fag",
   "fagot",
+  "faggot",
```
MathiasGruber, line 282: These four entries are inserted out of alphabetical order, in a list that is otherwise sorted. Measured on the post-merge file, only 6 of 537 adjacent pairs are out of order — and 2 of those 6 are introduced right here (`"fagot" > "faggot"` and `"faggots" > "faggotry"`). `"faggot"` sorts *before* `"fagot"` (`g` < `o`), and within the new group the order should be `faggot, faggotry, faggots, faggoty`.

This isn't cosmetic: sort order is the only mechanism keeping this 538-entry `Set` free of duplicates, and it has already failed. The file currently contains 6 duplicated entries — `assclown`, `asshole`, `beaner`, `boong`, `boonga`, `jigger` — each of which is silently swallowed by `new Set(...)`, so nothing surfaces them. Every out-of-order insertion makes the next reviewer's "is this word already listed?" scan unreliable and adds more of these.

Suggested placement:

```ts
  "fag",
  "faggot",
  "faggotry",
  "faggots",
  "faggoty",
  "fagot",
```

app/src/utils/profanity.ts
```diff
@@ -273,9 +267,12 @@ export const OFFENSIVE_WORDS = new Set([
   "escort",
   "faeces",
   "fag",
+  "faggot",
```
MathiasGruber, line 282: Anchoring here because this file is the only one in the diff — the actual defect is in `app/src/layout/Conversation.tsx`, in the `createConversationComment` mutation (roughly lines 491-531) and its `optimisticConversationUpdate` helper (lines 396-456).

Blocking a word via this list does not remove the message from the client. `comments.createConversationComment` returns the `checkForBadWords` rejection as a normal resolved value (`{ success: false, message }`), not a thrown tRPC error, so `onSuccess` runs with `data.success === false`, shows the toast, and leaves the optimistically inserted comment in the infinite-query cache. The blocked reply stays visible in the thread until a hard refresh, which reads as "it posted" to the player.

There is a second problem that also makes the existing `onError` rollback a no-op: `optimisticConversationUpdate` does `let old = utils.comments.getConversationComments.getInfiniteData()` and then **reassigns** `old` to the freshly built page data inside the `setInfiniteData` updater. The returned `context.old` is therefore the post-update state, so `setInfiniteData(queryKey, context?.old)` in `onError` restores the optimistic message rather than removing it.

Suggested fix — keep the snapshot and the new value in separate variables, return the snapshot, and roll back on both failure paths:

```ts
const previous = utils.comments.getConversationComments.getInfiniteData();
// ...build the updated pages into a separate local; never reassign `previous`
return { previous };
```

```ts
onSuccess: (data, _newComment, context) => {
  if (data.success) {
    // ...unchanged
  } else {
    utils.comments.getConversationComments.setInfiniteData(queryKey, context?.previous);
    showMutationToast({ success: false, message: data.message });
  }
},
onError: (error, _newComment, context) => {
  utils.comments.getConversationComments.setInfiniteData(queryKey, context?.previous);
  showMutationToast({ success: false, message: error.message });
},
```

Pre-existing rather than introduced here, but this PR is what exercises the path, so it is worth fixing alongside.

app/src/utils/profanity.ts
```diff
@@ -332,7 +329,6 @@ export const OFFENSIVE_WORDS = new Set([
   "hymie",
   "iceberg fuckers",
```
MathiasGruber, line 345: `checkForBadWords` tokenizes with `content.toLowerCase().match(/\b\w+\b/g)` and then does an exact `Set.has()` on each token, so no entry containing a space, hyphen or `$` can ever match — a token never contains those characters.

That makes 41 entries in this list permanently dead, including `iceberg fuckers` (this line), `towel-head`, `porch-monkey`, `half-breed`, `uncle-tom`, `carpet-muncher`, `welfare queen`, `moon-cricket` and `butt-fucker`. A few others (`butt-fuck`, `fuck-you`, `nigga$`) only get caught incidentally via their `fuck` / `nigga` component.

Since the list is being cleaned up anyway, either make them reachable or drop them so the list reflects what is actually filtered. Reachable version: split the set once at module load and check phrases as a consecutive token run, which handles hyphens and spaces with the same code path (`"towel-head"` tokenizes to `["towel","head"]`, same as the entry):

```ts
const PHRASES = [...OFFENSIVE_WORDS]
  .filter((w) => !/^\w+$/.test(w))
  .map((w) => w.split(/\W+/).filter(Boolean))
  .filter((p) => p.length > 1);
```

then scan `words` for each phrase as a consecutive slice.

app/src/utils/profanity.ts
```diff
@@ -273,9 +267,12 @@ export const OFFENSIVE_WORDS = new Set([
   "escort",
   "faeces",
   "fag",
+  "faggot",
+  "faggotry",
+  "faggots",
+  "faggoty",
```
MathiasGruber, line 285: Matching is exact whole-word, so each inflection has to be listed on its own — which is precisely why these four `faggot*` forms are needed here.

The same treatment is missing for the most common case in the list: `fuck` is present (line 290), but `fucking`, `fucked`, `fucker`, `fuckers` and `fucks` are not. So `"stop fucking around"` and `"what a fucker"` pass moderation today while `"fuck this"` is blocked. Worth adding those forms while you are in here.

app/src/utils/profanity.ts
```diff
@@ -273,9 +267,12 @@ export const OFFENSIVE_WORDS = new Set([
...
   "fagot",
   "fannyfucker",
-  "farting",
   "fastfuck",
   "fat",
```
MathiasGruber: This PR drops `charlie`, `illegal` and `farting` as over-broad, but the same class of entry is still present and some are worse. The check blocks the whole comment (and is also applied to usernames in `register.ts`), so each of these is a hard false positive on ordinary text:

- `fat` (this line) — "fat chance", "fat damage bonus"
- `escort` (line 267) — "escort quest" is standard game vocabulary
- `niggles` / `niggling` / `nigglings` (lines 397-399) and `sniggers` (line 490) — ordinary English words with no relation to the slur they sit next to; `nigger` / `niggers` are already listed separately
- `aboriginal` (line 29) — a neutral demonym; blocking it is its own problem

Suggest removing at least the `niggle` / `snigger` entries and `aboriginal` in this pass.

Phrosfire: I remove fat, escort, and aboriginal.  The others I kept because people will see it as a "Valid" workaround

app/src/layout/Conversation.tsx
```diff
@@ -500,24 +514,37 @@ const Conversation: React.FC<ConversationProps> = (props) => {
...
-              utils.comments.getConversationComments.setInfiniteData(
-                queryKey,
-                newComment,
-              );
-            }
+          if (data.commentId && context?.optimisticCommentId) {
+            // Stamp the server id onto the matching optimistic comment only.
+            const { commentId } = data;
+            const { optimisticCommentId } = context;
+            utils.comments.getConversationComments.setInfiniteData(
```
MathiasGruber: This inline cache writer (521-539) is the fourth copy of the same block in this file: the `!oldQueryData?.pages[0]?.convo` guard, the `type ValidPage = Extract<...>` assertion and the `{ pageParams, pages: (... as ValidPage[]).map(...) }` rebuild also appear at 271-286 (`optimisticReactionUpdate`), 463-476 (`optimisticConversationUpdate`) and 486-493 (`removeOptimisticComment`). The PR adds two of those four.

Worth collapsing into one local helper, e.g.

```ts
const updateCachedComments = (
  mapPageData: (data: ReturnedComment[]) => ReturnedComment[],
) => {
  utils.comments.getConversationComments.setInfiniteData(queryKey, (old) => {
    if (!old?.pages[0]?.convo) return old;
    type ValidPage = Extract<(typeof old.pages)[number], { convo: object }>;
    return {
      pageParams: old.pageParams,
      pages: (old.pages as ValidPage[]).map((page) => ({
        ...page,
        data: mapPageData(page.data),
      })),
    };
  });
};
```

Then this becomes `updateCachedComments((data) => data.map((c) => c.id === optimisticCommentId ? { ...c, id: commentId } : c))`, `removeOptimisticComment` becomes a one-line `filter`, and the reaction/insert paths can use it too. The `ValidPage` assertion then lives in exactly one place, which matters because it is an unchecked cast that only holds under that guard.

app/src/layout/Conversation.tsx
```diff
@@ -392,19 +392,19 @@ const Conversation: React.FC<ConversationProps> = (props) => {
...
     if (searchQuery && "comment" in newMessage) {
       if (!newMessage.comment.toLowerCase().includes(searchQuery.toLowerCase())) {
-        return { old };
+        return {};
       }
     }
 
+    // Temp id only when we actually insert an optimistic comment. Success/failure
+    // reconcile by this id so we never touch an unrelated cache entry.
+    const optimisticCommentId = "id" in newMessage ? undefined : nanoid();
```
MathiasGruber: The comment says the id is minted "only when we actually insert an optimistic comment", but it is minted before `setInfiniteData` runs, and the updater still bails out at 418 (`if (!oldQueryData) return undefined`) and 421 (`if (!oldQueryData.pages[0]?.convo) return oldQueryData`). In those cases nothing is inserted yet the context reports an id, so the success/error handlers do a pointless cache write. It also forces the two workarounds it introduces: the `&& optimisticCommentId` at 427 (which, if it were ever false, would silently build a comment with empty `content`) and the `optimisticCommentId!` non-null assertion at 437 — TS cannot narrow `string | undefined` from the separate `"id" in newMessage` check.

The whole `next` object only depends on component scope (`newMessage`, `quoteIds`, `allComments`, `userData`, `conversation`), not on `oldQueryData`, so build it before the updater and let the type flow:

```ts
const optimisticComment = "id" in newMessage ? null : { id: nanoid(), createdAt: new Date(), /* ... */ content: processMentions(quoteText + newMessage.comment).processedContent };
const next = optimisticComment ?? newMessage;
```

That drops the `!`, drops the redundant `&& optimisticCommentId` guard, and lets you return `{ optimisticCommentId: optimisticComment?.id }` only on the path that actually inserted.

app/src/utils/profanity.ts
```diff
@@ -559,3 +567,18 @@ export const OFFENSIVE_WORDS = new Set([
...
+
+/**
+ * Multi-token phrases derived from list entries that contain non-word characters.
+ * "towel-head" and "welfare queen" both become consecutive token runs that match
+ * the same way content is tokenized (`\b\w+\b`).
+ */
+const OFFENSIVE_PHRASES = [...OFFENSIVE_WORDS]
+  .filter((w) => !/^\w+$/.test(w))
+  .map((w) => w.split(/\W+/).filter(Boolean))
+  .filter((p) => p.length > 1);
```
MathiasGruber: `.filter((p) => p.length > 1)` silently discards any entry that normalises to a single token, so an entry lands in neither derived set. Today that hits `nigga$` and `n1gga$` — both happen to be covered by the plain `nigga`/`n1gga` entries, so there is no coverage gap right now, but it is a quiet failure mode: add `sh*t` or `b!tch` to `OFFENSIVE_WORDS` later and it will be dropped with no signal.

Route single-token results back into the word set instead of dropping them:

```ts
const normalized = [...OFFENSIVE_WORDS].map((w) => w.split(/\W+/).filter(Boolean));
const OFFENSIVE_SINGLE_WORDS = new Set(
  normalized.filter((p) => p.length === 1).map((p) => p[0]!),
);
const OFFENSIVE_PHRASES = normalized.filter((p) => p.length > 1);
```

One pass, no entry can fall through, and it removes the duplicated `/^\w+$/` predicate on 573 and 582.

app/src/layout/Conversation.tsx
```diff
@@ -500,24 +514,37 @@ const Conversation: React.FC<ConversationProps> = (props) => {
...
         } else {
+          removeOptimisticComment(context?.optimisticCommentId);
           showMutationToast({ success: false, message: data.message });
         }
       },
       onError: (error, _newComment, context) => {
-        utils.comments.getConversationComments.setInfiniteData(queryKey, context?.old);
+        removeOptimisticComment(context?.optimisticCommentId);
         showMutationToast({ success: false, message: error.message });
       },
```
MathiasGruber, line 506: Anchoring here because this is the nearest changed line — the defect is in `app/src/app/forum/[boardid]/[threadid]/page.tsx` around line 61, in the `createForumComment` `onSuccess` handler.

This PR gives the inbox path the right behaviour: on `data.success === false` the optimistic comment is removed and the composer is left alone, so the player keeps their text. The forum reply flow does not match. Its `onSuccess` calls `reset()` unconditionally:

```ts
onSuccess: async (data) => {
  showMutationToast(data);
  reset();
  // ...
},
```

`createForumComment` returns the `checkForBadWords` rejection as a resolved `{ success: false, message }` rather than throwing, so a moderated reply still reaches this branch, shows the failure toast, and then wipes the composer. Reproduced on the preview: typing a reply containing `camel jockey` in a forum thread clears the box, while the same phrase in inbox chat keeps the draft.

This PR is what makes it bite — the phrase matching added here means many more submissions now land on the failure path, including multi-word entries that previously could never match at all.

Guard the success-only work the same way the inbox does:

```ts
onSuccess: async (data) => {
  showMutationToast(data);
  if (!data.success) return;
  reset();
  // ...page/refetch handling
},
```

app/src/app/forum/[boardid]/[threadid]/page.tsx
```diff
@@ -60,6 +60,7 @@ export default function Thread(props: { params: Promise<{ threadid: string }> })
     api.comments.createForumComment.useMutation({
       onSuccess: async (data) => {
         showMutationToast(data);
+        if (!data.success) return;
```
Comment (MathiasGruber) [inline app/src/app/forum/[boardid]/[threadid]/page.tsx:63]: This guard fixes the *create* path, but the *edit* path still has the same problem, so moderation rejection behaves inconsistently depending on how the user got there.

The actual location is `app/src/layout/Comment.tsx:173-176` (not part of this diff, so anchoring here on the closest related change):

```ts
const onSubmit = handleSubmit((data) => {
  if (props.editComment) props.editComment(data);
  props.setEditing(false);
});
```

`setEditing(false)` runs unconditionally at submit time, before the mutation resolves. Consequence: submitting an edit that fails moderation shows the rejection toast, but the inline editor has already closed and the original comment is back on screen, with no inline error and no obvious way to recover what was typed. Failed conversation/forum edits and failed deletes all collapse the editor the same way.

It also makes the success-only closes redundant — `setEditing(false)` inside the `onSuccess` handlers at `Comment.tsx:63`, `73`, `108` and `118` is guarded by `data.success`, but the editor is already closed by then, so those guards never do anything.

Suggested fix, mirroring the pattern used here: drop `props.setEditing(false)` from `onSubmit` and let the existing `if (data.success) { ... setEditing(false) }` blocks own the transition.

app/src/app/forum/[boardid]/[threadid]/page.tsx
```diff
@@ -60,6 +60,7 @@ export default function Thread(props: { params: Promise<{ threadid: string }> })
...
         reset();
         if (totalComments && totalPages && allComments) {
           const newPage = totalComments % limit === 0 ? totalPages : totalPages - 1;
```
Comment (MathiasGruber) [inline app/src/app/forum/[boardid]/[threadid]/page.tsx:66]: Anchoring here since the offending line (`await refetch()`, a few lines below) is outside the diff hunk, but it is in the block this PR edits.

When `newPage !== page`, this fires two network round-trips instead of one. `setPage(newPage)` only schedules a re-render; `refetch()` then runs synchronously against the observer's *current* options, so it refetches the page the user is leaving. The subsequent re-render swaps the query key to `cursor: newPage` and issues a second fetch for the page actually being shown. The first response is discarded. `getForumComments` does a count plus a join per call, so that is a wasted server round-trip on roughly every `limit`-th post (whenever `totalComments % limit === 0` pushes the user onto a new page).

Only refetch when the page did not change — the page change fetches on its own:

```ts
if (newPage !== page) {
  setPage(newPage);
} else {
  await refetch();
}
```

app/src/layout/Conversation.tsx
```diff
@@ -250,43 +273,17 @@ const Conversation: React.FC<ConversationProps> = (props) => {
...
-              }
-              return c;
-            }),
-          };
-        }),
-      };
-      return old;
-    });
-    return { old };
+    const cached = utils.comments.getConversationComments.getInfiniteData();
```
MathiasGruber: `getInfiniteData()` is called with no input, so tRPC builds the key `[path, { type: "infinite" }]`, while the cached entry lives under `[path, { input: queryKey, type: "infinite" }]` (`getQueryKeyInternal` in @trpc/react-query 11.17.0, and `getInfiniteData` resolves to `queryClient.getQueryData(key)` which matches keys exactly). So `cached` is always `undefined`, `comment` is never found, and the function returns at the `if (!comment) return;` guard. Net effect: optimistic reaction updates never apply, and incoming pusher `reaction` events (which also route through this function) never render until a refetch.

This is a regression: on main the read was only used for the unused rollback context, and the actual update ran inside the `setInfiniteData` updater, which does receive the real cache data.

Fix: pass the key, matching what `optimisticConversationUpdate` already does on line 453 — `utils.comments.getConversationComments.getInfiniteData(queryKey)`.

app/src/utils/profanity.ts
```diff
@@ -4,15 +4,26 @@
...
-  if (found) {
+  const foundWord = words.find((word) => OFFENSIVE_SINGLE_WORDS.has(word));
+  if (foundWord) {
     return {
       success: false,
-      message: `Your comment was flagged for inappropriate language and will not be shown to others. Details: ${found}`,
+      message: `Your comment was flagged for inappropriate language and will not be shown to others. Details: ${foundWord}`,
     };
   }
+  for (const phrase of OFFENSIVE_PHRASES) {
```
MathiasGruber: Phrase matching runs over `words`, which has had all punctuation stripped by the `\b\w+\b` tokenizer, so a phrase matches across sentence and clause boundaries: `"Is the bar full? Blood everywhere."` tokenizes to `[..., "full", "blood", ...]` and gets flagged by the `full-blood` entry.

This matters because the change activates 47 multi-token entries that were previously dead (no token ever contained a space or hyphen, so those entries could never match). Several are generic word pairs that collide with normal game chat — the `full-blood` entry ("blood"/"bloodline" is core game vocabulary), the `bounty-bar` entry (the game has both a bounty system and a bar), plus `ali baba`, `choc-ice`, `ang mo`. Given the rest of the PR removes false-positive-prone singles (`fat`, `escort`, `illegal`, `aboriginal`, `charlie`), letting these through undercuts the goal.

Suggest tokenizing into punctuation-delimited runs (e.g. `content.toLowerCase().split(/[^\w]+|[.,!?;:]/)`-style segmentation, matching phrases only within a run) so a phrase can't straddle a sentence break, and dropping the low-signal generic pairs from the list.

MathiasGruber (approved):

PR #1417, opened 2026-07-26
Apply limit on heal jutsu

Phrosfire:
# Pull Request

Applies a limit on the number of equipped heal jutsu.

Mass unequip jutsu/items now unequips all inactive loadouts as well.
Mass unequip jutsu/items no longer unequip them from AI

Require forbidden jutsu to be learned by scrolls instead of through training

## License

## Summary by CodeRabbit

* **New Features**
  * Added limits for equipping healing jutsu in ranked PvP loadouts.
  * Ranked loadout screens now display healing jutsu restrictions and warnings.
  * Jutsu management prevents equipping or automatically selecting healing jutsu beyond the allowed limit.
  * Staff bulk unequipping now clears affected non-AI players’ gear, jutsu, and loadouts while leaving AI opponents unchanged.

* **Bug Fixes**
  * Improved reliability of jutsu equip and loadout updates during simultaneous actions.

* **Documentation**
  * Updated the Ranked PvP manual with healing jutsu limits.

* **Tests**
  * Added coverage verifying healing jutsu equip limits are enforced.

### DIFF
--- a/app/src/libs/ranked_pvp.ts
+++ b/app/src/libs/ranked_pvp.ts
@@ -1,19 +1,11 @@
 import type { RankedRank } from "@/drizzle/constants";
 import {
-  JUTSU_MAX_EVENT_EQUIPPED,
   RANKED_DIVISIONS,
   RANKED_LEGEND_LP_REQUIREMENT,
-  RANKED_LOADOUT_MAX_BARRIER_JUTSUS,
   RANKED_LOADOUT_MAX_CONSUMABLES,
   RANKED_LOADOUT_MAX_INCREASECOST_ITEMS,
-  RANKED_LOADOUT_MAX_INCREASECOST_JUTSUS,
   RANKED_LOADOUT_MAX_JUTSUS,
   RANKED_LOADOUT_MAX_POISON_ITEMS,
-  RANKED_LOADOUT_MAX_POISON_JUTSUS,
-  RANKED_LOADOUT_MAX_RESIDUAL_JUTSUS,
-  RANKED_LOADOUT_MAX_SHIELD_JUTSUS,
-  RANKED_LOADOUT_MAX_STUN_JUTSUS,
-  RANKED_LOADOUT_MAX_SUMMON_JUTSUS,
   RANKED_LOADOUT_MAX_WEAPONS,
   RANKED_MIN_LP_GAIN,
   RANKED_QUEUE_MAX_WAIT_SECS,
@@ -22,6 +14,7 @@ import {
   RANKED_STREAK_BONUS,
 } from "@/drizzle/constants";
 import type { Item, Jutsu, UserData } from "@/drizzle/schema";
+import { getJutsuCapFlags, getJutsuCategoryDef, RANKED_JUTSU_CAPS } from "@/libs/jutsu";
 
 /**
  * Determine player rank based on LP and top players
@@ -149,74 +142,14 @@ export const validateJutsuLoadout = (jutsus: Jutsu[]) => {
   let check = true;
   let message = "";
 
-  // Check residual jutsu limit
-  const residualJutsus = jutsus.filter((jutsu) =>
-    jutsu.effects.some((e) => "residualModifier" in e && e.residualModifier),
-  );
-  if (residualJutsus.length > RANKED_LOADOUT_MAX_RESIDUAL_JUTSUS) {
-    check = false;
-    message = `You can only equip up to ${RANKED_LOADOUT_MAX_RESIDUAL_JUTSUS} residual jutsu in ranked PvP`;
-  }
-
-  // Check poison jutsu limit
-  const poisonJutsus = jutsus.filter((jutsu) =>
-    jutsu.effects.some((e) => e.type === "poison"),
-  );
-  if (poisonJutsus.length > RANKED_LOADOUT_MAX_POISON_JUTSUS) {
-    check = false;
-    message = `You can only equip up to ${RANKED_LOADOUT_MAX_POISON_JUTSUS} poison jutsu in ranked PvP`;
-  }
-
-  // Check increasecost jutsu limit
-  const increasecostJutsus = jutsus.filter((jutsu) =>
-    jutsu.effects.some((e) => e.type === "increasepoolcost"),
-  );
-  if (increasecostJutsus.length > RANKED_LOADOUT_MAX_INCREASECOST_JUTSUS) {
-    check = false;
-    message = `You can only equip up to ${RANKED_LOADOUT_MAX_INCREASECOST_JUTSUS} increasecost jutsu in ranked PvP`;
-  }
-
-  // Check summon jutsu limit
-  const summonJutsus = jutsus.filter((jutsu) =>
-    jutsu.effects.some((e) => e.type === "summon"),
-  );
-  if (summonJutsus.length > RANKED_LOADOUT_MAX_SUMMON_JUTSUS) {
-    check = false;
-    message = `You can only equip up to ${RANKED_LOADOUT_MAX_SUMMON_JUTSUS} summon jutsu in ranked PvP`;
-  }
-
-  // Check barrier jutsu limit
-  const barrierJutsus = jutsus.filter((jutsu) =>
-    jutsu.effects.some((e) => e.type === "barrier"),
-  );
-  if (barrierJutsus.length > RANKED_LOADOUT_MAX_BARRIER_JUTSUS) {
-    check = false;
-    message = `You can only equip up to ${RANKED_LOADOUT_MAX_BARRIER_JUTSUS} barrier jutsu in ranked PvP`;
-  }
-
-  // Check stun jutsu limit
-  const stunJutsus = jutsus.filter((jutsu) =>
-    jutsu.effects.some((e) => e.type === "stun"),
-  );
-  if (stunJutsus.length > RANKED_LOADOUT_MAX_STUN_JUTSUS) {
-    check = false;
-    message = `You can only equip up to ${RANKED_LOADOUT_MAX_STUN_JUTSUS} stun jutsu in ranked PvP`;
-  }
-
-  // Check shield jutsu limit
-  const shieldJutsus = jutsus.filter((jutsu) =>
-    jutsu.effects.some((e) => e.type === "shield"),
-  );
-  if (shieldJutsus.length > RANKED_LOADOUT_MAX_SHIELD_JUTSUS) {
-    check = false;
-    message = `You can only equip up to ${RANKED_LOADOUT_MAX_SHIELD_JUTSUS} shield jutsu in ranked PvP`;
-  }
-
-  // Check event jutsu limit
-  const eventJutsus = jutsus.filter((jutsu) => jutsu.jutsuType === "EVENT");
-  if (eventJutsus.length > JUTSU_MAX_EVENT_EQUIPPED) {
-    check = false;
-    message = `You can only equip up to ${JUTSU_MAX_EVENT_EQUIPPED} event jutsu in ranked PvP`;
+  const flagged = jutsus.map((jutsu) => getJutsuCapFlags(jutsu));
+  for (const cap of RANKED_JUTSU_CAPS) {
+    const count = flagged.filter((flags) => flags[cap.key]).length;
+    if (count > cap.max) {
+      check = false;
+      const label = getJutsuCategoryDef(cap.key).label;
+      message = `You can only equip up to ${cap.max} ${label} jutsu in ranked PvP`;
+    }
   }
 
   if (jutsus.length > RANKED_LOADOUT_MAX_JUTSUS) {
--- a/app/src/server/api/routers/jutsu.ts
+++ b/app/src/server/api/routers/jutsu.ts
@@ -1079,19 +1079,13 @@ export const jutsuRouter = createTRPCRouter({
         questDataFull = trackers;
       }
       const questDataForDb = filterQuestTrackersForDbPersist(questDataFull, user);
-      // Pre-mutation snapshot for refund if the new-jutsu insert fails after deduct.
-      const questDataBeforeTrain = filterQuestTrackersForDbPersist(
-        user.questData ?? [],
-        user,
-      );
 
-      // Deduct money
+      // Deduct money first, but persist new-jutsu quest progress only after the
+      // ownership insert is confirmed. A duplicate request can then refund its
+      // own money without restoring a stale questData snapshot over the winner.
       const moneyUpdate = await ctx.drizzle
         .update(userData)
-        .set({
-          money: sql`${userData.money} - ${trainCost}`,
-          questData: questDataForDb,
-        })
+        .set({ money: sql`${userData.money} - ${trainCost}` })
         .where(and(eq(userData.userId, ctx.userId), gte(userData.money, trainCost)));
       if (moneyUpdate.rowsAffected !== 1) {
         return errorResponse("You don't have enough money");
@@ -1123,17 +1117,15 @@ export const jutsuRouter = createTRPCRouter({
         const newUserJutsuId = nanoid();
         // onDuplicateKeyUpdate no-op (`id = id`) absorbs unique-index races. PlanetScale
         // reports changed rows, so a duplicate yields rowsAffected === 0 (not 2). Deduct
-        // already happened above — refund money + quest snapshot when this request did
-        // not insert (otherwise jutsus_mastered / train_specific_jutsu stay incremented).
+        // already happened above, so refund it when this request definitively did not
+        // insert. Quest progress has not been persisted yet and needs no compensation.
         const refundTrainCost = () =>
           ctx.drizzle
             .update(userData)
-            .set({
-              money: sql`${userData.money} + ${trainCost}`,
-              questData: questDataBeforeTrain,
-            })
+            .set({ money: sql`${userData.money} + ${trainCost}` })
             .where(eq(userData.userId, ctx.userId));
 
+        let insertedByThisRequest = false;
         try {
           const insertResult = await ctx.drizzle
             .insert(userJutsu)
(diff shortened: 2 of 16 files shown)


app/src/server/api/routers/staff.ts
```diff
@@ -273,17 +274,31 @@ export const staffRouter = createTRPCRouter({
...
           .update(userJutsu)
           .set({ equipped: false })
-          .where(ne(userJutsu.equipped, false)),
-        ctx.drizzle.update(jutsuLoadout).set({ jutsuIds: [] }),
+          .where(
+            and(
+              ne(userJutsu.equipped, false),
+              sql`${userJutsu.userId} IN (
+                SELECT ${userData.userId} FROM ${userData} WHERE ${userData.isAi} = false
+              )`,
```
MathiasGruber: `IN (SELECT userId FROM UserData WHERE isAi = false)` makes the engine materialize a set containing essentially *every human account* in the game, and it does so for both statements in this `Promise.all` (and again in `unequipAllGear`). The `UserData_isAi_idx` index does not help here — `isAi = false` matches almost the whole table, so this degenerates into a full index scan plus a large temp table, on top of an already table-wide UPDATE.

AI users are a tiny fraction of `UserData`, so invert the predicate and let the index do real work:

```ts
sql`${userJutsu.userId} NOT IN (
  SELECT ${userData.userId} FROM ${userData} WHERE ${userData.isAi} = true
)`
```

That materializes a few hundred ids instead of millions. One semantic difference worth confirming is intended: rows whose `userId` no longer exists in `UserData` (orphans) would now be included — harmless for an unequip-all, but call it out if you'd rather keep them excluded.

app/src/server/api/routers/staff.ts
```diff
@@ -295,14 +310,32 @@ export const staffRouter = createTRPCRouter({
...
+      await Promise.all([
+        ctx.drizzle
+          .update(userItem)
+          .set({ equipped: "NONE" })
+          .where(
+            and(
+              ne(userItem.equipped, "NONE"),
+              sql`${userItem.userId} IN (
+                SELECT ${userData.userId} FROM ${userData} WHERE ${userData.isAi} = false
+              )`,
```
MathiasGruber: Same as in `unequipAllJutsus`: both subqueries here (this one and the `itemLoadout` one on line 331-333) enumerate every non-AI user. Flip to `NOT IN (SELECT userId FROM UserData WHERE isAi = true)` so the `isAi` index selects the small side instead of scanning the entire user base twice per invocation.

app/src/server/api/routers/jutsu.ts
```diff
@@ -2285,3 +2404,286 @@ export const selectJutsuLoadout = async (
...
+
+/**
+ * Materialized COUNT subquery for MySQL UPDATE guards. Wrapping in a derived table
+ * avoids ERROR 1093 (can't select from the table being updated in a subquery).
+ */
+const equippedCountSql = (userId: string) => sql`(
+  SELECT cnt FROM (
+    SELECT COUNT(*) AS cnt
+    FROM ${userJutsu} AS uj_tot
+    WHERE uj_tot.userId = ${userId} AND uj_tot.equipped = true
```
MathiasGruber: These count subqueries now run on the hot equip path (every `toggleEquip` and every auto-equip in `startTraining`), and they all filter on `userId = ? AND equipped = true`. There is no index supporting that combination: `UserJutsu` has a unique `(userId, jutsuId)` index and a standalone `Jutsu_equipped_idx` on `equipped` alone (useless — boolean, ~50/50). So each of these subqueries walks *all* of the user's `UserJutsu` rows (veteran accounts own hundreds) and filters `equipped` afterwards, then joins to `Jutsu` and evaluates `JSON_SEARCH`/`JSON_EXTRACT` on the survivors.

Worth adding a composite index `(userId, equipped)` in `app/drizzle/schema.ts` + migration so these guards touch ~20 rows rather than the user's whole jutsu collection.

app/src/server/api/routers/jutsu.ts
```diff
@@ -2285,3 +2404,286 @@ export const selectJutsuLoadout = async (
...
+    );
+  }
+  if (flags.isHeal) {
+    underCapGuards.push(
+      sql`${equippedEffectTypeCountSql(userId, "heal")} < ${JUTSU_MAX_HEAL_EQUIPPED}`,
+    );
+    overCapGuards.push(
+      sql`${equippedEffectTypeCountSql(userId, "heal")} > ${JUTSU_MAX_HEAL_EQUIPPED}`,
+    );
+  }
```
MathiasGruber: Every cap flag appends its own independent scalar subquery, and the same set is rebuilt for `overCapGuards`. A jutsu that is, say, heal + stun + residual produces 4 separate aggregate subqueries in the equip UPDATE and 4 more in the rollback UPDATE — 8 independent scans of `UserJutsu ⨝ Jutsu` for one equip click. This lands on a path that previously cost a single `Promise.all` of two trivial UPDATEs and now costs 2-5 strictly sequential round-trips (`tryEquipUserJutsuAtomically` alone is always 2 statements when the pre-check passes), so per-statement cost matters more than it used to.

Suggest computing all counts once per statement in a single derived table and referencing its columns, e.g.

```sql
(SELECT total FROM (
   SELECT COUNT(*) AS total,
          SUM(<heal predicate>)   AS heal_cnt,
          SUM(<stun predicate>)   AS stun_cnt,
          ...
   FROM UserJutsu uj JOIN Jutsu j ON j.id = uj.jutsuId
   WHERE uj.userId = ? AND uj.equipped = true
 ) c)
```

One join + one pass, with each guard reading a different column of the same derived table, instead of N full re-scans per statement.

app/src/server/api/routers/jutsu.ts
```diff
@@ -2285,3 +2404,286 @@ export const selectJutsuLoadout = async (
...
+const EQUIPPED_RESIDUAL_EFFECT_SCAN_WINDOW = 16;
+const equippedResidualCountSql = (userId: string) => {
+  const hasResidualModifier = sql.join(
+    Array.from({ length: EQUIPPED_RESIDUAL_EFFECT_SCAN_WINDOW }, (_, i) =>
+      sql.raw(
+        `IFNULL(JSON_EXTRACT(j_res.effects, '$[${i}].residualModifier'), 0) != 0`,
+      ),
+    ),
+    sql` OR `,
+  );
```
MathiasGruber: Two things here. (1) The 16-term OR chain is fully static — nothing in it depends on `userId` — yet it is rebuilt (16 `sql.raw` fragments + `sql.join`) on every call, and this function can be called twice per equip. Hoist it to a module-level constant. (2) At execution time it forces 16 `JSON_EXTRACT` calls per candidate row, un-indexable, and the comment concedes the real max effect count is 4. Either drop the window to a value tied to the actual data, or replace the chain with a single wildcard extraction — `JSON_EXTRACT(j_res.effects, '$[*].residualModifier')` returns all matches in one array, which you can then test in one shot instead of 16 per-row function evaluations.

app/src/server/api/routers/jutsu.ts
```diff
@@ -1125,37 +1135,67 @@ export const jutsuRouter = createTRPCRouter({
...
             userId: ctx.userId,
-            jutsuId: input.jutsuId,
-            finishTraining: new Date(Date.now() + trainTime),
-            equipped: canAutoEquip,
-          })
-          .onDuplicateKeyUpdate({ set: { id: sql`id` } });
+            userJutsuId: newUserJutsuId,
+            maxEquip,
+            flags: capFlags,
+          });
```
MathiasGruber: The new-jutsu branch of `startTraining` now issues 4 sequential DB round-trips (money deduct → insert → equip CAS → rollback CAS) where it used to issue 2. One of those is avoidable: the row was just created by this request, so instead of inserting `equipped: false` and then CAS-flipping it, insert with `equipped: canAutoEquip` and keep only the over-cap rollback statement from `tryEquipUserJutsuAtomically`.

The safety property is unchanged — the JS `canAutoEquip` pre-check already rejects the common at-cap case from the snapshot, and the rollback UPDATE is what actually handles the concurrent-equip race by unequipping this row if any cap ends up exceeded. The SQL `underCapGuards` on the equip statement only re-checks what the rollback re-checks a moment later. That would cut one round-trip off every new-jutsu training request.

app/src/server/api/routers/jutsu.ts
```diff
@@ -1341,26 +1389,97 @@ export const jutsuRouter = createTRPCRouter({
...
+                .update(userJutsu)
+                .set({ equipped: false })
+                .where(
+                  and(
+                    eq(userJutsu.id, input.userJutsuId),
+                    eq(userJutsu.userId, ctx.userId),
+                    eq(userJutsu.equipped, true),
+                  ),
+                );
+              return errorResponse("Failed to update loadout — please retry");
```
MathiasGruber: This is the one earlier review point that is still only mitigated, not closed: the compensation is a read (`isJutsuInActiveLoadout`) followed by a separate write, so it is still check-then-act. If a concurrent `selectJutsuLoadout` (or another `toggleEquip`) appends this jutsuId to the active loadout after the read returns `false`, this UPDATE still flips `equipped` back to `false` and leaves the active loadout listing a jutsu that is no longer equipped.

Fold the membership check into the compensation UPDATE itself so it is one statement, and drop `isJutsuInActiveLoadout` entirely (that also removes a round-trip from this path):

```ts
const rolledBack = await ctx.drizzle
  .update(userJutsu)
  .set({ equipped: false })
  .where(
    and(
      eq(userJutsu.id, input.userJutsuId),
      eq(userJutsu.userId, ctx.userId),
      eq(userJutsu.equipped, true),
      sql`NOT EXISTS (
        SELECT 1 FROM (
          SELECT ${jutsuLoadout.jutsuIds} AS ids
          FROM ${userData}
          INNER JOIN ${jutsuLoadout} ON ${jutsuLoadout.id} = ${userData.jutsuLoadout}
          WHERE ${userData.userId} = ${ctx.userId}
            AND ${jutsuLoadout.userId} = ${ctx.userId}
        ) AS active_loadout
        WHERE JSON_CONTAINS(active_loadout.ids, JSON_QUOTE(${userjutsuObj.jutsuId}))
      )`,
    ),
  );
if (rolledBack.rowsAffected === 1) {
  return errorResponse("Failed to update loadout — please retry");
}
```

The derived-table wrapper is needed for the same ERROR 1093 reason as the cap subqueries. With this, either the row is rolled back because the loadout genuinely does not contain it, or a concurrent writer already made the state consistent and we leave it alone — no window in between.

app/src/server/api/routers/jutsu.ts
```diff
@@ -1341,26 +1389,97 @@ export const jutsuRouter = createTRPCRouter({
...
+          userId: ctx.userId,
+          userJutsuId: input.userJutsuId,
+          maxEquip,
+          flags: getJutsuCapFlags(userjutsuObj.jutsu),
+        });
+        if (!equipped) {
+          return errorResponse(
+            "Could not equip jutsu — an equip limit was reached. Please retry.",
+          );
+        }
```
MathiasGruber: `tryEquipUserJutsuAtomically` returns `false` for three distinct reasons: the pre-check `underCapGuards` rejected it, the over-cap rollback fired, or the row simply did not match `equipped = false` (already equipped by a concurrent request / double-click). This message asserts a cap was hit, which is wrong in the last case and pushes the user to unequip something for no reason.

Either widen the wording (e.g. "Could not equip jutsu — it may already be equipped, or an equip limit was reached. Please refresh and retry.") or have `tryEquipUserJutsuAtomically` return a discriminated result (`"equipped" | "already-equipped" | "cap-reached"`) so the caller can pick the right message. This was raised in an earlier review round and is still open.

app/src/server/api/routers/jutsu.ts
```diff
@@ -2285,3 +2404,286 @@ export const selectJutsuLoadout = async (
...
+ * Materialized COUNT subquery for MySQL UPDATE guards. Wrapping in a derived table
+ * avoids ERROR 1093 (can't select from the table being updated in a subquery).
+ */
+const equippedCountSql = (userId: string) => sql`(
+  SELECT cnt FROM (
+    SELECT COUNT(*) AS cnt
+    FROM ${userJutsu} AS uj_tot
+    WHERE uj_tot.userId = ${userId} AND uj_tot.equipped = true
+  ) AS tot_equip_cnt
+)`;
```
MathiasGruber: These count helpers do not apply the same row filter as `fetchUserJutsus`, which every JS pre-check in `toggleEquip` / `startTraining` is built from.

`fetchUserJutsus` (same file, ~line 2040) filters with `ne(jutsu.jutsuType, "AI")`, but `equippedCountSql` counts *every* equipped `UserJutsu` row, and `equippedEffectTypeCountSql` / `equippedEventCountSql` / `equippedResidualCountSql` join `Jutsu` without excluding AI-type rows either.

Any user who ends up owning an equipped AI-type jutsu (a content admin flipping an existing jutsu's `jutsuType` to `AI`, or the staff grant path in `profile.ts` that inserts with `equipped: true`) then gets a permanent disagreement: the snapshot pre-check says "under cap", the CAS guard says "at cap", `tryEquipUserJutsuAtomically` returns false, and `toggleEquip` answers `Could not equip jutsu — an equip limit was reached. Please retry.` on every attempt. It is unrecoverable from the UI because the AI jutsu is not listed there to unequip.

Suggest mirroring the filter in SQL: join `Jutsu` in `equippedCountSql` and add `AND j.jutsuType != 'AI'` to all four helpers, so both sides count the same set.

app/src/server/api/routers/jutsu.ts
MathiasGruber: The fixed 16-index scan fails *open*. If a jutsu's `effects` array ever exceeds 16 entries, a `residualModifier` at index >= 16 is invisible to this count, so the guard under-counts and lets a user exceed `JUTSU_MAX_RESIDUAL_EQUIPPED` — exactly what the CAS is here to prevent. I can't find anything in the schema or the jutsu editor validation that bounds `effects.length`, so "max 4 today" is an assumption rather than an invariant, and a content admin can break it with no signal.

Cheapest fix is to fail closed instead: add `OR JSON_LENGTH(j_res.effects) > ${EQUIPPED_RESIDUAL_EFFECT_SCAN_WINDOW}` to `hasResidualModifier`, so an over-long array is conservatively counted as residual rather than silently escaping the cap. Alternatively enforce a max `effects.length` in the jutsu validator and assert it in a test, so the window stays provably sufficient.

app/src/server/api/routers/jutsu.ts
```diff
@@ -2285,3 +2404,286 @@ export const selectJutsuLoadout = async (
...
+    .set({ equipped: false })
+    .where(
+      and(
+        eq(userJutsu.id, userJutsuId),
+        eq(userJutsu.userId, userId),
+        eq(userJutsu.equipped, true),
+        or(...overCapGuards),
+      ),
+    );
+  return rollback.rowsAffected === 0;
```
MathiasGruber: The unconditional rollback makes both racers lose. Two equips from the same user landing concurrently (easy to hit by clicking two jutsu in quick succession) both pass the under-cap pre-check, both equip, and then *both* execute this rollback, both observe the over-cap condition, and both roll back. Net result: the user ends up below the cap, an equip that legitimately had room is discarded, and both requests return `Could not equip jutsu — an equip limit was reached. Please retry.`

Worth letting one racer win deterministically — e.g. only roll back when this row is not among the first N equipped rows for the affected category (ordered by `updatedAt`/`id`), or re-attempt the guarded equip once after a rollback before returning false.

app/src/server/api/routers/jutsu.ts
```diff
@@ -2285,3 +2404,286 @@ export const selectJutsuLoadout = async (
...
+    );
+  }
+  if (flags.isEvent) {
+    underCapGuards.push(
+      sql`${equippedEventCountSql(userId)} < ${JUTSU_MAX_EVENT_EQUIPPED}`,
+    );
+    overCapGuards.push(
+      sql`${equippedEventCountSql(userId)} > ${JUTSU_MAX_EVENT_EQUIPPED}`,
+    );
+  }
```
MathiasGruber: This block is six copies of the same three lines differing only by flag, count helper and constant, and it is the fifth place the same category table is spelled out. Adding `heal` in this PR required edits in `JutsuCapFlags`, `getJutsuCapFlags`, the counter/guard/increment triple in `computeJutsuLoadoutAssignments`, the derived-count + guard pair in `toggleEquip` (lines ~1316 and ~1392), the `canAutoEquip` chain in `startTraining` (~line 1151), the evolution flag check (~line 592), and here — which makes the `getJutsuCapFlags` doc comment's claim that "a new capped category only needs editing in one place" no longer true.

Suggest driving all of it from one table next to `getJutsuCapFlags`, e.g.

```ts
export const JUTSU_EQUIP_CAPS = [
  { key: "isHeal", max: JUTSU_MAX_HEAL_EQUIPPED, label: "heal", countSql: (uid: string) => equippedEffectTypeCountSql(uid, "heal") },
  // ...
] as const;
```

then build `underCapGuards`/`overCapGuards` with a single `for (const cap of JUTSU_EQUIP_CAPS) if (flags[cap.key]) { ... }`, and replace the guard chains in `toggleEquip`, `startTraining` and `computeJutsuLoadoutAssignments` with loops over the same table (`label` gives the error text for free). That collapses ~150 lines and makes the next capped category a one-line change.

app/src/libs/ranked_pvp.ts
```diff
@@ -212,6 +213,15 @@ export const validateJutsuLoadout = (jutsus: Jutsu[]) => {
...
   }
 
+  // Check heal jutsu limit
+  const healJutsus = jutsus.filter((jutsu) =>
+    jutsu.effects.some((e) => e.type === "heal"),
+  );
+  if (healJutsus.length > RANKED_LOADOUT_MAX_HEAL_JUTSUS) {
+    check = false;
+    message = `You can only equip up to ${RANKED_LOADOUT_MAX_HEAL_JUTSUS} heal jutsu in ranked PvP`;
+  }
```
MathiasGruber: `validateJutsuLoadout` re-implements category predicates that already exist in `getJutsuCapFlags` (`@/libs/jutsu`) — `e.type === "heal"` here, `e.type === "shield"` just above, and the `"residualModifier" in e && e.residualModifier` check at line ~154 is a verbatim third copy (a fourth lives in `PvpRank.tsx` ~line 761). Nothing keeps them in sync, so a future change to what counts as a heal/residual jutsu would silently apply to the equip caps but not to the ranked caps.

`getJutsuCapFlags` accepts `Pick<Jutsu, "effects" | "jutsuType">`, so the `Jutsu[]` here is already compatible. Suggest computing `const flags = getJutsuCapFlags(jutsu)` once per jutsu and filtering on `flags.isHeal` / `flags.isShield` / `flags.isBarrier` / `flags.isStun` / `flags.isResidual` / `flags.isEvent`, and extending `JutsuCapFlags` with the ranked-only categories (`poison`, `increasepoolcost`, `summon`) so each category has exactly one definition.

MathiasGruber (approved):

PR #1418, opened 2026-07-30
Fix: Delete Conversations

### LINKED ISSUE #1379: Delete button in inbox does not delete threads
**Describe the bug**
Currently, when you are in the Inbox you are unable to delete threads.  For conversations with a single user, clicking the trash bin just opens up the conversation.  

If it is a group conversation, there is currently no trash bin to delete the thread.

**To Reproduce**

1. Go to 'Inbox'
2. Click on the trash bin
3. Thread opens instead of deletes

**Expected behavior**
A confirmation modal opens to confirm you want to delete, then deletes it when you confirm.

**Further Information**

Group conversations should also have the trash bin available to delete it

milanig-creator:
# Pull Request

- Added a confirmation modal for conversation exit and fixed row-vs-trash interaction in app/src/app/inbox/page.tsx.
- Hardened modal keyboard behavior (Enter handling, shared accept path, visible-action gating) in app/src/layout/Modal2.tsx.
- Improved exit mutation UX by scoping pending state per target conversation and avoiding awaited refetch in app/src/app/inbox/page.tsx.
- Strengthened server-side exit safety in app/src/server/api/routers/comments.ts: early raid-chat guard and concurrency-safe cleanup (membership-first, guarded conversation delete, conditional comment cleanup).

## License

## Summary by CodeRabbit

* **UI Improvements**
  * Added a confirmation dialog displaying the conversation title before exiting.
  * Exit and confirmation controls are disabled while the request is processing.
  * Pending confirmation state clears when the dialog closes or the request completes.

* **Accessibility**
  * Improved keyboard support and interaction handling for confirmation dialogs.
  * Prevented keyboard actions from triggering unavailable or disabled controls.

* **Bug Fixes**
  * Improved conversation exit handling and refresh behavior after a successful exit.
  * Prevented unsupported chat types from being exited.

### DIFF
--- a/app/src/app/inbox/page.tsx
+++ b/app/src/app/inbox/page.tsx
@@ -38,6 +38,7 @@ import Confirm2 from "@/layout/Confirm2";
 import ContentBox from "@/layout/ContentBox";
 import Conversation from "@/layout/Conversation";
 import Loader from "@/layout/Loader";
+import Modal2 from "@/layout/Modal2";
 import RichInput from "@/layout/RichInput";
 import UserBlacklistControl from "@/layout/UserBlacklistControl";
 import UserSearchSelect from "@/layout/UserSearchSelect";
@@ -119,6 +120,8 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
   // Get user data & destructure
   const { data: userData } = useRequiredUserData();
   const { selectedConvo, setSelectedConvo } = props;
+  const [pendingConvoId, setPendingConvoId] = useState<string | null>(null);
+  const [isExitConfirmOpen, setIsExitConfirmOpen] = useState(false);
 
   // Fetch conversations. Note we pass the selected convo to automatically re-fetch when it changes
   const {
@@ -131,11 +134,15 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
   );
 
   // Mutations
-  const { mutate: exitConversation } = api.comments.exitConversation.useMutation({
-    onSuccess: async (data) => {
+  const {
+    mutate: exitConversation,
+    isPending: isExitingConversation,
+    variables: exitConversationVariables,
+  } = api.comments.exitConversation.useMutation({
+    onSuccess: (data) => {
       showMutationToast(data);
       if (data.success) {
-        await refetch();
+        void refetch();
       }
     },
   });
@@ -146,6 +153,11 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
     const hasNewMessages = !user?.lastReadAt || user.lastReadAt < c.updatedAt;
     return { ...c, hasNewMessages };
   });
+  const pendingConversation = filteredConversations?.find(
+    (convo) => convo.id === pendingConvoId,
+  );
+  const exitingConvoId = exitConversationVariables?.convo_id;
+  const isExitingPending = isExitingConversation && exitingConvoId === pendingConvoId;
 
   // Render
   return (
@@ -170,61 +182,106 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
             </li>
 
             <hr />
-            {filteredConversations?.map((convo) => (
-              <li
-                className={`relative mx-3 my-3 flex h-12 flex-row items-center rounded-lg hover:bg-popover ${selectedConvo && selectedConvo === convo.id ? "bg-popover" : ""}`}
-                key={convo.id}
-              >
-                <button
-                  type="button"
-                  className="absolute inset-0 h-full w-full"
-                  onClick={() => setSelectedConvo(convo.id)}
-                  aria-label={`Select conversation with ${convo.users.map((u) => u.userData.username).join(", ")}`}
-                />
-                {convo.users.length > 0 &&
-                  convo.users.map((relation, i) => {
-                    const user = relation.userData;
-                    return (
-                      <div
-                        key={user.userId}
-                        className={`absolute w-14`}
-                        style={{ left: `${i * 2}rem` }}
-                      >
-                        <AvatarImage
-                          href={user.avatar}
-                          userId={user.userId}
-                          alt={user.username}
-                          size={50}
-                          priority
-                        />
-                      </div>
-                    );
-                  })}
-                <span
-                  className="... grow truncate text-sm"
-                  style={{
-                    marginLeft: `${(convo.users.length * 2 + 1.5).toString()}rem`,
-                  }}
+            {filteredConversations?.map((convo) => {
+              const isExiting = isExitingConversation && exitingConvoId === convo.id;
+              return (
+                <li
+                  className={`relative mx-3 my-3 flex h-12 flex-row items-center rounded-lg hover:bg-popover ${selectedConvo && selectedConvo === convo.id ? "bg-popover" : ""}`}
+                  key={convo.id}
                 >
-                  {convo.title}
-                  <br />
-                  {convo.createdAt.toDateString()}
-                </span>
-                <div className="grow"></div>
-                {convo.hasNewMessages && (
-                  <BellRing className="h-6 w-6 animate-[wiggle_1s_ease-in-out_infinite] text-red-500 hover:cursor-pointer hover:text-orange-500" />
-                )}
-                <Trash2
-                  className="mx-2 h-6 w-6 rounded-full hover:cursor-pointer hover:text-orange-500"
-                  onClick={(e) => {
-                    e.preventDefault();
-                    e.stopPropagation();
-                    exitConversation({ convo_id: convo.id });
-                  }}
-                />
-              </li>
-            ))}
+                  <button
+                    type="button"
+                    className="absolute inset-0 h-full w-full"
+                    onClick={() => setSelectedConvo(convo.id)}
+                    aria-label={`Select conversation with ${convo.users.map((u) => u.userData.username).join(", ")}`}
+                  />
+                  {convo.users.length > 0 &&
+                    convo.users.map((relation, i) => {
+                      const user = relation.userData;
+                      return (
+                        <div
+                          key={user.userId}
+                          className={`absolute w-14`}
+                          style={{ left: `${i * 2}rem` }}
+                        >
+                          <AvatarImage
+                            href={user.avatar}
+                            userId={user.userId}
+                            alt={user.username}
+                            size={50}
+                            priority
+                          />
+                        </div>
+                      );
+                    })}
+                  <span
+                    className="... grow truncate text-sm"
+                    style={{
+                      marginLeft: `${(convo.users.length * 2 + 1.5).toString()}rem`,
+                    }}
+                  >
+                    {convo.title}
+                    <br />
+                    {convo.createdAt.toDateString()}
+                  </span>
+                  <div className="grow"></div>
+                  {convo.hasNewMessages && (
+                    <BellRing className="h-6 w-6 animate-[wiggle_1s_ease-in-out_infinite] text-red-500 hover:cursor-pointer hover:text-orange-500" />
+                  )}
+                  <Button
+                    type="button"
+                    variant="ghost"
+                    size="icon"
+                    className="relative z-10 mx-2 rounded-full"
+                    aria-label={`Exit conversation ${convo.title}`}
+                    disabled={isExiting}
+                    onClick={(e) => {
+                      e.preventDefault();
+                      e.stopPropagation();
+                      setPendingConvoId(convo.id);
+                      setIsExitConfirmOpen(true);
+                    }}
+                  >
+                    <Trash2 className="h-6 w-6 hover:text-orange-500" />
+                  </Button>
+                </li>
+              );
+            })}
           </ul>
+          <Modal2
+            title="Confirm exiting conversation"
+            isOpen={isExitConfirmOpen}
+            setIsOpen={setIsExitConfirmOpen}
+            isLoading={isExitingPending}
+            proceed_loading_label="Exiting..."
+            proceed_label="Proceed"
+            proceedDisabled={!pendingConvoId || isExitingPending}
+            onAccept={(e) => {
+              e.preventDefault();
+              e.stopPropagation();
+              if (!pendingConvoId) return;
+              const requestedConvoId = pendingConvoId;
+              exitConversation(
+                { convo_id: requestedConvoId },
+                {
+                  onSettled: () => {
+                    setPendingConvoId((currentConvoId) =>
+                      currentConvoId === requestedConvoId ? null : currentConvoId,
+                    );
+                  },
+                },
+              );
+            }}
+            onClose={() => {
+              if (!isExitingPending) {
+                setPendingConvoId(null);
+              }
+            }}
+          >
+            {pendingConversation
+              ? `You are about to exit "${pendingConversation.title}". Are you sure?`
+              : "You are about to exit this conversation. Are you sure?"}
+          </Modal2>
           <div className="m-3 italic">- Messages deleted after 14 days</div>
         </div>
       )}
--- a/app/src/app/inbox/page.tsx
+++ b/app/src/app/inbox/page.tsx
@@ -42,6 +42,7 @@ import Modal2 from "@/layout/Modal2";
 import RichInput from "@/layout/RichInput";
 import UserBlacklistControl from "@/layout/UserBlacklistControl";
 import UserSearchSelect from "@/layout/UserSearchSelect";
+import { isRaidChatConversationId } from "@/libs/raids";
 import { showMutationToast } from "@/libs/toast";
 import { canPostAsAi, getMessagingRestriction } from "@/utils/permissions";
 import { useRequiredUserData } from "@/utils/UserContext";
@@ -184,6 +185,9 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
             <hr />
             {filteredConversations?.map((convo) => {
               const isExiting = isExitingConversation && exitingConvoId === convo.id;
+              // Raid chat membership follows the raid queue, so the server always
+              // rejects a manual exit; don't offer the control for those rows.
+              const canExit = !isRaidChatConversationId(convo.id);
               return (
                 <li
                   className={`relative mx-3 my-3 flex h-12 flex-row items-center rounded-lg hover:bg-popover ${selectedConvo && selectedConvo === convo.id ? "bg-popover" : ""}`}
@@ -228,22 +232,24 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
                   {convo.hasNewMessages && (
                     <BellRing className="h-6 w-6 animate-[wiggle_1s_ease-in-out_infinite] text-red-500 hover:cursor-pointer hover:text-orange-500" />
                   )}
-                  <Button
-                    type="button"
-                    variant="ghost"
-                    size="icon"
-                    className="relative z-10 mx-2 rounded-full"
-                    aria-label={`Exit conversation ${convo.title}`}
-                    disabled={isExiting}
-                    onClick={(e) => {
-                      e.preventDefault();
-                      e.stopPropagation();
-                      setPendingConvoId(convo.id);
-                      setIsExitConfirmOpen(true);
-                    }}
-                  >
-                    <Trash2 className="h-6 w-6 hover:text-orange-500" />
-                  </Button>
+                  {canExit && (
+                    <Button
+                      type="button"
+                      variant="ghost"
+                      size="icon"
+                      className="relative z-10 mx-2 rounded-full"
+                      aria-label={`Exit conversation ${convo.title}`}
+                      disabled={isExiting}
+                      onClick={(e) => {
+                        e.preventDefault();
+                        e.stopPropagation();
+                        setPendingConvoId(convo.id);
+                        setIsExitConfirmOpen(true);
+                      }}
+                    >
+                      <Trash2 className="h-6 w-6 hover:text-orange-500" />
+                    </Button>
+                  )}
                 </li>
               );
             })}
@@ -252,6 +258,10 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
             title="Confirm exiting conversation"
             isOpen={isExitConfirmOpen}
             setIsOpen={setIsExitConfirmOpen}
+            // Keep Modal2 from self-closing on accept, so this destructive action
+            // stays visible in its disabled "Exiting..." state until the request
+            // settles and onSettled closes the dialog.
+            isValid={false}
             isLoading={isExitingPending}
             proceed_loading_label="Exiting..."
             proceed_label="Proceed"
@@ -268,6 +278,7 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
                     setPendingConvoId((currentConvoId) =>
                       currentConvoId === requestedConvoId ? null : currentConvoId,
                     );
+                    setIsExitConfirmOpen(false);
                   },
                 },
               );
--- a/app/src/layout/Modal2.tsx
+++ b/app/src/layout/Modal2.tsx
@@ -3,7 +3,7 @@
  * This is a modal that is used to display a modal.
  */
 import type React from "react";
-import { useEffect } from "react";
+import { useCallback, useEffect } from "react";
 import { Button } from "@/components/ui/button";
 import {
   Dialog,
@@ -48,23 +48,44 @@ const Modal2: React.FC<Modal2Props> = (props) => {
 
   const isProceedDisabled = props.isLoading || props.proceedDisabled;
 
+  const handleAccept = useCallback(
+    (
+      e:
+        | React.MouseEvent<HTMLButtonElement, MouseEvent>
+        | React.KeyboardEvent<KeyboardEvent>,
+    ) => {
+      if (!props.onAccept) return;
+      props.onAccept(e);
+      if (props.isValid === undefined || props.isValid) {
+        props.setIsOpen(false);
+      }
+    },
+    [props.onAccept, props.isValid, props.setIsOpen],
+  );
+
   // Handle key-presses for Enter key only when this modal is open
   useEffect(() => {
+    if (!props.isOpen) return;
+
     const onDocumentKeyDown = (event: KeyboardEvent) => {
-      if (!props.isOpen) return;
       // Don't trigger if the active element is a button (it will handle Enter itself)
       const activeElement = document.activeElement;
       const isButton = activeElement?.tagName === "BUTTON";
 
-      if (event.key === "Enter" && props.onAccept && !isButton && !isProceedDisabled) {
-        props.onAccept(event as unknown as React.KeyboardEvent<KeyboardEvent>);
+      if (
+        event.key === "Enter" &&
+        props.proceed_label &&
+        !isButton &&
+        !isProceedDisabled
+      ) {
+        handleAccept(event as unknown as React.KeyboardEvent<KeyboardEvent>);
       }
     };
     document.addEventListener("keydown", onDocumentKeyDown);
     return () => {
       document.removeEventListener("keydown", onDocumentKeyDown);
     };
-  }, [props.isOpen, props.onAccept, isProceedDisabled]);
+  }, [props.isOpen, props.proceed_label, isProceedDisabled, handleAccept]);
 
   const handleDialogClose = () => {
     if (props.onClose) props.onClose();
@@ -112,10 +133,7 @@ const Modal2: React.FC<Modal2Props> = (props) => {
                 onClick={(e) => {
                   e.preventDefault();
                   e.stopPropagation();
-                  if (props.onAccept) props.onAccept(e);
-                  if (props.isValid === undefined || props.isValid) {
-                    props.setIsOpen(false);
-                  }
+                  handleAccept(e);
                 }}
                 className={`z-30 rounded-lg ${confirmBtnClassName}`}
               >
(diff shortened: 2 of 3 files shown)


app/src/app/inbox/page.tsx
```diff
@@ -214,14 +214,18 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
                 {convo.hasNewMessages && (
                   <BellRing className="h-6 w-6 animate-[wiggle_1s_ease-in-out_infinite] text-red-500 hover:cursor-pointer hover:text-orange-500" />
                 )}
-                <Trash2
-                  className="mx-2 h-6 w-6 rounded-full hover:cursor-pointer hover:text-orange-500"
+                <button
```
MathiasGruber: Now that this button is actually clickable (it was previously dead because the `absolute inset-0` row-select button on line 178 covered it), it fires an irreversible destructive mutation on a single tap, with no confirmation and no pending guard.

`comments.exitConversation` (`app/src/server/api/routers/comments.ts:406-449`) deletes the `user2conversation` row unconditionally, and when the caller is the last remaining member it also hard-deletes the `conversation` row and every `conversationComment`. So a mis-tap on this 24px icon — which sits directly on top of a full-row click target, on mobile in particular — permanently destroys the message history with no undo. That risk did not exist before this PR because the icon was inert.

Two concrete problems:

1. **No confirmation.** The established pattern in this codebase for destructive icon actions inside a list is to wrap the icon in `Confirm2` (see `app/src/app/forum/[boardid]/page.tsx:319-345` for pin/lock/delete). `Confirm2` is already imported in this file.
2. **No pending guard.** `exitConversation` is destructured without `isPending`, and the row stays mounted until `refetch()` resolves, so a double-click fires the mutation twice. The second call fails and surfaces a red toast to the user even though the delete succeeded — `"Conversation not found"` for a 1:1 convo that was hard-deleted, or `"You are not allowed to view this conversation"` for a group convo where the membership row is already gone (`canViewConversation` no longer matches). The forum code pairs `Confirm2` with `disabled={isDeletingThread}` for exactly this reason.

Suggest wrapping in `Confirm2` with a `disabled={isPending}` prop and keeping the `preventDefault`/`stopPropagation` on the accept handler.

milanig-creator: I wrapped the exit control in `Confirm2`, added an `isPending` guard that disables the control while the mutation is running, and preserved `preventDefault()`/ `stopPropagation()` in the accept handler. I also switched the trigger to the shared `Button` component. TypeScript, lint, and tests pass.

app/src/app/inbox/page.tsx
```diff
@@ -214,14 +215,29 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
                 {convo.hasNewMessages && (
                   <BellRing className="h-6 w-6 animate-[wiggle_1s_ease-in-out_infinite] text-red-500 hover:cursor-pointer hover:text-orange-500" />
                 )}
-                <Trash2
-                  className="mx-2 h-6 w-6 rounded-full hover:cursor-pointer hover:text-orange-500"
-                  onClick={(e) => {
+                <Confirm2
```
MathiasGruber: Minor perf note: this mounts one `Confirm2` -> `Modal2` per conversation row. A closed Radix dialog renders no DOM, but `Modal2` registers a **document-level `keydown` listener unconditionally** (`app/src/layout/Modal2.tsx`, the `useEffect` with deps `[props.isOpen, props.onAccept, isProceedDisabled]`) and only checks `isOpen` inside the handler. So an inbox with N conversations now keeps N global listeners alive for dialogs that are all closed.

It is also re-bound more often than it needs to be: `onAccept` is an inline closure over `convo.id`, so every render of the list tears down and re-adds all N listeners. `isExitingConversation` makes that happen on each mutation start/settle (and it disables the trash icon on *every* row, not just the one being deleted).

If you want to keep it cheap, a single `Confirm2` at the list level driven by a `pendingConvoId` state gives the same UX with one dialog instance and one listener, and lets the pending flag apply only to the affected row.

app/src/app/inbox/page.tsx
```diff
@@ -214,14 +215,29 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
...
+                      disabled={isExitingConversation}
+                    >
+                      <Trash2 className="h-6 w-6 hover:text-orange-500" />
+                    </Button>
+                  }
+                  disabled={isExitingConversation}
+                  onAccept={(e) => {
                     e.preventDefault();
                     e.stopPropagation();
                     exitConversation({ convo_id: convo.id });
```
MathiasGruber: This click path is effectively new (the old `Trash2` had no `z-10`, so the `absolute inset-0` select-overlay swallowed the click), which makes `comments.exitConversation` newly reachable from the inbox. Worth tightening its writes while it is in scope: in `app/src/server/api/routers/comments.ts` (~lines 434-447) the mutation awaits three deletes sequentially -- `user2conversation`, then, when the caller is the last member, `conversation` and `conversationComment`.

The last two are independent of each other and of the first (the `convo.users.length === 1` branch is decided from the already-fetched `convo`), so they can be batched:

```ts
await Promise.all([
  ctx.drizzle.delete(user2conversation).where(and(...)),
  ...(convo.users.length === 1
    ? [
        ctx.drizzle.delete(conversation).where(eq(conversation.id, convo.id)),
        ctx.drizzle.delete(conversationComment).where(eq(conversationComment.conversationId, convo.id)),
      ]
    : []),
]);
```

That takes the last-member case from three round-trips to one, in line with the parallel-mutation rule in CLAUDE.md. (`getUserConversations` already parallelises its read and the `inboxNews` reset, so nothing to change there.)

app/src/app/inbox/page.tsx
```diff
@@ -214,14 +215,29 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
...
+                <Confirm2
+                  title="Confirm exiting conversation"
+                  button={
+                    <Button
+                      type="button"
+                      variant="ghost"
+                      size="icon"
+                      className="relative z-10 mx-2 rounded-full"
+                      aria-label={`Exit conversation ${convo.title}`}
+                      disabled={isExitingConversation}
```
MathiasGruber: `isExitingConversation` comes from the single mutation hook in `ShowConversations`, so it is shared by every row in the map: leaving one conversation disables the trash icon on *all* conversations. It also stays disabled longer than you might expect, since react-query keeps the mutation `pending` until `onSuccess` resolves and `onSuccess` awaits `refetch()`. If the intent is only to block a double submit of the same row, scope it with the mutation variables:

```ts
const { mutate: exitConversation, isPending, variables } = api.comments.exitConversation.useMutation({...});
// ...
const isExiting = isPending && variables?.convo_id === convo.id;
```

app/src/app/inbox/page.tsx
```diff
@@ -131,7 +134,11 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
   );
 
   // Mutations
-  const { mutate: exitConversation } = api.comments.exitConversation.useMutation({
+  const {
+    mutate: exitConversation,
+    isPending: isExitingConversation,
```
MathiasGruber, line 140: Wiring the UI to this mutation's `isPending` makes the existing `await refetch()` in `onSuccess` (line 145) a lot more expensive than it looks.

TanStack Query v5 awaits `onSuccess`/`onSettled` before dispatching `success`, so `isExitingConversation` stays `true` for the exit RTT **plus** a full `getUserConversations` round-trip. That query is not cheap: it joins `userData` -> `creatorBlacklist` + every `user2conversation` -> `conversation` -> `users` -> `userData`, and it also issues a write (`update userData set inboxNews = 0`) on every call. So each delete now costs a heavy read + a write, and the row's trash `Button` (`disabled={isExiting}`) and the modal's `isLoading`/`proceedDisabled` are held across that whole window rather than just the mutation.

Two options, either fixes it:
- Drop the row from the cache instead of refetching: `utils.comments.getUserConversations.setData({ selectedConvo }, (old) => old?.filter((c) => c.id === variables.convo_id ? false : true))` — zero extra round-trips, and the list updates the instant the mutation returns.
- If you want to keep the server as source of truth, at least don't `await` it (`void refetch()`), so the pending state ends with the mutation and doesn't span the refetch.

app/src/server/api/routers/comments.ts
```diff
@@ -431,20 +431,24 @@ export const commentsRouter = createTRPCRouter({
         return errorResponse("Raid chat membership is managed by the raid queue");
```
MathiasGruber: This guard is a pure string prefix check (`isRaidChatConversationId`) but it runs after the `Promise.all([fetchUser, fetchConversation])` above (lines ~410-417). `fetchConversation` is called with `id: input.convo_id` and filters on `eq(conversation.id, id)`, so `convo.id === input.convo_id` unconditionally — the check can be made on the raw input.

Raid chat conversations are not filtered out of `getUserConversations`, so they do render in the inbox list with a trash button, which means this path is reachable from the normal UI. Hoisting `if (isRaidChatConversationId(input.convo_id)) return errorResponse(...)` above the `Promise.all` short-circuits two DB reads (a full `userData` row fetch and a `conversation` + `users` join) for every such attempt.

The rest of the procedure is fine — guards read in parallel and the deletes are correctly collapsed into one `Promise.all`.

app/src/layout/Modal2.tsx
```diff
@@ -50,14 +50,19 @@ const Modal2: React.FC<Modal2Props> = (props) => {
...
-      if (!props.isOpen) return;
       // Don't trigger if the active element is a button (it will handle Enter itself)
       const activeElement = document.activeElement;
       const isButton = activeElement?.tagName === "BUTTON";
 
-      if (event.key === "Enter" && props.onAccept && !isButton && !isProceedDisabled) {
+      if (event.key === "Enter" && !isButton && !isProceedDisabled) {
+        if (!props.onAccept) return;
         props.onAccept(event as unknown as React.KeyboardEvent<KeyboardEvent>);
+        if (props.isValid === undefined || props.isValid) {
```
MathiasGruber: `props.isValid` is now read inside the effect, but it isn't in the dependency array (`[props.isOpen, props.onAccept, isProceedDisabled]`). The listener therefore decides whether to close using whatever `isValid` was when it was last registered, while the pointer path at line 121 reads it fresh on every click — so the two paths can disagree.

This matters because `reactCompiler: true` is on (app/next.config.mjs:19). Where `onAccept` is a stable reference and no `isLoading`/`proceedDisabled` is passed, nothing in the dep array changes when `isValid` flips, so the effect never re-registers. Concrete: `app/src/layout/Clan.tsx:972` passes `isValid={editForm.formState.isValid}` with `onAccept={onEdit}` and no loading props. Dialog opens with `isValid === false`; the user fills the form and validity flips to `true`; pressing Enter from inside a text input runs `onEdit` but leaves the dialog open (stale `false`), whereas clicking Proceed submits *and* closes. Same shape at `app/src/layout/PublicUser.tsx:1495`, `app/src/app/anbu/[anbuid]/page.tsx:338`, and `app/src/layout/Report.tsx:93`. The mirror case is worse: `isValid` true at registration then flipping false means Enter closes on an invalid submit.

The close decision is now duplicated verbatim in two places (here and lines 121-123), which is what let them drift. Suggest extracting one handler and calling it from both:

```ts
const handleAccept = (e) => {
  if (!props.onAccept) return;
  props.onAccept(e);
  if (props.isValid === undefined || props.isValid) props.setIsOpen(false);
};
```

Call it from the Button's `onClick` and from the keydown listener; keep it in a ref (or add `props.isValid` to the deps) so the listener always sees current props.

app/src/layout/Modal2.tsx
```diff
@@ -50,14 +50,19 @@ const Modal2: React.FC<Modal2Props> = (props) => {
...
+    if (!props.isOpen) return;
+
     const onDocumentKeyDown = (event: KeyboardEvent) => {
-      if (!props.isOpen) return;
       // Don't trigger if the active element is a button (it will handle Enter itself)
       const activeElement = document.activeElement;
       const isButton = activeElement?.tagName === "BUTTON";
 
-      if (event.key === "Enter" && props.onAccept && !isButton && !isProceedDisabled) {
+      if (event.key === "Enter" && !isButton && !isProceedDisabled) {
```
MathiasGruber: The Enter path isn't gated on `props.proceed_label`, but the Proceed button it's meant to mirror is only rendered when `proceed_label` is truthy (line 112). Now that Enter also closes the dialog, the keyboard can do something the visible UI cannot.

Concrete regression: `app/src/layout/OccupationCrafting.tsx:669` sets `proceed_label={repairAllMutation.isPending ? undefined : "Repair All"}` with `onAccept={() => repairAllMutation.mutate()}`, and passes neither `isLoading` nor `confirmDisabled`, so `isProceedDisabled` is falsy. While the repair is in flight the button is deliberately gone, but the document listener is still live — Enter re-fires the mutation (pre-existing) and now *also* closes the dialog out from under the in-flight repair. `app/src/layout/CraftingCatalog.tsx:311` has the identical shape with `handleCraft`, and there `setIsOpen` is a custom setter that unwinds the whole drill-down back to the category grid. `app/src/layout/WarSystem.tsx:599` passes a no-op `onAccept` on its informational branches where `proceedLabel` is `undefined`, so those info-only dialogs now dismiss on Enter.

Adding `&& props.proceed_label` to the condition would keep the keyboard path in step with what's actually clickable.

app/src/app/inbox/page.tsx
```diff
@@ -170,61 +181,106 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
...
+                  </Button>
+                </li>
+              );
+            })}
           </ul>
+          <Modal2
+            title="Confirm exiting conversation"
+            isOpen={isExitConfirmOpen}
+            setIsOpen={setIsExitConfirmOpen}
+            isLoading={isExitingConversation}
```
MathiasGruber: `isLoading` and the `isExitingConversation` half of `proceedDisabled` come from the shared mutation hook's global pending flag, not from the conversation this dialog is about, and they end up applying to the wrong row.

Because no `isValid` is passed, Modal2's Proceed handler calls `setIsOpen(false)` immediately after `onAccept` (Modal2.tsx:120-123), so the dialog closes synchronously while the mutation is still in flight. The result is inverted from the intent:

- For the conversation actually being exited, the dialog is already gone — `proceed_loading_label="Exiting..."` and the disabled Proceed button are never seen.
- For a *different* conversation, they are: exit A (dialog closes, request still in flight), then click the trash on B. B's dialog opens with Proceed disabled and labelled "Exiting..." even though nothing about B is pending, and Escape/Close won't clear `pendingConvoId` either because the `onClose` guard at line 275 also tests the global `isExitingConversation`. The user has to wait out A's request before they can act on B.

Scoping the flag to the row fixes both halves:

```ts
const isExitingPending = isExitingConversation && exitingConvoId === pendingConvoId;
```

and use that for `isLoading`, `proceedDisabled`, and the `onClose` guard. If the dialog is meant to stay open with the "Exiting..." state while the request runs, it also needs `isValid={false}` plus an explicit `setIsExitConfirmOpen(false)` from the mutation's `onSuccess`.

app/src/server/api/routers/comments.ts
```diff
@@ -431,20 +431,24 @@ export const commentsRouter = createTRPCRouter({
...
+            ),
           ),
-        );
-      if (convo.users.length === 1) {
-        await ctx.drizzle.delete(conversation).where(eq(conversation.id, convo.id));
-        await ctx.drizzle
-          .delete(conversationComment)
-          .where(eq(conversationComment.conversationId, convo.id));
-      }
+        ...(convo.users.length === 1
```
MathiasGruber: The parallelisation itself is fine (three independent tables, no FK constraints declared in schema.ts, and `getUserConversations` already null-guards `c.conversation` at line 361 so a partial failure can't break the inbox). The condition it guards is the problem, and it's worth tightening while this block is being touched.

`convo.users.length === 1` is read as "the caller is the last member", but nothing here checks that the caller is a member at all. `canViewConversation` returns true for *any* `isPublic` conversation regardless of membership (utils/permissions.ts:1007), and `Conversation.isPublic` defaults to `true` in the schema. So a non-member can call `exitConversation` on a public conversation: their own `user2conversation` delete is a no-op, and if that conversation happens to have exactly one membership row, the branch below hard-deletes the conversation row and every `conversationComment` belonging to someone else. This procedure is also `mcp: { enabled: true }`, so it isn't gated by the inbox UI only listing your own conversations.

That's the same failure mode the raid-chat guard above was added to prevent, just reachable by a different route. Suggest:

```ts
const isLastMember =
  convo.users.length === 1 && convo.users[0]?.userId === ctx.userId;
```

app/src/app/inbox/page.tsx
milanig-creator: Updated `app/src/app/inbox/page.tsx`: removed awaited refetch in mutation success (`void refetch()`), so pending state ends with the mutation instead of mutation+refetch.

app/src/app/inbox/page.tsx
milanig-creator: Implemented in `app/src/app/inbox/page.tsx`: scoped pending state to the active confirmation (`isExitingPending = isExitingConversation && exitingConvoId === pendingConvoId`) and now use it for `isLoading`, `proceedDisabled`, and the `onClose` guard, so pending UI no longer leaks across rows. Kept current close-on-confirm behavior as intended for this flow.

app/src/layout/Modal2.tsx
milanig-creator: Updated `app/src/layout/Modal2.tsx: extracted a shared handleAccept` used by both click and Enter paths, and wired effect dependencies through that callback (including current `isValid`), eliminating stale validity reads and keeping close behavior consistent across input methods.

app/src/server/api/routers/comments.ts
milanig-creator: Updated `app/src/server/api/routers/comments.ts`: hoisted the raid-chat guard to the top of `exitConversation` (`isRaidChatConversationId(input.convo_id)`) before `Promise.all`, so unsupported exits short-circuit before user and conversation DB reads.

app/src/server/api/routers/comments.ts
milanig-creator: Updated `app/src/server/api/routers/comments.ts`: tightened the sole-member cleanup condition to `isLastMember = convo.users.length === 1 && convo.users[0]?.userId === ctx.userId`, so hard delete of conversation/comments only occurs when the caller is actually the final member.

app/src/app/inbox/page.tsx
```diff
@@ -119,6 +120,8 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
   // Get user data & destructure
   const { data: userData } = useRequiredUserData();
   const { selectedConvo, setSelectedConvo } = props;
+  const [pendingConvoId, setPendingConvoId] = useState<string | null>(null);
```
MathiasGruber, line 124: This re-implements `Confirm2`, which is already imported in this file and used for exactly this pattern at line 415. Because `Confirm2` owns its own open state per instance, rendering one inside the `.map()` next to each row removes the need for `pendingConvoId`, `isExitConfirmOpen`, `pendingConversation`, `exitConversationVariables`, `exitingConvoId`, `isExitingPending`, the `onSettled` reset, and the `onClose` guard — roughly 45 lines of state machinery replaced by:

```tsx
<Confirm2
  title="Confirm exiting conversation"
  proceed_label="Proceed"
  button={<Trash2 className="mx-2 h-6 w-6 rounded-full hover:cursor-pointer hover:text-orange-500" />}
  onAccept={() => exitConversation({ convo_id: convo.id })}
>
  {`You are about to exit "${convo.title}". Are you sure?`}
</Confirm2>
```

That also drops the new `Modal2` import. (Anchored here since the finding covers the whole added state block rather than one line.)

app/src/layout/Modal2.tsx
```diff
@@ -48,23 +48,44 @@ const Modal2: React.FC<Modal2Props> = (props) => {
 
   const isProceedDisabled = props.isLoading || props.proceedDisabled;
 
+  const handleAccept = useCallback(
+    (
+      e:
+        | React.MouseEvent<HTMLButtonElement, MouseEvent>
+        | React.KeyboardEvent<KeyboardEvent>,
+    ) => {
+      if (!props.onAccept) return;
```
MathiasGruber: `handleAccept` bails out entirely when `onAccept` is undefined, which changes behaviour for every Modal2 that renders a Proceed button without an accept handler. On `main` the click handler was:

```ts
if (props.onAccept) props.onAccept(e);
if (props.isValid === undefined || props.isValid) {
  props.setIsOpen(false);
}
```

so a Proceed button with no `onAccept` still closed the dialog. Now it does nothing at all — the button is rendered, enabled, and inert.

Concrete call sites that regress:

- `app/src/layout/PublicUser.tsx:462` — `<Modal2 title="Update User Data" proceed_label="Done">` with no `onAccept`. The "Done" button no longer closes the edit dialog.
- `app/src/layout/Clan.tsx:1037` — `<Confirm2 title={`Village ${groupLabel} Overview`}>` with no `onAccept`. `Confirm2.tsx:60-62` defaults `proceed_label` to `"Proceed"`, so a Proceed button is rendered and is now dead.
- `app/src/layout/ItemWithEffects.tsx:302` — `<Confirm2 title="3d Model">`, same shape, same dead Proceed button.

Keep the close side-effect independent of whether `onAccept` exists:

```ts
const handleAccept = useCallback((e) => {
  props.onAccept?.(e);
  if (props.isValid === undefined || props.isValid) {
    props.setIsOpen(false);
  }
}, [props.onAccept, props.isValid, props.setIsOpen]);
```

The Enter path is already gated on `proceed_label`, so this does not reintroduce the keyboard-closes-an-info-dialog problem.

app/src/app/inbox/page.tsx
```diff
@@ -170,61 +182,106 @@ const ShowConversations: React.FC<ShowConversationsProps> = (props) => {
...
+                  </Button>
+                </li>
+              );
+            })}
           </ul>
+          <Modal2
+            title="Confirm exiting conversation"
+            isOpen={isExitConfirmOpen}
+            setIsOpen={setIsExitConfirmOpen}
+            isLoading={isExitingPending}
```
MathiasGruber, line 265: `isLoading={isExitingPending}`, `proceed_loading_label="Exiting..."`, the `isExitingPending` half of `proceedDisabled` on line 258, and the `!isExitingPending` guard in `onClose` on line 276 are all unreachable.

No `isValid` is passed, so `Modal2.handleAccept` (`app/src/layout/Modal2.tsx:59-61`) hits `props.isValid === undefined` and calls `setIsOpen(false)` synchronously right after `onAccept` returns. The dialog is therefore always closed by the time `isExitingConversation` flips to `true`. And `isExitingPending` requires `exitingConvoId === pendingConvoId`: while a request for A is in flight the trash button for A is `disabled={isExiting}` so the dialog cannot be reopened for A, and opening it for B sets `pendingConvoId = B`, which makes `isExitingPending` false. There is no state in which this dialog is open and `isExitingPending` is true.

Net effect: a destructive action fires with zero pending feedback — this is exactly what the preview review run reported as a Fail ("clicking `Proceed` dismissed the confirmation dialog immediately instead of keeping it visible in a disabled `Exiting...` state"), and it is the half of the earlier review thread that was answered with "kept current close-on-confirm behavior as intended" while the loading props were left in place.

Pick one:
- Keep close-on-confirm and delete `isLoading`, `proceed_loading_label`, the `isExitingPending` term in `proceedDisabled`, and the `onClose` guard, since none of them can ever apply.
- Or actually deliver the intended UX: pass `isValid={false}` so Modal2 does not self-close, and call `setIsExitConfirmOpen(false)` from the mutation's `onSettled`/`onSuccess`.

app/src/server/api/routers/comments.ts
```diff
@@ -421,29 +424,31 @@ export const commentsRouter = createTRPCRouter({
...
+        const conversationDeleteResult = await ctx.drizzle
+          .delete(conversation)
+          .where(
+            and(
+              eq(conversation.id, convo.id),
+              sql`NOT EXISTS (SELECT 1 FROM ${user2conversation} WHERE ${user2conversation.conversationId} = ${convo.id})`,
+            ),
+          );
+
+        if (conversationDeleteResult.rowsAffected === 1) {
```
MathiasGruber: Chaining the comment delete behind the conversation delete makes a last-member exit three sequential DB round-trips (membership -> conversation -> comments). An earlier revision of this PR had the cleanup deletes collapsed into one `Promise.all`; the CAS rework serialized them again, and it doesn't have to.

The dependency is artificial: both cleanup deletes only need to know that no membership rows remain, which is already established once the awaited membership delete has landed. Give the comment delete the same `NOT EXISTS` predicate and fire both concurrently:

```ts
if (membershipDeleteResult.rowsAffected === 1) {
  const noMembersLeft = sql`NOT EXISTS (SELECT 1 FROM ${user2conversation} WHERE ${user2conversation.conversationId} = ${convo.id})`;
  await Promise.all([
    ctx.drizzle
      .delete(conversation)
      .where(and(eq(conversation.id, convo.id), noMembersLeft)),
    ctx.drizzle
      .delete(conversationComment)
      .where(and(eq(conversationComment.conversationId, convo.id), noMembersLeft)),
  ]);
}
```

The concurrency fix is preserved — while any member remains both guards fail, and the last exiter satisfies both. It is never slower than the current version either: the comment delete is index-backed (`ConversationComment_conversationId_idx`, and the subquery hits the `user2conversation` PK), and it runs concurrently rather than adding a round-trip. Per CLAUDE.md the thing to minimize is round-trip count, not query count.

app/src/app/inbox/page.tsx
MathiasGruber: Addressed in 5558828, taking the second option: the dialog now stays open in its disabled `Exiting...` state until the request settles.

`isValid={false}` stops `Modal2.handleAccept` from self-closing, and the mutation's `onSettled` closes it (alongside the existing id-guarded `pendingConvoId` reset). `isExitingPending` is now reachable, so `isLoading`, `proceed_loading_label`, and the `isExitingPending` half of `proceedDisabled` all do real work — the latter also gates the Enter path, so the confirm cannot be double-submitted while in flight. On error, `onSettled` still closes and the error toast surfaces.

This should clear the Fail the preview run reported for this flow.

app/src/app/inbox/page.tsx
MathiasGruber: Not taking this one — the swap isn't a drop-in, and the sibling thread on the loading props pushes the other way. Three concrete blockers:

1. **It would reintroduce the bug this PR fixes.** `Confirm2` wraps its trigger in `<div role="button" className="inline-block …">`, which is statically positioned. Under the row's `absolute inset-0` overlay button, a statically-positioned trigger paints *below* the positioned overlay, so the trash icon goes dead again — exactly the original defect. `Confirm2Props.className` forwards to `Modal2`, not to that wrapper, so there's no way to give it `relative z-10` without first adding a trigger-className passthrough to `Confirm2`.

2. **`Confirm2` can't express the pending state.** `Confirm2Props` has no `isLoading` / `proceed_loading_label`, and it doesn't expose `isOpen`/`setIsOpen`. Resolving the other thread the "deliver the intended UX" way requires holding the dialog open in an `Exiting...` state across the request — which needs exactly the `pendingConvoId` / `isExitingPending` machinery this thread proposes deleting. The two threads are in tension; I went with the UX, so the state earns its keep.

3. **Per-row mounting was already flagged.** An earlier round on this PR raised one `Confirm2` → `Modal2` per conversation row as a perf note. Going back to per-row would re-open it.

Happy to revisit if `Confirm2` grows a trigger-className passthrough plus loading props, but that's a wider refactor than this PR.

MathiasGruber (approved):

app/src/layout/Modal2.tsx
MathiasGruber: Fixed in e21920f — `handleAccept` now runs the close side-effect independently of `onAccept`:

```ts
props.onAccept?.(e);
if (props.isValid === undefined || props.isValid) {
  props.setIsOpen(false);
}
```

All three call sites you listed are working again. I verified the `PublicUser.tsx:462` one in the running app: opening "Update User Data" and clicking **Done** now closes the dialog (`dialogStillOpenAfterDone: false`), where before it was inert.

One deliberate deviation from the suggested patch: I also gated the Enter path on `props.onAccept`, not just `proceed_label`. With `proceed_label` alone, Enter would have *newly* closed "Update User Data" from inside a text field — behaviour `main` never had, since the old condition included `props.onAccept`. Enter now requires a Proceed button that is visible, enabled, **and** backed by a handler, which keeps `main`'s keyboard semantics exactly while still fixing the click path.

app/src/server/api/routers/comments.ts
MathiasGruber: Fixed in e21920f — both cleanup deletes now share the `NOT EXISTS` predicate and run concurrently, exactly as suggested:

```ts
const noMembersLeft = sql`NOT EXISTS (SELECT 1 FROM ${user2conversation} WHERE ${user2conversation.conversationId} = ${convo.id})`;
await Promise.all([
  ctx.drizzle.delete(conversation).where(and(eq(conversation.id, convo.id), noMembersLeft)),
  ctx.drizzle.delete(conversationComment).where(and(eq(conversationComment.conversationId, convo.id), noMembersLeft)),
]);
```

Last-member exit drops from three sequential round-trips to two.

Verified rather than assumed:
- Rendered both statements via `toSQL()` — valid MySQL, correctly parameterised, and re-rendering proves the shared `SQL` fragment isn't mutated by being reused across two queries.
- Both predicates are index-backed: the subquery is a PK-prefix probe on `UsersInConversation (conversationId, userId)`, and the comment delete uses `ConversationComment_conversationId_idx`.
- Exercised end-to-end on a local dev server: last-member exit deleted the conversation row and left **0** orphaned comments; a multi-member exit correctly preserved the conversation (2 → 1 members, comments intact).

PR #1420, opened 2026-07-31
DEBUG: diagnose missing backend Sentry errors on Vercel

MathiasGruber:
Temporary debug branch. Adds `/api/sentry-probe` and `debugSentry.*` tRPC procedures to determine why server-side errors do not reach Sentry from Vercel deployments while they do locally.

Not for merge — findings will be moved into a clean PR.


MathiasGruber:
Diagnosis complete. Root cause and verified fix are in #1421.

PR #1440, opened 2026-08-16
Offense and Mastery System

### LINKED ISSUE #1277: Offense and Mastery System
Ninjutsu, Genjutsu, Taijutsu, Bukijutsu, offense, and defense will be replaced with Offense Stat and Defense Stat . 

Generals will remain as they are now at: Intelligence, Speed, Strength Willpower. 

Mastery: Mastery serves as a stat to gatekeep jutsu's and items locked behind hitting these requirements. 

Masteries do not count towards users experience and they do not affect combat damage dealt or taken. 

Masteries are:
Ninjutsu Mastery
Genjutsu Mastery
Taijutsu Mastery
Bukijutsu Mastery
Bloodline Mastery
Sage Mastery

Additional Mastery:

Add an Increase Mastery Tag that can increase players mastery. 

Bloodlines, Armor and Skill Tree when they are equipped can be used to increase a user's mastery granting them access to jutsus items and weapons. 

If a user mastery drops below the required level they are no longer able to use said item, jutsu or armor in combat 

New Caps:

- Offense: 1,322,420
- Defense: 1,322,420
- Generals: 400000
- Mastery: 1,500,000

Masteries uses the current timed training system. To allow players to train their offense/defense/Generals and Masteries at the same time

Implementation nots: 
All the respective Offenses(Ninjutsu Offense, Genjutsu Offense, Taijutsu Offense, Bukijutsu Offense) can be renamd to their masteries instead. The defenses can be removed. 

And then a new offense and defense stat type can be created.

MathiasGruber:
## Summary
- Replaces type-specific Ninjutsu/Genjutsu/Taijutsu/Bukijutsu offence and defence with a single **Offence** and **Defence** combat stat. Damage and experience use only these plus the four generals.
- Adds six **masteries** (Ninjutsu, Genjutsu, Taijutsu, Bukijutsu, Bloodline, Sage) that gate jutsu, items, and armor. Masteries do not grant experience or change combat damage.
- Adds `increasemastery` / `decreasemastery` tags, dual training slots (combat stats and a mastery in parallel), and a data-preserving migration that copies the highest old offences/defences and renames old offences to masteries.

Closes #1277

## Test plan
- [ ] Train a combat stat and a mastery at the same time; confirm combat training grants XP and mastery training does not
- [ ] Confirm changing training speed is blocked while either slot is active
- [ ] Assign unused XP / redistribute stats: only offence, defence, and generals appear
- [ ] Jutsu/item/armor with mastery requirements are hidden or unusable in combat when below the threshold, and usable when met
- [ ] Increase Mastery tag from bloodline/armor/skill tree raises mastery in combat
- [ ] After migration, existing characters keep GREATEST(old offences) as offence, GREATEST(old defences) as defence, and old offences as the four jutsu masteries
- [ ] Tutorial training step trains offence

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added six mastery categories: Ninjutsu, Genjutsu, Taijutsu, Bukijutsu, Bloodline, and Sage.
  * Added dedicated mastery training with progress tracking, caps, and separate start/stop controls.
  * Added mastery requirements for jutsus, items, evolutions, and combat effects.
  * Consolidated combat stats into shared Offence and Defence values.
  * Updated profile, training, combat, and equipment screens to display the new stat system.

* **Bug Fixes**
  * Improved stat allocation, cap handling, requirement validation, and equipment usability checks.
  * Added safeguards for training and mastery effects during combat.


theeneon:
@MathiasGruber when merging this you will need to add a query to take the players highest(nin,gen,tai,buki)offense and defense stat and apply it to the new Offense and Defense stats.

app/drizzle/migrations/0032_fearless_sauron.sql
```diff
@@ -0,0 +1,114 @@
...
+ALTER TABLE `UserData` ADD `bukijutsuMastery` double DEFAULT 10 NOT NULL;
+ALTER TABLE `UserData` ADD `bloodlineMastery` double DEFAULT 10 NOT NULL;
+ALTER TABLE `UserData` ADD `sageMastery` double DEFAULT 10 NOT NULL;
+ALTER TABLE `UserData` ADD `masteryTrainingStartedAt` datetime(3);
+ALTER TABLE `UserData` ADD `currentlyTrainingMastery` enum('ninjutsuMastery','genjutsuMastery','taijutsuMastery','bukijutsuMastery','bloodlineMastery','sageMastery');
+UPDATE `UserData`
+SET
+	`offence` = GREATEST(`ninjutsuOffence`, `genjutsuOffence`, `taijutsuOffence`, `bukijutsuOffence`),
+	`defence` = GREATEST(`ninjutsuDefence`, `genjutsuDefence`, `taijutsuDefence`, `bukijutsuDefence`),
+	`ninjutsuMastery` = `ninjutsuOffence`,
```
MathiasGruber: **Blocker: this backfill is asymmetric with the Jutsu requirement backfill below, and turns evolution-only thresholds into usage gates.**

Line 27-30 derive each mastery from the old **offence** column only:
```sql
`ninjutsuMastery` = `ninjutsuOffence`,
```
while line 79-92 derive each jutsu's requirement from **both** old columns:
```sql
`requiredNinjutsuMastery` = ... GREATEST(COALESCE(`requiredNinjutsuOffence`,0), COALESCE(`requiredNinjutsuDefence`,0))
```

That asymmetry only matters because this PR also changes what those columns *mean*. On `main` they are `// Evolution fields` in `schema.ts` and are read only by `canEvolveJutsu` — `git show main:app/src/libs/train.ts` confirms `canUseJutsu` never touched them, so they never gated *using* a jutsu you already own. This PR adds `hasMasteryRequirements(userdata, jutsu)` to `canUseJutsu` (`libs/train.ts:284`) and to `availableUserActions` (`libs/combat/actions.ts:184`).

Concrete case: a jutsu with `requiredTaijutsuOffence = 200000, requiredTaijutsuDefence = 320000`, owned and equipped by a user with `taijutsuOffence = 210000, taijutsuDefence = 330000`. Today they can use it unconditionally. After migration the jutsu requires `320000` mastery but the user's `taijutsuMastery = 210000` (their 330k defence investment goes only into the unified `defence` column on line 26). The jutsu disappears from their combat action list and `/jutsus` shows "You do not have the required mastery to use this jutsu." Masteries aren't part of redistribution (`getAssignedCombatStatTotal` covers offence/defence/generals only), so the only route back is grinding the new mastery slot. A defence-only requirement (`offence req NULL`, `defence req 300000`) is the worst case: mastery is backfilled from an offence value that may be as low as 10.

Two things to decide:
1. Make the halves symmetric — `ninjutsuMastery = GREATEST(ninjutsuOffence, ninjutsuDefence)` and so on — so nobody drops below a threshold they already met.
2. Decide explicitly whether these thresholds should become usage gates at all. If they were only ever meant as evolution prerequisites, leave `required*Mastery` NULL where `parentJutsuId IS NOT NULL` and keep the check in `canEvolveJutsu`.

Either way, worth sizing first:
```sql
SELECT COUNT(*) FROM Jutsu WHERE requiredNinjutsuDefence IS NOT NULL OR requiredGenjutsuDefence IS NOT NULL OR requiredTaijutsuDefence IS NOT NULL OR requiredBukijutsuDefence IS NOT NULL;
```
I couldn't run this against `tnr/main-1` myself. If it returns 0 the whole issue is moot — but the `GREATEST`/`COALESCE` handling on line 79 suggests such rows exist.

Related: `requiredStrength`/`Speed`/`Intelligence`/`Willpower` sit in the same `// Evolution fields` block and are **not** promoted to usage gates, so after this PR half that field group gates usage and half doesn't.

MathiasGruber (commented):
## Review: Offence & Mastery System

Reviewed at `4a265f89b`. I checked out the branch, ran `bun run typecheck` (**passes**), `bun run lint` (**passes**), the unit suite (**516 pass**; the 21 failing files fail identically on `main` from missing local env), `biome check` on every changed file (only the pre-existing `validators/combat.ts` format error remains), and I wrote throwaway tests to trace the mastery buff lifecycle empirically. Then I ran a 14-agent adversarial review across seven lenses and verified each surviving finding by hand.

**Already fixed by `44c5f76c` + `4a265f89` while I was reviewing** — no action needed, listing so it's clear they were checked: the `sageMode.ts` typecheck blocker (it needed the six new `required*Mastery: null` keys, not just the eight removed ones — `test_typecheck`/`test_build`/both Vercel deployments were red before this), the vestigial `highestOffence`/`highestDefence`/`targetHighest*` fields, dead `getStatTypeFromStat`, dead `normalizeBattleUsersState`, dead `getMasteryRequirementEntries`/`emptyMasteries`, the `getAffected` message reporting "Ninjutsu" for a unified-stat change, the `sageMastery` naming-collision docs, helper/import ordering in `train.ts`, and the `profile/edit` format violation. Good cleanup.

The overall shape of the change is sound and the migration is genuinely thoughtful. What's left below is grouped into three buckets.

### 1. The one I'd block on

**The migration converts evolution-only thresholds into usage gates, asymmetrically.** On `main`, `requiredNinjutsuOffence` et al. are labelled `// Evolution fields` in `schema.ts` and are read **only** by `canEvolveJutsu` — `canUseJutsu` never checked them, so they never gated using a jutsu you already own. This PR adds `hasMasteryRequirements` to `canUseJutsu` *and* `availableUserActions`, while the migration derives user mastery from the old **offence** column alone but derives each jutsu's requirement from `GREATEST(old offence req, old defence req)`. Any jutsu whose defence threshold exceeded its offence threshold now gates above what the migrated user has. See the inline comment on the migration.

### 2. Gameplay / data correctness worth resolving before merge

- Items are not mastery-gated at equip time, only silently neutered in battle (inline on `combat.ts`).
- `scaleUserStats` normalises masteries as a *share of experience*, but masteries are absolute thresholds — scaled AI lose their own jutsu, and `availableUserActions` has no `isAi` escape hatch (inline on `profile.ts:226`).
- The same function now splits the experience budget over 6 stats instead of 12, which exactly doubles every runtime-scaled entity's offence/defence given the current `ai.sql` seed values (inline on `profile.ts:207`).
- Two training slots both consume `dailyTrainings` and both credit `minutes_training` (inline on `train.ts`).
- Redistribution budget changes meaning; this is a real bug fix but an unannounced economy change (inline on `blackmarket.ts`).
- Mastery gains have no server-side cap and nothing ever clamps them afterwards (inline on `train.ts:281`).
- Mastery gates are a no-op in ranked PvP (inline on `constants.ts:1867`).

### 3. Simplification / consistency — the "could this be simpler" question

The biggest one: `increasemastery`/`decreasemastery` apply to `newTarget` while the structurally identical `increasestat`/`decreasestat` apply to `curTarget`. That single divergence is what forces the entire new `storeMasteryBases`/`resetMasteriesToBase` mechanism, and it leaves `baseStatsForModifiers` with two different lifetimes inside one object. Details inline on `process.ts:1215`.

Also flagged inline: `getSoftCappedExperience` is dead code that gained new unit tests; `canEquipJutsu` in `jutsus/page.tsx` re-implements `canUseJutsu` and drifted; `statTypes` is now decorative on damage tags but still shifts the stat-vs-generals balance; `MAX_STATS_CAP = 1322420` is unexplained; the damage simulator has a stale `- 120` baseline and unmigrated saved state; the Simple stat presets preview a split they no longer apply.

### Out-of-diff files (can't be commented inline)

- **`app/src/server/api/routers/item.ts:2578`** and **`app/src/libs/item.ts:382`** — the equip-requirement blocks check `requiredLevel` and `bloodlineId` but not the new `required*Mastery` fields. See the `combat.ts` comment for the full failure path.
- **`app/src/server/api/routers/stealth.ts:444`** — the pre-existing covert-training slot is the closest precedent for both the cap (`LEAST(x + gain, CAP)`) and the daily-budget question (it consumes no `dailyTrainings`). Worth mirroring.

### Notes, no action needed

- `app/data/ai.sql` seeds `bloodlineMastery` and `sageMastery` at `10` for every AI, so no AI can ever use Bloodline- or Sage-mastery-gated content. Fine if intentional, but it means those two masteries are player-only gates today.
- `preferredStat` is now much weaker than its name suggests: it only picks `highestMasteryType`, whose sole consumer is `getEfficiencyRatio` tag matching. It no longer influences damage at all.
- The `vi.mock("@/server/db", ...)` lines added to `damage_formula.test.ts` and `damage_modifiers.test.ts` are unrelated to this PR — those files fail on `main` too, purely from missing local env, and only 3 of ~21 affected files got the workaround.

### Test-plan gaps

Nothing covers: the `decreasemastery` *percentage* path, mastery cap enforcement, the migration itself, or `scaleUserStats` mastery behaviour at a level other than the entity's own. The migration items above are the ones I'd most want a staging dry-run against `tnr/main-1` for.

app/src/server/api/routers/combat.ts
```diff
@@ -3017,7 +2991,10 @@ export const processUsersForBattle = async (
               ui.equipped = "NONE" as const;
             } else {
               // Add item effects to user (only if user has required bloodline)
-              if (!ui.item.bloodlineId || ui.item.bloodlineId === user.bloodlineId) {
+              if (
+                (!ui.item.bloodlineId || ui.item.bloodlineId === user.bloodlineId) &&
+                hasMasteryRequirements(user, ui.item)
```
MathiasGruber: **Items are not mastery-gated at equip time — the gate only takes effect here, silently.**

`toggleEquipItem` (`routers/item.ts:2577-2599`) checks level, bloodline, crafting, auction and imbuement, but nothing checks `required*Mastery`. Same in the ranked-loadout validator at `libs/item.ts:382`. So an ARMOR with `requiredNinjutsuMastery = 500000` equips cleanly for a player with `ninjutsuMastery = 100000` and shows as equipped on their profile.

Then at battle start this line skips realizing its effects — but the item is still pushed into `user.items` with `equipped !== "NONE"` a few lines down (only the bloodline gate blocks that push), so the ARMOR durability loops in `process.ts` keep decrementing `ui.durability` on every hit. Net result: full durability wear, zero damage reduction, and no error at equip time or line in the combat log explaining why. Equipped WEAPONs are likewise silently dropped from the action bar by `actions.ts:241`.

Note the asymmetry with jutsu, which this PR *does* gate at equip: `canUseJutsu` gained `hasMasteryRequirements`, and `toggleEquip` for jutsu routes through it.

Suggest adding the guard inside the `if (doEquip)` block in `item.ts:2577` (the loadout path at `item.ts:2103` also routes through `toggleEquipItem`, so one guard covers both), naming the missing mastery in the error. And set `ui.equipped = "NONE"` here so pre-existing equips stop burning durability for nothing.

app/src/libs/profile.ts
```diff
@@ -191,55 +192,79 @@ export function scaleUserStats(
...
+
+  const masterySum = [
+    user.ninjutsuMastery ?? 0,
+    user.genjutsuMastery ?? 0,
+    user.taijutsuMastery ?? 0,
+    user.bukijutsuMastery ?? 0,
+    user.bloodlineMastery ?? 0,
+    user.sageMastery ?? 0,
+  ].reduce((a, b) => a + b, 0);
+  const calcMastery = (stat: keyof MasteryDistribution) => {
```
MathiasGruber, line 217: **Masteries are absolute thresholds, but this scales them as a proportional share of experience — so scaled AI lose their own jutsu.**

`calcMastery` gives each mastery `10 + (mastery/masterySum) * exp`, where `exp` comes from the entity's *current* level. For an AI scaled down to a low-level player via `scaleTarget` (`routers/combat.ts:2014-2017`), `exp` at level 10 is 27,000, so the four jutsu masteries land at roughly 6,760 (× `statsMultiplier`). Any equipped jutsu with a mastery requirement above that now fails `hasMasteryRequirements`.

That was harmless before this PR because nothing read these values as gates. Now `availableUserActions` (`actions.ts:184`) filters on them with **no `isAi` exemption** — unlike the loadout filter in `routers/combat.ts:2894`, which is explicitly `&& !user.isAi`, and `canUseJutsu` itself, which early-returns `true` for AI. So the AI passes the loadout filter, enters battle with the jutsu equipped, and then can't see it. A scaled AI whose whole kit is gated ends up with only basic actions.

Two options: either exempt AI from the mastery gate in `availableUserActions` (consistent with every other AI exemption in the codebase), or don't scale masteries proportionally at all — they aren't experience-derived, so carrying the stored values through unchanged is arguably the correct behaviour for a threshold.

Related: `app/data/ai.sql` seeds `bloodlineMastery` and `sageMastery` at `10` for every AI, so Bloodline/Sage-gated content is unusable by AI regardless of scaling.

app/src/libs/profile.ts
```diff
@@ -191,55 +192,79 @@ export function scaleUserStats(
...
+    user.offence ?? 0,
+    user.defence ?? 0,
     user.strength ?? 0,
     user.intelligence ?? 0,
     user.willpower ?? 0,
     user.speed ?? 0,
   ].reduce((a, b) => a + b, 0);
-  const calcStat = (stat: keyof StatDistribution) => {
-    return 10 + Math.floor(((user[stat] ?? 0) / sum) * exp * 100) / 100;
+  const calcCombatStat = (stat: keyof CombatStatDistribution) => {
```
MathiasGruber, line 198: **This halves the divisor, which exactly doubles offence/defence for every runtime-scaled entity.**

`combatSum` is now over 6 fields where it used to be over 12. The `ai.sql` seed values are the *fixed point* of the old formula, so this isn't a neutral refactor — I checked the arithmetic against a real row:

`Clown of Chaos` has all six stat columns at `834780` and `statsMultiplier = 3`, level 100 (`exp = 3,339,000`).
- Old: `(10 + 3,339,000/12) x 3 = 278,260 x 3 = 834,780` — identical to the stored value, i.e. scaling was idempotent.
- New: `(10 + 3,339,000/6) x 3 = 556,510 x 3 = 1,669,530` — exactly 2x.

That applies to every summon, every clone, and every `scaleTarget` overworld/quest AI. Since `sqrt` feeds the damage formula, their stat-derived attack and defence power both rise ~41%. Non-scaled AI keep their stored values and are unaffected, so PvE difficulty now diverges depending on whether the encounter scales.

Either re-seed `ai.sql` for the 6-stat formula, or normalise differently so the seed stays a fixed point. Whichever you pick, it's a balance change that deserves a line in the PR description.

app/src/server/api/routers/train.ts
```diff
@@ -155,54 +164,28 @@ export const trainRouter = createTRPCRouter({
               ? {
                   experience: sql`experience + ${trainingAmount}`,
                   dailyTrainings: sql`dailyTrainings + 1`,
```
MathiasGruber, line 167: **Both training slots consume the same daily allowance, so running them in parallel halves the effective cap.**

`stopTraining` here and `stopMasteryTraining` at line 281 both do `dailyTrainings + 1`, and `assertCanStartTraining` gates both on `dailyTrainings >= MAX_DAILY_TRAININGS`. A player using both slots — which the PR explicitly encourages ("Train in parallel with combat stats") — burns the daily budget twice as fast, and `showTrainingCapcha` fires twice as often for the same wall-clock time.

Both also credit `minutes_training` (lines 153 and 269) from the same elapsed wall clock, so parallel training double-counts that quest objective.

The existing precedent points the other way: the covert-training slot in `routers/stealth.ts` is a second parallel training track and consumes no `dailyTrainings` at all. If masteries are meant to be a genuinely free parallel track, drop the increment here; if they should share a budget, the cap probably needs raising and `minutes_training` should be credited from one slot only.

app/src/server/api/routers/train.ts
```diff
@@ -228,19 +211,134 @@ export const trainRouter = createTRPCRouter({
...
+      const trained = user.currentlyTrainingMastery;
+      const [result] = await Promise.all([
+        ctx.drizzle
+          .update(userData)
+          .set({
+            masteryTrainingStartedAt: null,
+            currentlyTrainingMastery: null,
+            ...(trainingAmount > 0
+              ? {
+                  dailyTrainings: sql`dailyTrainings + 1`,
```
MathiasGruber, line 283: **No server-side mastery cap, and unlike combat stats nothing ever clamps masteries afterwards.**

These six increments are bare `sql\`x + amount\``, and `assertCanStartTraining` has no cap check either — the only gate is the client greying out the button, which a direct tRPC call (or the MCP tool this mutation exposes) bypasses.

For combat stats an overshoot eventually gets corrected: `capUserStats` runs in `profile.assignStats` and writes the clamped values back. Masteries have no such path — `assignStats` clamps `user.ninjutsuMastery` in memory but its update payload omits masteries entirely, and the `capUserStats` call in `initiateBattle` only mutates the in-memory battle copy. So a mastery pushed past `MAX_MASTERY_CAP` (or past a GENIN's 60,000 rank cap) stays there permanently in the DB while combat silently clamps it — the player sees a mastery high enough to satisfy a gate, learns and equips the jutsu, and it never appears in battle.

`routers/stealth.ts:444` already has the pattern for this:
```ts
stealth: sql`LEAST(${userData.stealth} + ${gained}, ${STEALTH_SENSORY_CAP})`
```
Worth using `LEAST(..., mastery_cap)` here, plus an "Already capped" guard in `assertCanStartTraining` so the MCP path can't start a pointless session.

app/src/server/api/routers/blackmarket.ts
```diff
@@ -609,22 +610,16 @@ export const blackMarketRouter = createTRPCRouter({
         return { success: false, message: "Not enough reputation points" };
       }
       const inputSum = round(Object.values(input).reduce((a, b) => a + b, 0));
-      const availableStats = round(user.experience + 120);
+      const availableStats = round(getAssignedCombatStatTotal(user));
```
MathiasGruber, line 613: **This is a real bug fix, but it silently changes the redistribution economy — worth calling out in the PR description.**

The old `user.experience + 120` was broken in both directions: it granted stat points for experience that had never been assigned (`earnedExperience` isn't zeroed by redistribution, so you could bank the same XP twice), and the exact-sum check made redistribution outright unsubmittable for anyone holding unassigned XP. Using the assigned-stat sum fixes both. Good change.

The wrinkle is the interaction with the migration: `GREATEST()` collapses eight stats into two while leaving `experience` untouched. A maxed pre-migration player goes from `8x450,000 + 4x200,000 = 4.4M` assigned points down to `450,000 + 450,000 + 800,000 = 1.7M`, while their `experience` stays at ~3.34M. Under the old formula they'd get 3.34M points to redistribute into the new higher caps; under the new one they get 1.7M. That's the correct conservative behaviour, but it's a visible nerf to a paid action (`COST_RESET_STATS` reputation) that players will notice and report as a bug.

Also: `capUserStats` still isn't applied here, and `createStatSchema()` is built with no user so per-field maxima are the global caps rather than the caller's rank caps. Pre-existing, but concentrating everything into two stats makes it easier to land above your rank cap than it was across eight.

app/src/libs/combat/process.ts
```diff
@@ -1207,6 +1211,10 @@ export const applySingleEffect = (
           info = absorb(effect, usersEffects, consequences, curTarget);
         } else if (effect.type === "increasestat") {
           info = increaseStats(effect, newUsersEffects, curTarget);
+        } else if (effect.type === "increasemastery") {
+          info = increaseMastery(effect, newUsersEffects, newTarget);
```
MathiasGruber, line 1221: **The simplification question: these two lines apply to `newTarget` while the structurally identical `increasestat`/`decreasestat` on line 1213/1229 apply to `curTarget`.**

That one divergence is the entire reason `storeMasteryBases` + `resetMasteriesToBase` had to be invented. I confirmed the existing semantics empirically with a throwaway test: an `increasestat` static buff writes to `curTarget`, which is the transient pre-round copy that gets discarded, so stat buffs are recomputed from the persisted base every round and need no reset machinery. Masteries write to `newTarget`, which persists — hence the need to reset them at the top of `applyEffects` (line 294) and to remember bases in `util.ts:2953`.

I don't think the choice is *wrong*: the mastery gate in `availableUserActions` reads `battle.usersState`, so a mastery buff genuinely has to persist to affect next round's action list. But two consequences are worth weighing:

1. `baseStatsForModifiers` now has two different lifetimes in one object — the stat keys written by `applyPercentageStatModifier` live for a single round on a discarded copy, while the mastery keys written by `storeMasteryBases` persist across the battle. Anything that clears or copies that object has to know which half it's touching, which is exactly why `clone` and `summon` needed `delete newAi.baseStatsForModifiers`. That's a sharp edge for whoever next touches this.
2. The two tag families now behave observably differently for content authors: a mastery buff cannot stack, a stat buff is recomputed from base each round. Same tag shape, different rules, no comment explaining why.

If persistence is required, a short comment at line 294 and on `storeMasteryBases` explaining *why* masteries diverge from stats would go a long way. The alternative worth considering is deriving the effective mastery for gating from base + active mastery effects at read time in `availableUserActions`, which lets masteries use `curTarget` like every other buff and deletes both new helpers.

app/drizzle/constants.ts
```diff
@@ -793,8 +810,9 @@ export const SP_PER_LVL = 50;
 export const CP_PER_LVL = 50;
 export const MAX_ATTRIBUTES = 5;
 export const RYO_CAP = 3000000000;
-export const MAX_STATS_CAP = 450000;
-export const MAX_GENS_CAP = 200000;
+export const MAX_STATS_CAP = 1322420;
```
MathiasGruber, line 813: **`1322420` is an unexplained magic number, and `MAX_GENS_CAP` silently doubles.**

I couldn't derive `1322420` from anything in the repo — it isn't the old total stat budget (`8x450,000 + 4x200,000 = 4.4M` would need `1,400,000`), the old soft cap, or `calcLevelRequirements(100)`. If it came from a query against live player data (e.g. so nobody lands above cap post-migration), that reasoning is worth a comment; otherwise it reads as arbitrary.

These two constants have wide reach, and the changes ride along in a refactor PR without being mentioned in the description:
- `getSoftCappedExperience` goes from 3,050,000 to 4,244,840 (though see the separate note — it has no callers).
- `makeCappedNullableNumber(MAX_GENS_CAP)` doubles the allowed maximum for jutsu/item general requirements.
- `RANKED_PVP_STATS` equalises to the new values, so ranked stat magnitudes nearly triple.
- The stat-vs-generals ratio in the damage formula shifts: `sqrt(1,322,420)/sqrt(400,000) = 1.82` versus `sqrt(450,000)/sqrt(200,000) = 1.50`, so stats gain ~21% relative weight against generals for any asymmetric matchup.

app/drizzle/constants.ts
```diff
@@ -1819,14 +1862,14 @@ export const RANKED_PVP_STATS = {
...
-  ninjutsuDefence: MAX_STATS_CAP,
-  genjutsuOffence: MAX_STATS_CAP,
-  genjutsuDefence: MAX_STATS_CAP,
-  taijutsuOffence: MAX_STATS_CAP,
-  taijutsuDefence: MAX_STATS_CAP,
-  bukijutsuOffence: MAX_STATS_CAP,
-  bukijutsuDefence: MAX_STATS_CAP,
+  offence: MAX_STATS_CAP,
+  defence: MAX_STATS_CAP,
+  ninjutsuMastery: MAX_MASTERY_CAP,
```
MathiasGruber, line 1886: **Equalising all six masteries to `MAX_MASTERY_CAP` makes mastery gating a no-op in ranked.**

`RANKED_PVP_STATS` flows through `manuallyAssignUserStats` in `initiateBattle`, and `capUserStats` doesn't reduce it for CHUNIN+ since their `mastery_cap` *is* `MAX_MASTERY_CAP`. So every ranked participant satisfies every mastery requirement. On top of that, the `canUseJutsu` loadout filter in `routers/combat.ts:2883` is skipped entirely for `RANKED_PVP`/`RANKED_SPARRING`.

Defensible as "ranked equalises everything" — but it means mastery-gated jutsu are free in ranked and grinded-for everywhere else, which is a meaningful strategic asymmetry to decide deliberately rather than inherit. If the intent is that gates still apply, these six lines should use the rank-appropriate baseline instead of the cap.

app/src/libs/combat/tags.ts
```diff
@@ -1434,25 +1355,10 @@ export const damageCalc = (
...
-        b = `${lower}Defence`;
-      }
-      if (a in origin && b in target) {
-        const left = origin[a as keyof typeof origin] as number;
-        const right = target[b as keyof typeof target] as number;
-        atkPowerFromStats += Math.sqrt(Math.max(0, left));
-        defPowerFromStats += Math.sqrt(Math.max(0, right));
-      }
-    });
+    if (effect.statTypes && effect.statTypes.length > 0) {
```
MathiasGruber, line 1358: **`statTypes` is now purely decorative here, but dropping the per-type accumulation shifts the damage balance.**

Only `.length > 0` matters, so `["Ninjutsu"]` and `["Ninjutsu","Genjutsu","Taijutsu","Bukijutsu"]` are now identical. Correct for a unified stat — but the old loop accumulated `sqrt(stat)` *once per listed type*, and four-type lists are common in existing content (several rows in `ai.sql` carry all four).

Since `powerEffect` only uses `atkPower/defPower` as a ratio, the effect isn't a straight nerf — it's a reweighting. For a damage effect that listed four types, the stat term drops to a quarter of its former size while the generals term is unchanged, so `(4S_a + G_a)/(4S_d + G_d)` becomes `(S_a + G_a)/(S_d + G_d)`: generals now dominate the advantage calculation for exactly those jutsu. Worth a query for `damage`/`pierce` effects with `calculation: "formula"` and more than one `statTypes` entry to see how much content is affected.

Separately, since only the count matters, consider whether `statTypes` should stay on damage tags at all. It's still load-bearing for `weakness` matching (`calcDmgModifier`) and for `getEfficiencyRatio`, so it can't just be deleted — but a comment distinguishing "selects which stat" (no longer true) from "tags the damage for matching" (still true) would stop the next reader assuming the former.

app/src/libs/profile.ts
```diff
@@ -106,51 +106,52 @@ export const calcCP = (level: number) => {
...
 
 /**
- * The purpose of this function is to calculate the user experience, capped at the soft experience cap
- * i.e. the point where the user has one full offence, all 4 defences, and all 4 generals.
- * @param user
- * @returns
+ * Soft experience cap: one full offence, one full defence, and all 4 generals.
+ * Masteries do not count toward experience.
  */
 export function getSoftCappedExperience(user: UserData) {
```
MathiasGruber: **Dead code: this function has no callers, but the PR changed its formula and added two unit tests for it.**

`grep -rn getSoftCappedExperience app/` outside `node_modules` returns only this definition and `tests/libs/mastery.test.ts`. So the `5 * stats_cap` -> `2 * stats_cap` change has no runtime effect, and the new tests assert behaviour nothing depends on.

Either delete it along with its tests, or — if the soft cap was supposed to gate XP accrual and the wiring got lost at some point — that's a separate bug worth its own issue. Worth resolving either way, since a tested-but-unused function reads as load-bearing to the next person.

app/src/app/jutsus/page.tsx
```diff
@@ -363,6 +364,9 @@ export default function MyJutsu() {
           if (!checkJutsuBloodlineItem(uj.jutsu, userItems)) {
             warning = "You do not have the required bloodline item equipped.";
           }
+          if (!hasMasteryRequirements(userData, uj.jutsu)) {
```
MathiasGruber: **This warning is correct, but `canEquipJutsu` at line 1269 doesn't apply the same check — so the page contradicts itself.**

`categorizeJutsus` hand-rolls the eligibility test (`checkJutsuItems && checkJutsuBloodlineItem && checkJutsuElements && hasRequiredRank && hasRequiredLevel && checkJutsuRank && checkJutsuVillage && checkJutsuBloodline`) and omits `hasMasteryRequirements`. A mastery-blocked jutsu therefore lands in an available section with a working-looking Equip button, while this line simultaneously warns you can't use it. Clicking Equip succeeds — `toggleEquip` in `routers/jutsu.ts:1303` only calls `canUseJutsu` when `!isEquipped`, so the jutsu ends up equipped and then vanishes in combat.

The real fix is to stop duplicating the list: that chain is `canUseJutsu` minus the mastery check, so `canEquipJutsu` should just call `canUseJutsu(uj.jutsu, userData, true)` and inherit every future requirement automatically. As written, this is the second place that has already drifted from the canonical check.

app/src/layout/StatsDistributionForm.tsx
```diff
@@ -233,13 +230,22 @@ const SimpleDistribution: React.FC<SimpleDistributionProps> = (props) => {
     const pointsPerStat = Math.floor(availableStats / 4);
     const leftoverPoints = availableStats - pointsPerStat * 4;
 
+    let unusedPoints = 0;
```
MathiasGruber, line 234: **The confirm dialog previews an even 25% split, but this loop applies a different one.**

The preview at line 298-303 recomputes `pointsPerStat = floor(availableStats / 4)` per stat and lists `(+N)` for each. This loop then clamps each stat to its remaining room and dumps the whole shortfall into the first stats of `option.stats` in order, so whenever any stat is capped the applied split doesn't match what the user just confirmed. Either compute the distribution once and render it in both places, or drop the per-stat numbers from the preview.

Two related things while you're here:
- The spillover only redistributes within `option.stats`. If all four are capped, `unusedPoints` is silently dropped and the form total won't equal `availableStats`, which fails the "You must assign all points" tutorial check with no explanation.
- With masteries out of stat distribution, all four presets now assign offence + defence and differ only in their two generals — so "Taijutsu" means strength+speed and "Ninjutsu" means willpower+intelligence, while still carrying `IMG_TRAIN_*_OFF` artwork named after jutsu types that are no longer involved. Worth renaming these to what they actually do, or linking them to the new mastery training so the names mean something again.

app/src/app/manual/damage_calcs/[[...damageSimulationId]]/page.tsx
```diff
@@ -359,7 +359,7 @@ export default function Simulator(props: {
             <hr />
             <UserInput
               id="u1"
-              ignoreContains={showAll ? "Defence" : "None"}
+              ignoreContains={showAll ? "defence" : "None"}
```
Comment (MathiasGruber) [inline app/src/app/manual/damage_calcs/[[...damageSimulationId]]/page.tsx:360]: **Two leftovers in this page from the 12-stat world.**

**1. `calcExperience` (line 163) still subtracts a hardcoded `120`.** That constant was `12 stats x 10`; `statNames` now comes from `statSchema` and has six entries, so the baseline should be 60. On a fresh page load every stat defaults to 10, giving `60 - 120 = -60`, and the page renders `Experience: -60` with `calcLevel(-60)` collapsing to level 1. This is the same class of bug as the `experience + 120` in `blackmarket.ts` that the PR did fix — just the missed sibling. `statNames.length * 10` would keep it honest.

**2. Saved `DamageCalculation.state` JSON isn't migrated.** Rows saved before deploy hold `ninjutsuOffence` etc. Nothing maps them onto `offence`/`defence`, so `damageCalc` reads `Math.sqrt(Math.max(0, undefined))` and the history chart plots `NaN` (the point just disappears). Reloading an entry is worse than blank: `activateEntry` calls `setValue("ninjutsuOffence", ...)` for fields that no longer exist while `offence`/`defence` keep the schema prefault of 10, yet the still-valid generals *are* applied — so you get a confident, plausible-looking damage number computed for a 10-offence attacker with no warning.

Admin tooling only, and `createDamageSimulation` keeps just the newest 20 rows per user so stale entries age out. A `DELETE FROM DamageCalculation;` in the migration would be consistent with the `DELETE FROM Battle;` it already does, or `statSchema.safeParse` the state and skip rows that fail.

theeneon:
@MathiasGruber wouod you like to finish this one today so I can test it/make adjustments while I am at home today

PR #1441, opened 2026-08-16
New Item Features and Fixes

Phrosfire:
# Pull Request

**Changes:**
- Gives independent levels to items (Shares jutsu leveling logic where possible)
- Allows evolution of items  (Shares jutsu evolution logic where possible)
- Power scaling scales off the items level
- Items can only be leveled through pvp
- Displays item levels in inventory and home storage
- Added warning when evolving an imbued item into one that cannot be imbued
- Block imbuements from loading into battle if the imbuement is already on the item before getting disabled
- Allow users to return imbuements to inventory if imbuements were disabled for the item
- Fix jutsu experience multiplier
- Add item experience multiplier

## License

## Summary by CodeRabbit

* **New Features**
  * Added item leveling and experience gains from eligible PvP battles, including configurable boosts.
  * Added item evolution with prerequisites, eligibility checks, visibility controls, and progression resets.
  * Added level and experience displays across inventory, combat, and item details.
  * Added evolution configuration for items and jutsu.
  * Added staff controls for adjusting item progression.
  * Added automatic imbuement handling during evolution.
  * Added notifications for active item experience boosts.

* **Bug Fixes**
  * Item-based combat scaling now uses the item’s level.
  * Improved evolution-item purchase and crafting restrictions.
  * Improved imbuement removal ownership checks and crystal refunds.

### DIFF
--- a/app/src/app/items/page.tsx
+++ b/app/src/app/items/page.tsx
@@ -60,6 +60,7 @@ import {
   nonCombatConsume,
   partitionImbuementsForItemTransfer,
   showsItemLevelBadge,
+  userItemActionBadges,
 } from "@/libs/item";
 import { calculateKitsToUse, getRepairKits } from "@/libs/repair";
 import { showMutationToast, showRewardToast } from "@/libs/toast";
@@ -721,13 +722,7 @@ const Backpack: React.FC<BackpackProps> = (props) => {
     ...applyActiveVariant(useritem as UserItemWithVariants),
     ...useritem,
   }));
-  // Badges: red-bordered ownership level (bottom-left); amber stack quantity (bottom-right)
-  const itemStackCounts = items
-    ?.filter((ui) => !showsItemLevelBadge(ui) && ui.quantity > 1)
-    .map((ui) => ({ id: ui.id, quantity: ui.quantity }));
-  const itemLevels = items
-    ?.filter((ui) => showsItemLevelBadge(ui))
-    .map((ui) => ({ id: ui.id, level: ui.level }));
+  const itemBadges = userItemActionBadges(useritems);
   const sellPrice = calcItemSellingPrice(userData, useritem, structures);
   const repairItems = (useritems || []).filter(
     (userItem: UserItemWithRelations) =>
@@ -756,8 +751,8 @@ const Backpack: React.FC<BackpackProps> = (props) => {
       <ActionSelector
         className="grid-cols-6 pt-3 sm:grid-cols-4 md:grid-cols-4"
         items={items}
-        counts={itemStackCounts}
-        levels={itemLevels}
+        counts={itemBadges.counts}
+        levels={itemBadges.levels}
         selectedId={useritem?.id}
         showBgColor={false}
         showLabels={false}
@@ -1144,13 +1139,7 @@ const Character: React.FC<CharacterProps> = (props) => {
     ...applyActiveVariant(useritem as UserItemWithVariants),
     ...useritem,
   }));
-  // Badges: red-bordered ownership level (bottom-left); amber stack quantity (bottom-right)
-  const itemStackCounts = items
-    ?.filter((ui) => !showsItemLevelBadge(ui) && ui.quantity > 1)
-    .map((ui) => ({ id: ui.id, quantity: ui.quantity }));
-  const itemLevels = items
-    ?.filter((ui) => showsItemLevelBadge(ui))
-    .map((ui) => ({ id: ui.id, level: ui.level }));
+  const itemBadges = userItemActionBadges(useritems);
   const equipped = items?.find((item) => item.equipped === slot);
   const repairItems = (useritems || []).filter(
     (userItem: UserItemWithRelations) =>
@@ -1255,8 +1244,8 @@ const Character: React.FC<CharacterProps> = (props) => {
                 items={items?.filter(
                   (item) => slot?.includes(item.slot) && isEquippableUserItem(item),
                 )}
-                counts={itemStackCounts}
-                levels={itemLevels}
+                counts={itemBadges.counts}
+                levels={itemBadges.levels}
                 showBgColor={false}
                 showLabels={false}
                 greyedIds={items
--- a/app/src/app/items/page.tsx
+++ b/app/src/app/items/page.tsx
@@ -5,6 +5,7 @@ import {
   CircleDollarSign,
   CircleFadingArrowUp,
   Cookie,
+  Gem,
   Merge,
   Shirt,
   Split,
@@ -25,7 +26,11 @@ import {
 } from "@/components/ui/dialog";
 import { Input } from "@/components/ui/input";
 import { Label } from "@/components/ui/label";
-import { COST_EXTRA_ITEM_SLOT, IMG_EQUIP_SILHOUETTE } from "@/drizzle/constants";
+import {
+  COST_EXTRA_ITEM_SLOT,
+  IMG_EQUIP_SILHOUETTE,
+  ITEM_LEVEL_CAP,
+} from "@/drizzle/constants";
 import type {
   Item,
   ItemSlot,
@@ -45,6 +50,7 @@ import Loader from "@/layout/Loader";
 import Modal2 from "@/layout/Modal2";
 import NavTabs from "@/layout/NavTabs";
 import { applyActiveVariant } from "@/libs/combat/util";
+import { meetsEvolutionStatRequirements } from "@/libs/evolution";
 import {
   calcItemSellingPrice,
   calcMaxEventItems,
@@ -52,9 +58,12 @@ import {
   calcMaxMaterials,
   isEquippableUserItem,
   nonCombatConsume,
+  partitionImbuementsForItemTransfer,
+  showsItemLevelBadge,
 } from "@/libs/item";
 import { calculateKitsToUse, getRepairKits } from "@/libs/repair";
 import { showMutationToast, showRewardToast } from "@/libs/toast";
+import { hasRequiredLevel, remainingXpToLevel } from "@/libs/train";
 import type { UserWithRelations } from "@/routers/profile";
 import { useRequiredUserData } from "@/utils/UserContext";
 import { displayCostType } from "@/validators/item";
@@ -665,6 +674,38 @@ const Backpack: React.FC<BackpackProps> = (props) => {
       },
     });
 
+  const { mutate: evolveItem, isPending: isEvolving } = api.item.evolveItem.useMutation(
+    {
+      onSuccess: async (data) => {
+        showMutationToast(data);
+        if (data.success) {
+          await Promise.all([
+            utils.item.getUserItemsWithVariants.invalidate(),
+            utils.item.getEvolutions.invalidate(),
+            utils.item.getItemLoadouts.invalidate(),
+          ]);
+        }
+      },
+      onSettled,
+    },
+  );
+
+  const { mutate: removeImbuement, isPending: isRemovingImbuement } =
+    api.occupation.removeImbuement.useMutation({
+      onSuccess: async (data) => {
+        showMutationToast(data);
+        if (data.success) {
+          await utils.item.getUserItemsWithVariants.invalidate();
+        }
+      },
+      onSettled,
+    });
+
+  const { data: availableEvolutions } = api.item.getEvolutions.useQuery(
+    { itemId: useritem?.itemId ?? "" },
+    { enabled: !!useritem?.itemId && isOpen, staleTime: 5 * 60 * 1000 },
+  );
+
   // Derived
   const structures = userData?.village?.structures;
   const isLoading =
@@ -673,11 +714,20 @@ const Backpack: React.FC<BackpackProps> = (props) => {
     isSelling ||
     isEquipping ||
     isSplitting ||
-    isUsingRepairItem;
+    isUsingRepairItem ||
+    isEvolving ||
+    isRemovingImbuement;
   const items = useritems?.map((useritem) => ({
     ...applyActiveVariant(useritem as UserItemWithVariants),
     ...useritem,
   }));
+  // Badges: red-bordered ownership level (bottom-left); amber stack quantity (bottom-right)
+  const itemStackCounts = items
+    ?.filter((ui) => !showsItemLevelBadge(ui) && ui.quantity > 1)
+    .map((ui) => ({ id: ui.id, quantity: ui.quantity }));
+  const itemLevels = items
+    ?.filter((ui) => showsItemLevelBadge(ui))
+    .map((ui) => ({ id: ui.id, level: ui.level }));
   const sellPrice = calcItemSellingPrice(userData, useritem, structures);
   const repairItems = (useritems || []).filter(
     (userItem: UserItemWithRelations) =>
@@ -706,7 +756,8 @@ const Backpack: React.FC<BackpackProps> = (props) => {
       <ActionSelector
         className="grid-cols-6 pt-3 sm:grid-cols-4 md:grid-cols-4"
         items={items}
-        counts={items}
+        counts={itemStackCounts}
+        levels={itemLevels}
         selectedId={useritem?.id}
         showBgColor={false}
         showLabels={false}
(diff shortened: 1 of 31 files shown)


Phrosfire:
@coderabbitai review

Phrosfire:
@coderabbitai review

app/src/server/api/routers/item.ts
```diff
@@ -474,6 +572,177 @@ export const itemRouter = createTRPCRouter({
...
+      const canViewHidden = !!user && canChangeContent(user.role);
+      return filterVisibleEvolutions(evolutions, canViewHidden);
+    }),
+
+  evolveItem: protectedProcedure
+    .meta({ mcp: { enabled: true, description: "Evolve an item into its evolution" } })
+    .input(evolveItemSchema)
+    .output(baseServerResponse)
+    .mutation(async ({ ctx, input }) => {
+      const [{ user }, userItems, evolutionItem] = await Promise.all([
```
MathiasGruber: `evolveItem` costs 5 sequential DB round-trips on the happy path: this read batch -> the CAS `update` (L669) -> the two reads inside `finalizeItemEvolutionCleanup` -> that helper's write batch -> the `actionLog` insert (L732). Three of those are avoidable:

1. `fetchItemLoadouts(client, userId)` inside the helper depends on nothing computed here - hoist it into this `Promise.all`.
2. The `owned` re-read in the helper is redundant: `fetchUserItems` already loads `imbuements: { with: { item: true } }`, and evolving does not touch `UserItemImbuement` rows, so `userItemObj.imbuements` can be passed in. `rowsAffected === 1` (or the `alreadyEvolved` / `evolvedNow` branches) already proves the row now points at the evolution, which is all the `if (!owned) return false` check buys.
3. The `actionLog` insert can join the helper's final `Promise.all` rather than being awaited after it.

That takes the mutation down to 2 round-trips (one parallel read, one parallel write):

```ts
const [{ user }, userItems, evolutionItem, loadouts] = await Promise.all([
  fetchUpdatedUser({ client: ctx.drizzle, userId: ctx.userId }),
  fetchUserItems(ctx.drizzle, ctx.userId, { includeHidden: true }),
  fetchItem(ctx.drizzle, input.evolutionItemId),
  fetchItemLoadouts(ctx.drizzle, ctx.userId),
]);
```

then have `finalizeItemEvolutionCleanup` take `loadouts` + `imbuements` as args and return the write promises so the caller can `Promise.all` them together with the `actionLog` insert.

app/src/server/api/routers/item.ts
```diff
@@ -534,6 +803,97 @@ export const itemRouter = createTRPCRouter({
...
+      z.object({
+        userId: z.string(),
+        userItemId: z.string(),
+        level: z.number().int().min(1).max(ITEM_LEVEL_CAP),
+        experience: z.number().int().min(0).optional(),
+      }),
+    )
+    .output(baseServerResponse)
+    .mutation(async ({ ctx, input }) => {
+      const user = await fetchUser(ctx.drizzle, ctx.userId);
```
MathiasGruber: `fetchUser` and `fetchUserItems` are independent here (the second keys off `input.userId`, not off the role), so this is an avoidable serial round-trip - and `fetchUserItems` is a full-inventory read with the `imbuements`/`item` joins, so the latency is not trivial. Same pattern in `getPublicUserItems` above (L810/L823).

```ts
const [user, userItems] = await Promise.all([
  fetchUserItems(ctx.drizzle, input.userId, { includeHidden: true }),
]);
// guards after
```

Per CLAUDE.md, prefetching in parallel is preferred over sequencing behind a permission guard.

app/src/hooks/item.ts
```diff
@@ -106,6 +107,9 @@ export const useItemEditForm = (
     enabled: canBeCrafted as boolean | undefined,
   });
 
+  // All item names for evolution parent picker
+  const { data: allItemNames } = api.item.getAllNames.useQuery();
```
MathiasGruber, line 107: This unconditional `getAllNames` query resolves to the same React Query cache entry as the one at L107-109, so the `enabled: canBeCrafted` flag there is now dead — the query fires on every item edit page regardless of `canBeCrafted`.

Drop the conditional query and reuse `allItemNames` for both pickers (`allItemNames?.filter((i) => i.id !== item.id)` for the crafting requirement values as well).

app/src/libs/combat/database.ts
```diff
@@ -1335,6 +1337,24 @@ export const updateUser = async (
...
+    );
+    // Group equipped items by their definition's xpToLevel for batched CAS updates
+    const equippedXpGroups = new Map<number, string[]>();
+    if (iExp > 0) {
+      for (const ui of user.items) {
+        if (ui.equipped === "NONE" || ui.quantity <= 0) continue;
+        const def = getItem(curBattle, ui.itemId);
+        const xpToLevel = def?.xpToLevel ?? ITEM_XP_TO_LEVEL;
+        const ids = equippedXpGroups.get(xpToLevel) ?? [];
+        ids.push(ui.id);
```
MathiasGruber: Equipped items will never gain XP from `RANKED_PVP`, even though `RANKED_PVP` is listed in `ITEM_XP_BATTLE_TYPES`.

Ranked battles go through `initiateBattle` with `forceLoadouts`, which replaces `user.items` with synthetic rows built as `{ id: nanoid(), ... }` (combat.ts ~1780). So `ui.id` here is a random id that does not exist in `UserItem`, and both CAS updates below (`inArray(userItem.id, ids)`) match 0 rows. This PR deliberately loads the owned level/experience into ranked battles via `forceLoadoutItemProgression`, so items *matter* in ranked but can never level there.

Note the jutsu block right above works because it keys on `userJutsu.jutsuId`, not on the row id.

Fix: carry the real `UserItem.id` through the forced-loadout mapping (the progression query at combat.ts ~1700 doesn't even select `id` today — add it, and store it on the synthetic entry), then group on that id here. Alternatively key the update on `(userId, itemId)` like the jutsu path does.

app/src/server/api/routers/item.ts
```diff
@@ -434,9 +505,36 @@ export const itemRouter = createTRPCRouter({
...
+              ) as \`evoSiblingCount\`
+            ) < ${EVOLUTION_MAX_CHILDREN}`,
+          ]
+        : [];
+
+      const updateResult = await ctx.drizzle
+        .update(item)
+        .set(input.data)
+        .where(and(eq(item.id, input.id), ...evolutionUpdateGuards));
+      if (updateResult.rowsAffected === 0) {
```
MathiasGruber, line 547: This early return can permanently destroy an item's crafting recipe. The `craftingRequirement` rows for `input.id` are deleted unconditionally at line 504-506, *before* this guarded UPDATE. When the guard legitimately fails (parent item deleted concurrently, or the sibling cap was reached between the read and the write) we return here and the re-insert of `newRequirements` further down never runs — the item is left with zero crafting requirements and no way to recover except manual re-entry.

There's a second, more likely trigger: `rowsAffected` for an UPDATE is MySQL's *changed*-row count, and `input.data` doesn't touch `updatedAt`. An admin who edits only `craftingRequirements` produces a client-side diff (so the mutation fires) but writes identical values to every `Item` column, giving `rowsAffected === 0` → spurious "Update failed — parent may have been deleted" error *and* the wiped recipe.

Fix: move the `craftingRequirement` delete + re-insert after the guarded update succeeds (or run both in the same `Promise.all` at the end), and don't treat `rowsAffected === 0` as failure unless the evolution guards are actually present.

app/src/server/api/routers/item.ts
```diff
@@ -284,55 +304,74 @@ export const itemRouter = createTRPCRouter({
...
-        const variants = await ctx.drizzle.query.itemVariant.findMany({
-          where: eq(itemVariant.itemId, input.id),
-          columns: { id: true },
+      if (!canChangeContent(user.role)) {
+        return { success: false, message: `Not allowed to delete item` };
+      }
+
+      // Write-time guard: only delete if no child evolutions currently point here.
+      // Nested derived table avoids MySQL errno 1093 (can't update/delete target in FROM).
+      const deleteResult = await ctx.drizzle.delete(item).where(
```
MathiasGruber: This inverts the leaf-first delete ordering that the removed comment explicitly documented. The parent `Item` row is now deleted first, and `itemVariant` / `userItemVariant` / `userItem` / `userItemImbuement` are cleaned up afterwards in separate round-trips.

Failure scenario: the PlanetScale request for the variant cleanup fails (Vitess connection drop, timeout). The `Item` row is already gone, so `itemVariant` rows now point at a missing item, `getItemVariants` returns nothing for it, and there is no longer any admin path to find or clean those rows — the exact dangling-reference state the old ordering was designed to avoid. Same for the `userItem` rows in the `Promise.all` below, which then survive forever (silently hidden because `fetchUserItems` filters on `ui.item`).

Fix: keep the child-evolution check as a read-only guard (you already fall back to a `findMany` for the error message), do the leaf-first cleanup, and delete the `Item` row last.

app/src/server/api/routers/item.ts
```diff
@@ -474,6 +572,177 @@ export const itemRouter = createTRPCRouter({
...
+            return errorResponse(
+              "Evolution failed - item may have already been evolved",
+            );
+          }
+        }
+      }
+
+      // Fresh evolve with incompatible slot strips parent from loadouts; retries remap
+      // any leftover parent refs so cleanup stays recoverable.
+      const remapParentInLoadouts = didEvolveThisCall ? canKeepEquipped : true;
```
MathiasGruber: Loadouts are keyed by `itemId`, not `userItemId`, so this remap rewrites entries that belong to *other* copies of the parent item.

Scenario: user owns two Kunai and a loadout with `[{ITEM_1, kunai}, {ITEM_2, kunai}]`. They evolve one copy. `finalizeItemEvolutionCleanup` rewrites *both* entries to the evolution id, but they only own one evolution — `computeLoadoutAssignments` can now only fill one slot, so the second silently becomes an invalid item and the still-owned second Kunai is dropped from the loadout.

Related: the `alreadyEvolved` retry path reaches this cleanup with `didEvolveThisCall === true` even when nothing was evolved, so a user who simply owns an evolution item can call `evolveItem` on it and have their loadout entries for the parent item rewritten/removed with no evolution taking place (and still get a "Evolved into X!" success toast).

Fix: only remap a loadout entry when the user no longer owns any other copy of `parentItemId` (you already fetch `userItems` above), and skip the loadout mutation entirely on the `alreadyEvolved` no-op path unless an evolve actually happened in this call.

app/src/layout/PublicUser.tsx
```diff
@@ -1278,6 +1299,23 @@ const EditUserComponent: React.FC<EditUserComponentProps> = ({ userId, profile }
...
+    ...ui.item,
+    id: ui.id,
+    name:
+      ui.quantity > 1
+        ? `${ui.item.name} x${ui.quantity}`
+        : ui.equipped !== "NONE"
+          ? `${ui.item.name} [${ui.equipped}]`
+          : ui.item.name,
+  }));
+  const userItemCounts = userItems?.map((ui) => ({
```
MathiasGruber: This smuggles the item *level* through the `quantity` field of the `counts` prop, so it renders in the amber stack-count badge (bottom-right) instead of the red level badge. This PR adds a dedicated `levels` prop to `ActionSelector` for exactly this — use `levels={userItems?.map((ui) => ({ id: ui.id, level: ui.level }))}` and drop `counts` (or pass real quantities), so the staff view matches the badge semantics documented in `CombatActions.tsx` and used on `/items` and `/home`.

app/src/app/items/page.tsx
```diff
@@ -673,11 +714,20 @@ const Backpack: React.FC<BackpackProps> = (props) => {
...
-    isUsingRepairItem;
+    isUsingRepairItem ||
+    isEvolving ||
+    isRemovingImbuement;
   const items = useritems?.map((useritem) => ({
     ...applyActiveVariant(useritem as UserItemWithVariants),
     ...useritem,
   }));
+  // Badges: red-bordered ownership level (bottom-left); amber stack quantity (bottom-right)
+  const itemStackCounts = items
```
MathiasGruber: This exact stack-count/level split is now written out three times: here, again in `Character` (lines 1148-1153), and as `userItemActionBadges` in `app/src/app/home/page.tsx` (line 42). All three apply the same rule (`!showsItemLevelBadge(ui) && quantity > 1` → counts, `showsItemLevelBadge(ui)` → levels), so a change to the badge rule has to be made in three places.

Fix: move `userItemActionBadges` out of `home/page.tsx` into `@/libs/item` next to `showsItemLevelBadge` and call it from all three sites.

app/src/server/api/routers/item.ts
MathiasGruber: Done in 3163609dd: loadouts are fetched in the initial `Promise.all`, the cleanup re-read is gone (the imbuements already loaded by `fetchUserItems` are reused), and the `actionLog` insert joins the final write batch. The happy path is now 3 round-trips (parallel read → guarded CAS → parallel writes) — the cleanup writes have to stay behind the CAS because they are conditional on the evolve actually happening, so 2 isn't safely reachable.

app/src/server/api/routers/item.ts
MathiasGruber: Fixed in 3163609dd — both staff endpoints now run their fetches in one `Promise.all`, and `adjustUserItem` uses a targeted `userItem.findFirst` instead of the full-inventory read. Also moved both inline zod schemas to `validators/item.ts` and dropped the unused optional `experience` input.

app/src/hooks/item.ts
MathiasGruber: Fixed in 3163609dd — a single unconditional `getAllNames` query now feeds both the crafting-requirements picker and the evolution parent picker.

app/src/libs/combat/database.ts
MathiasGruber: Fixed in 3163609dd. The progression query now selects the real `UserItem.id`, and the synthetic loadout entries carry it as `progressionRowId` on `BattleUserItem`, which `updateUser` targets for the XP grant — the synthetic row id itself is kept so battle-end quantity/durability writes still can't touch real inventory rows. While in there, the two-statement CAS pair (for both items and jutsu) became a single atomic UPDATE (`level = level + (exp >= threshold)`, `experience = IF(...)`) via a shared helper — the pair raced itself inside the `Promise.all` (premature level-up or doubled XP depending on commit order).

app/src/server/api/routers/item.ts
MathiasGruber: Fixed in 3163609dd — the guarded UPDATE now runs first and sets `updatedAt` explicitly (making `rowsAffected` a reliable changed-rows signal even on a requirements-only save), and the `craftingRequirement` delete + re-insert happen as one ordered step only after the guard passes.

app/src/server/api/routers/item.ts
MathiasGruber: Fixed in 3163609dd — child evolutions and variants are read in the initial `Promise.all` as a read-only guard, cleanup is leaf-first (`userItemVariant` → `itemVariant` → `userItem`/`userItemImbuement`), and the parent `Item` row is deleted last, still guarded by the not-exists subquery against a concurrently added evolution. A partial failure now leaves a state the endpoint can retry.

app/src/server/api/routers/item.ts
MathiasGruber: Fixed in 3163609dd — loadout entries are only remapped/stripped when the user owns no other copy of the parent item, and the `alreadyEvolved` path no longer reports "Evolved into X!" (it returns a cleanup-only message and skips the actionLog). Related: incompatible imbuements removed by evolution now refund their crystals, matching `removeImbuement`'s policy for system-forced removals.

app/src/layout/PublicUser.tsx
MathiasGruber: Fixed in 3163609dd — the staff editor passes `levels={...}` and drops the counts/labelSingles hack (stack quantities already show as ×N in the item names).

app/src/app/items/page.tsx
MathiasGruber: Fixed in 3163609dd — `userItemActionBadges` now lives in `@/libs/item` next to `showsItemLevelBadge`, and all three sites (home page, Backpack, Character) call it.

MathiasGruber:
Pushed 3163609dd addressing all outstanding review threads (each replied/resolved inline). Beyond the threads, the commit also fixes issues surfaced in a deeper review pass:

- **Item XP CAS race**: the paired increment/level-up UPDATEs ran concurrently in one `Promise.all` and could double-grant XP or level up early depending on commit order — now a single atomic UPDATE per table, shared by jutsu and items (`xpGainOrLevelUpSet`).
- **AI gear nerf**: equipped-gear passive effects and item action costs now keep **character-level** scaling for AI users (they can never earn item XP), so bosses/raid content aren't silently pinned to item level 1. Player items scale from item level per the PR's design. Note existing player items still start at level 1 post-migration — that cliff is inherent to the design and may deserve a rebalance pass on `powerPerLevel`/`*CostReducePerLvl` content.
- **Stack progression loss**: `splitItemStack` now carries `level`/`experience` onto the split-off row, and the stack-merge bucket key includes `level` so merging can't silently drop a leveled stack's progression.
- **Spurious update failures**: `item.update` sets `updatedAt` explicitly so a no-change re-save can't read as a failed evolution guard (MySQL reports *changed* rows).
- Evolution-forced imbuement removals now **refund crystals**, matching `removeImbuement`'s policy.

Design notes left as-is (flagging for visibility): equipped consumables/thrown weapons still earn XP and scale invisibly (`showsItemLevelBadge` hides their badge), and evolution items can still be granted via quest rewards/staff tools/auctions — only shop buy and crafting are guarded. Both are one-line policy decisions if you want them changed.

MathiasGruber (approved):

PR #1446, opened 2026-08-18
Use correct loadout item copy

Phrosfire:
# Pull Request

When using loadout swap, it sometimes pulls in the wrong copy of an item.  So if you have two loadouts, each with the same weapon but different imbuement, it'll sometimes pull the wrong weapon when you swap.

This change will prioritize useritemID, then fall back to itemID if it can't be reached.

## License

## Summary by CodeRabbit

- **New Features**
  - Loadouts now preserve the specific inventory item assigned to each slot.
  - Equipped items synchronize more reliably when loadouts change.

- **Bug Fixes**
  - Improved loadout behavior for duplicate items and removed inventory items.
  - Evolution updates now retain compatible assignments and remove invalid ones.

- **Tests**
  - Added coverage for exact item matching, fallback selection, deleted items, and consumed duplicates.

### DIFF
--- a/app/src/libs/item.ts
+++ b/app/src/libs/item.ts
@@ -335,13 +335,25 @@ export interface ComputedLoadout {
 }
 
 /**
+ * Captures the equipped inventory rows for a loadout. `itemId` remains alongside
+ * the unique row id so legacy entries and entries whose row was removed can
+ * still fall back to another owned copy of the same item.
+ */
+export const buildItemLoadoutData = (
+  useritems: Array<{ id: string; itemId: string; equipped: ItemSlot }>,
+): Array<{ userItemId: string; itemId: string; slot: ItemSlot }> =>
+  useritems
+    .filter((ui) => ui.equipped !== "NONE")
+    .map((ui) => ({ userItemId: ui.id, itemId: ui.itemId, slot: ui.equipped }));
+
+/**
  * Pure decision logic for applying an item loadout. Validates every saved
  * entry against the user's current inventory and equip limits, assigning each
  * to a unique slot and a unique owned row. Skipped entries are reported in
  * `invalidItems`. No database access — fully unit-testable.
  */
 export const computeLoadoutAssignments = (
-  itemData: Array<{ itemId: string; slot: ItemSlot }>,
+  itemData: Array<{ userItemId?: string; itemId: string; slot: ItemSlot }>,
   useritems: UserItemWithRelations[],
   user: { level: number; bloodlineId: string | null },
   now: Date = new Date(),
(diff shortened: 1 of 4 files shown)


app/src/server/api/routers/item.ts
```diff
@@ -764,39 +765,35 @@ export const itemRouter = createTRPCRouter({
...
-                    eq(itemLoadout.id, loadout.id),
-                    eq(itemLoadout.userId, ctx.userId),
-                  ),
-                ),
-            );
-          });
-      }
+      const referencesEvolvedCopy = (entry: ItemLoadout["itemData"][number]) =>
+        entry.userItemId
+          ? entry.userItemId === input.userItemId
```
MathiasGruber, line 780: A *dangling* `userItemId` (an entry whose inventory row no longer exists) takes the truthy branch and is compared against `input.userItemId`, so it never reaches the legacy `itemId` remap. That is a regression against main for a state the codebase treats as normal: rows are deleted without any loadout cleanup (`sellUserItem` around line 1534, combat item destruction in `app/src/libs/combat/database.ts`), which is precisely why `computeLoadoutAssignments` was given the `itemId` fallback in this same PR.

Concrete case: loadout holds `{ userItemId: "<sold row>", itemId: P }`, and the user evolves their only remaining copy of `P`. On main the entry was remapped to the evolution; now it stays pointing at `P`, which nobody owns, and resolves to `Item not found` forever.

Fix — treat an entry whose row is gone as legacy:

```ts
const ownsSavedRow = (id?: string) => !!id && userItems.some((ui) => ui.id === id);
const referencesEvolvedCopy = (entry: ItemLoadout["itemData"][number]) =>
  ownsSavedRow(entry.userItemId)
    ? entry.userItemId === input.userItemId
    : !ownsOtherParentCopy && entry.itemId === parentItemId;
```

app/src/libs/item.ts
```diff
@@ -353,13 +365,17 @@ export const computeLoadoutAssignments = (
...
 
   for (const entry of itemData) {
-    // Prefer an unconsumed unit that is actually equippable (carried, not
-    // auction/crafting/imbuing); fall back to any unconsumed unit so a
-    // duplicated itemId still reports a precise reason instead of being
-    // dropped silently.
-    const owned = useritems.filter(
-      (it) => it.itemId === entry.itemId && !consumedRowIds.has(it.id),
-    );
+    const unconsumed = useritems.filter((it) => !consumedRowIds.has(it.id));
```
MathiasGruber: The behaviour here is correct, but the shape obscures it: `unconsumed` is a fresh full copy of the inventory built per entry, and the `catalogMatches.filter((it) => it.id !== savedUserItem.id)` on line 377 exists only to undo the duplicate that `unconsumed` itself introduced. Three intermediate arrays for what is a two-line preference rule.

Suggest collapsing to a single predicate so the "saved row first, then other copies" intent is readable at a glance:

```ts
const isFree = (it: UserItemWithRelations) => !consumedRowIds.has(it.id);
const savedUserItem = entry.userItemId
  ? useritems.find((it) => it.id === entry.userItemId && isFree(it))
  : undefined;
const owned = [
  ...(savedUserItem ? [savedUserItem] : []),
  ...useritems.filter(
    (it) => it.itemId === entry.itemId && isFree(it) && it.id !== savedUserItem?.id,
  ),
];
```

app/src/server/api/routers/item.ts
MathiasGruber: Fixed in 2f609dd35: `referencesEvolvedCopy` now gates on `ownsSavedRow`, so an entry whose saved inventory row no longer exists falls through to the legacy `itemId` path and gets remapped/stripped exactly as on main. The remap also now sets `userItemId: input.userItemId` alongside the new `itemId`, so legacy and dangling entries are upgraded to copy-specific references instead of keeping a dead row id.

app/src/libs/item.ts
MathiasGruber: Fixed in 2f609dd35: collapsed to the suggested `isFree` predicate shape (hoisted above the loop, since it closes over `consumedRowIds`) — the saved row is looked up directly in `useritems` and other copies are appended behind it, with no intermediate `unconsumed`/`catalogMatches` arrays. Also took the chance to type the parameter as `ItemLoadout["itemData"]` instead of restating the entry shape, and added a test for the duplicate-entry case where an earlier entry already consumed the saved row.

MathiasGruber (approved):

PR #1478, opened 2026-08-30
perf(profile): read global game state once per request, not once per procedure

MathiasGruber:
First fix out of the performance audit: the top-ranked finding, `fetchUpdatedUser` being re-run in full by every procedure in a tRPC batch.

## The problem

`fetchUpdatedUser` opens with a seven-way `Promise.all`. Six of those seven reads are the same for every caller in a request:

| read | predicate mentions the user? | mutated by callers? |
|---|---|---|
| achievement catalogue | no | yes — quest objectives are rewritten per user |
| game settings | no | no |
| unvoted-poll probe | yes, but read-only | no |
| all active wars | no | no |
| active shrine lobbies | no | no |
| active raids | no | no |
| **`userData` + relations** | **yes** | **yes — every mutation writes it back** |

`httpBatchLink` packs a page's whole mount into one POST and tRPC resolves those procedures concurrently, so a page load paid for all six once *per procedure*.

## The fix

A request-scoped memo (`AsyncLocalStorage`, entered once in the tRPC route handler), with the six qualifying reads routed through it. `src/server/requestScope.ts` documents the three conditions a read has to meet to belong there.

The `userData` row is deliberately **not** memoized. It is the row every mutation reads, mutates in memory and writes back, and it gates what actions a player is allowed to take — it stays a private per-caller read, exactly as before.

Two details that matter:

- **Every caller gets a `structuredClone`, the first one included.** Callers mutate what they read back: `controlShownQuestLocationInformation` rewrites quest objectives in place to hide locations from players outside the right sector, and that rewrite is per-user. Handing out the memoized object would leak one player's view into the next caller's copy — and `hospital.userHeal` really does call `fetchUpdatedUser` for two different users in one request.
- **`load()` is awaited into a real promise before it is memoized.** A Drizzle query builder is a thenable that re-runs `execute()` on every `then`, so memoizing the builder caches the *intent* to query, not its result — a cache that looks like it hits while every reader still goes to the database. This one actually bit during validation; there is a regression test for it.

Outside a request scope — crons, scripts, tests — `scopedRead` is a plain pass-through, so nothing off the tRPC path changes behaviour.

## Validation

Measured against a local dev server with a provisioned level-50 test account, instrumenting the PlanetScale driver's `fetch` and attributing every query to its request through the same `AsyncLocalStorage`. The workload is one batch of the five no-input query procedures that call `fetchUpdatedUser`: `profile.getUser`, `home.getUserHome`, `home.getAvailableUpgrades`, `bloodline.getSwapInfo`, `skillTree.getResetInfo`. The memo was toggled with an env flag so both builds ran identical instrumentation.

| | before | after |
|---|---|---|
| DB round-trips per batch | **37** | **13** |
| median batch latency (15 samples) | 299 ms | 286 ms |
| max latency | 331 ms | 328 ms |
| clone overhead | — | 1.4 ms/request, all six keys |

Per-shape, before → after: achievement catalogue 5→1, game settings 5→1, poll probe 5→1, active wars 5→1, shrine lobbies 5→1, active raids 5→1, `userData` 5→5 (unchanged by design).

Local wall-clock barely moves because the seven reads already run in parallel against a Docker MySQL on loopback. The win is the 24 fewer HTTPS round-trips to PlanetScale per page load, which is what costs money and queues under contention in production.

**Functional equivalence:** the full batch response was captured under both builds and diffed after normalising timestamps — 5147 lines of JSON, differing on a single line, the throwaway test account's username.

## Ordering safety

A memoized read returns the snapshot the first caller took, so a write landing mid-request is observed on the next request. I scanned every procedure that calls `fetchUpdatedUser` for a write to one of the six tables followed by a later `fetchUpdatedUser` in the same procedure: zero hits. Two procedures call `fetchUpdatedUser` twice, both concurrently at the top (`hospital.userHeal`, for two different users), never write-then-read.

## Checks

- `make typecheck` clean (with `tsconfig.tsbuildinfo` removed first)
- `make test` — 1300 pass, 0 fail, including 9 new tests for `scopedRead`
- `make lint` — no new diagnostics in the changed files

---

> [!NOTE]
> **Medium Risk**
> Changes the hot `fetchUpdatedUser` path with memo snapshot semantics and cloning; behavior is documented as equivalent but regressions would affect most page loads.
> 
> **Overview**
> Adds **request-scoped read deduplication** so batched tRPC page loads stop repeating the same global DB queries once per procedure.
> 
> The tRPC route handler now runs inside `withRequestScope` (`AsyncLocalStorage`). In `fetchUpdatedUser`, six shared reads—achievements, game settings, active wars, shrine lobbies, raids, and the unvoted-poll probe—go through `scopedRead` helpers so a batch pays **one round-trip each** instead of one per procedure. Per-user `userData` stays a fresh read every time.
> 
> `scopedRead` **clones results** with `structuredClone` so downstream per-user quest/war/raid rewriting cannot leak between concurrent callers, memoizes **resolved promises** (not Drizzle thenables), and **evicts failed reads** so a later caller can retry. Off the tRPC path it is a no-op.
> 
> Separately, `fetchPublicUsers` loads the caller with **`columns: { role: true }`** only for the IP visibility guard, plus new tests for `scopedRead` and public-user IP rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c8e8a0144479d291becf821dd4981aeb159755b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Performance**
  * Improved request handling by reusing repeated data lookups within a single request.
  * Reduced duplicate work for profile-related achievements, settings, polls, wars, shrines, and raids.

* **Reliability**
  * Ensured failed lookups can be retried without affecting later requests.
  * Preserved data isolation and safe results when the same information is requested concurrently.

* **Tests**
  * Added coverage for request isolation, concurrent reads, result copying, date handling, and retry behavior.

---

## Follow-up in this PR

**Helpers moved below `fetchUpdatedUser`.** They were below the router but in front of the function they serve; the file already puts that function's other private helper (`persistPassiveRegenToDb`) after it. Pure move.

**`fetchPublicUsers` selects `role` instead of the whole row.** It re-reads the caller's own `UserData` row to decide whether the listing may show IPs, and was selecting ~90 columns (including the `effects` JSON) to use exactly one field at the two `canSeeIps` guards. `profile.getPublicUsers` backs the online-players sidebar and appeared in **14 of 40** tRPC batches across a ~20-page browse, so it rides along on most page loads. Bandwidth only — the query still runs.

Because that touches a permission guard, it comes with a DB-backed suite (`app/tests/server/api/profile.publicUserIps.test.ts`) that runs `fetchPublicUsers` against a real MySQL for a staff role, an ordinary player, and a signed-out caller, in both the listing and IP-search directions. Verified it has teeth: swapping the projection to one that drops `role` fails two of its five cases rather than silently hiding IPs from staff.

## Why nothing else got the memo treatment

I re-instrumented and browsed ~20 pages as a logged-in level-50 account — 40 tRPC batches, 297 queries — counting identical statements repeated inside one request. **7 duplicates out of 297 (2%)**, and 4 of those are the `userData`-with-relations read that this PR deliberately excludes (3× on `/home`: `getUserHome` + `getAvailableUpgrades` + `getUser`). The other 3 are one-offs on `/users`.

Memoizing the user row would be the only remaining win, and it is not free: it changes when `persistPassiveRegenToDb`'s write becomes visible, so a procedure that today reads `updatedAt = now` and skips its own persist would always re-persist. Not worth it to save 2 reads on one page.

### DIFF
--- a/app/src/app/api/trpc/[trpc]/route.ts
+++ b/app/src/app/api/trpc/[trpc]/route.ts
@@ -4,6 +4,7 @@ import { cookies, headers } from "next/headers";
 import type { NextRequest } from "next/server";
 import { appRouter } from "@/api/root";
 import { createAppTRPCContext } from "@/api/trpc";
+import { withRequestScope } from "@/server/requestScope";
 
 export const runtime = "nodejs";
 export const maxDuration = 90;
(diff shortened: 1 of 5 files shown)


MathiasGruber:
Thanks — that nitpick is already addressed. The review anchors to `67e1cdb` (the first commit), where the helpers sat in front of `fetchUpdatedUser`; `c75b8eb` moved them.

Current placement in `profile.ts`:

| line | symbol |
|---|---|
| 2421 | `fetchUpdatedUser` |
| 2801 | `persistPassiveRegenToDb` |
| 2879 | the six `scopedRead` helpers |
| 2956 | `fetchPublicUsers` |
| 3125 | `fetchAttributes` |

So they are in the bottom convenience-helper region, directly alongside `fetchPublicUsers` and `fetchAttributes`. They sit after `fetchUpdatedUser` rather than at the very end of the file because they are private to it, mirroring how `persistPassiveRegenToDb` is already placed.

PR #1479, opened 2026-08-30
perf(profile): send achievement progress, not the achievement catalogue

MathiasGruber:
Second fix out of the performance audit: `profile.getUser` shipping the whole achievement catalogue on every call.

## The problem

`fetchUpdatedUser` synthesises a `QuestHistory`-shaped entry for every achievement in the game (`mockAchievementHistoryEntries`) and attaches the **full `Quest` row** to each one — every objective with its description, scene assets and reward spec. Achievements are global static content, identical for every player and unchanged unless staff edit them, but they were re-serialised, re-transferred and re-parsed on every `getUser` call, and `getUser` runs on every page load.

Measured through the local test-user broker on a freshly provisioned level-50 account with an empty inventory and no learned jutsu — the floor, not a heavy account:

```
total response              116,417 bytes raw / 15,494 gzip
userData.userQuests          90,265 bytes  (26 entries, 24 of them achievements)
```

## The fix

Split the static catalogue from the per-user progress.

- `profile.getUser` moves achievement rows out of `userData.userQuests` into a new top-level `achievementProgress` field, carrying the progress counters alone.
- A new `quests.getAchievementCatalogue` query serves the definitions, fetched only when the logbook's Achievements tab is open — the one place in the client that renders achievements.
- `LogbookAchievements` joins progress + catalogue + the `questData` trackers it already used.

Both halves are cut from **one list**, `fetchPublishedAchievements`, rather than each re-deciding with a predicate. That matters: by the time `getUser` runs, `fetchUpdatedUser` has rewritten those quests' objectives in place — trackers write resolved coordinates back onto them, and `controlShownQuestLocationInformation` rewrites them again — so a predicate applied there and the same predicate applied to the raw database row can disagree. Cutting both sides from the same list makes the split a partition by construction: nothing is sent twice, and no progress row arrives whose definition the client cannot resolve. It also subsumes the hidden, orphaned and changed-`questType` cases without enumerating them — anything the catalogue does not publish simply keeps travelling on the user object, exactly as before.

An achievement whose objectives **reach into the overworld** is withheld from the catalogue and stays inline, because the sector view, the world map and `getActiveObjectives` read `userQuests[].quest.content.objectives` directly. In current content that is one quest ("Kill the Creator"); the other 29 move.

The list costs no round-trip: the underlying read is request-scoped-memoized (#1478) and `fetchUpdatedUser` has already taken it.

Only `profile.getUser` is affected. `fetchUpdatedUser` itself is untouched, so combat, quest tracking, reward checks and every other server caller still see the full achievement list.

## Validation

Same account, same broker, before/after on one local server:

| | before | after |
|---|---|---|
| `profile.getUser` response | **116,417 raw / 15,494 gzip** | **60,181 raw / 12,001 gzip** |
| `userData.userQuests` | 90,265 B (26 entries) | 26,970 B (3 entries) |
| `achievementProgress` | — | 7,061 B (23 entries) |

**−56 KB raw (−48%) and −3.5 KB gzip (−22%) on every page load.** The raw number is the one that matters most: this is main-thread `JSON.parse` + superjson-deserialize time on the critical path of every navigation and every post-action refresh. Gzip already ate most of the wire cost, because the catalogue is 30 near-identical reward blocks — the win here is CPU, not bandwidth.

Checked independently against production content: 30 non-hidden achievements, 71 KB of `Quest` JSON, and — weighting the per-village variation across the 5,227 accounts active in the last 30 days — a mean of **~49 KB raw removed per `getUser` call** in production, which lands within 2.3% of the broker measurement per entry. Production has **zero** orphaned achievement `QuestHistory` rows and **zero** rows whose `questType` disagrees with their quest, so no progress row can arrive without a resolvable definition.

The counter-cost is one catalogue fetch per session — 76,393 raw / 5,682 gzip — and only for players who open the Achievements tab. Net positive for any session past its second page load. When the Achievements tab is the persisted one, the query batches into the page's existing tRPC request, so it is not even a separate round-trip.

Verified in the browser against a real signed-in account: the Achievements tab renders the same 25 entries as before, with rewards, descriptions, objectives and progress counters intact, and nothing duplicated. `/travel` and `/missionhall` — the other pages that read `userQuests` — show no quest-related console errors.

Also verified against a catalogue that loads but is **short** by two definitions (the endpoint returning `.slice(0, -2)`, standing in for one fetched before staff published an achievement): the banner appears, the list renders 24 entries instead of 25, exactly one automatic refetch goes out — the id-keyed effect does not re-fire when the refetch returns the same rows — and Retry restores all 25 with the banner gone.

And against a forced 500 on the catalogue endpoint: the tab shows a "could not load" banner with a Retry that re-issues the query, and keeps the entries that did arrive on screen rather than replacing them. Worth noting because the obvious guard, `isError`, does **not** work — it reads false through the query's retry backoff, so the first attempt at this rendered nothing and the tab still silently showed a short list. The banner keys on the data instead: definitions absent while progress rows exist.

`make typecheck`, `make lint` and `make test` all pass; 15 new tests cover the partition, the counters, the divergence guard, the overworld predicate and, against a real MySQL, the published list itself.

## Behaviour changes worth knowing

- Completing an achievement mid-combat no longer pops the logbook modal (`Combat.tsx` / `Sector.tsx` look the quest up in `userData.userQuests`). That modal exists to show the *next* objective of a consecutive quest; every achievement in production has exactly one objective, so there was never a next one to show. The achievement still appears — and still auto-claims — in the logbook.
- Content changes still reach players on `getUser`'s own cadence, by three complementary routes: an achievement published or hidden since the last fetch appears as a progress row with no definition and triggers an immediate refetch; an edit to an existing one changes no ids, so a 5-minute `refetchInterval` covers it; and `staleTime` covers remount and refocus, which an interval cannot reach while unmounted. A staff quest edit also invalidates the cache outright.
- The interval is not a tax on the saving: the query is mounted only while the Achievements tab is open, and it polls on the same cadence `getUser` already uses — a poll that previously carried these very definitions to every player on every page. It also pauses while the tab is backgrounded (verified: unchanged across four intervals with the document hidden).

## Summary by CodeRabbit

* **New Features**
  * Added an achievement catalogue showing published, non-overworld achievements.
  * Achievement progress is now tracked and displayed separately from catalogue definitions.
  * Added automatic catalogue refreshes and retry messaging when achievement definitions are unavailable.
  * Quest updates now refresh achievement progress and catalogue data.

* **Bug Fixes**
  * Improved handling of missing, unpublished, or outdated achievement definitions while preserving progress.

### DIFF
--- a/app/src/layout/Logbook.tsx
+++ b/app/src/layout/Logbook.tsx
@@ -41,6 +41,9 @@ import Post from "./Post";
 const tabs = ["Active", "History", "Battles", "Achievements"] as const;
 type tabType = (typeof tabs)[number];
 
+/** Matches the interval `profile.getUser` refreshes its own achievement progress on. */
+const CATALOGUE_REFRESH_MS = 5 * 60 * 1000;
+
 const Logbook: React.FC = () => {
   // State
   const [tab, setTab] = useState<tabType | null>(null);
@@ -79,11 +82,19 @@ const LogbookAchievements: React.FC = () => {
   const [activeElement, setActiveElement] = useState<string>("");
 
   // Achievement definitions are the same for every player and change only when staff edit
-  // content, so profile.getUser sends progress alone and they are fetched here instead. The
-  // staleTime overrides the app-wide Infinity so an edit to an existing achievement is picked up
-  // the next time this remounts or the window regains focus; it does not schedule a fetch of its
-  // own, which is what the unresolved-row effect below is for. isLoading rather than isPending: a
-  // disabled query never stops being pending, which would leave the tab on its spinner forever.
+  // content, so profile.getUser sends progress alone and they are fetched here instead.
+  //
+  // The refresh knobs together cover the three ways the cache can fall behind. An achievement
+  // published or hidden since the last fetch shows up as a progress row with no definition, which
+  // the effect below reacts to at once. An edit to an existing one changes no ids and so is
+  // invisible to that check, which is what refetchInterval is for; staleTime covers the same case
+  // on remount and refocus, since an interval only runs while this is mounted and focused.
+  //
+  // Polling this costs less than the code it replaces rather than eating into the saving: it runs
+  // only while the Achievements tab is open, on the same cadence profile.getUser already polls at
+  // (UserContext), and that poll used to carry these very definitions for every player on every
+  // page. isLoading rather than isPending: a disabled query never stops being pending, which would
+  // leave the tab on its spinner forever.
   const {
     data: catalogue,
     isFetched,
@@ -92,7 +103,8 @@ const LogbookAchievements: React.FC = () => {
     refetch,
   } = api.quests.getAchievementCatalogue.useQuery(undefined, {
     enabled: !!userData,
-    staleTime: 5 * 60 * 1000,
+    staleTime: CATALOGUE_REFRESH_MS,
+    refetchInterval: CATALOGUE_REFRESH_MS,
   });
 
   const definitions = useMemo(
--- a/app/src/layout/Logbook.tsx
+++ b/app/src/layout/Logbook.tsx
@@ -84,10 +84,20 @@ const LogbookAchievements: React.FC = () => {
   // refetch, and a definition that stayed cached past that point would leave a freshly published
   // achievement rendering nothing. isLoading rather than isPending: a disabled query never stops
   // being pending, which would leave the tab on its spinner forever.
-  const { data: catalogue, isLoading } = api.quests.getAchievementCatalogue.useQuery(
-    undefined,
-    { enabled: !!userData, staleTime: 5 * 60 * 1000 },
-  );
+  const {
+    data: catalogue,
+    isLoading,
+    refetch,
+  } = api.quests.getAchievementCatalogue.useQuery(undefined, {
+    enabled: !!userData,
+    staleTime: 5 * 60 * 1000,
+  });
+
+  // Progress rows arrived but their definitions did not, so the list below is missing entries.
+  // Keyed on the data itself rather than on `isError`, which flickers false between the query's
+  // retries: what matters to the player is whether anything is missing, not how it went wrong.
+  const missingDefinitions =
+    !isLoading && !catalogue && (achievementProgress?.length ?? 0) > 0;
 
   const quests = useMemo(() => {
     const definitions = new Map(catalogue?.map((q) => [q.id, q]));
@@ -118,6 +128,18 @@ const LogbookAchievements: React.FC = () => {
 
   return (
     <div className="">
+      {/* Say so, and keep whatever did arrive on screen: a list that looks complete but is not
+          hides achievements, and one that never renders never auto-claims either. */}
+      {missingDefinitions && (
+        <div className="flex flex-row items-center gap-3 p-3">
+          <span className="text-muted-foreground text-sm">
+            Could not load the achievement list. Your progress is safe.
+          </span>
+          <Button type="button" variant="info" onClick={() => void refetch()}>
+            Retry
+          </Button>
+        </div>
+      )}
       {quests
         .filter((uq) => uq.completed === 0)
         .map((uq) => {
--- a/app/src/layout/Logbook.tsx
+++ b/app/src/layout/Logbook.tsx
@@ -80,40 +80,61 @@ const LogbookAchievements: React.FC = () => {
 
   // Achievement definitions are the same for every player and change only when staff edit
   // content, so profile.getUser sends progress alone and they are fetched here instead. The
-  // staleTime overrides the app-wide Infinity: progress arrives on getUser's own five-minute
-  // refetch, and a definition that stayed cached past that point would leave a freshly published
-  // achievement rendering nothing. isLoading rather than isPending: a disabled query never stops
-  // being pending, which would leave the tab on its spinner forever.
+  // staleTime overrides the app-wide Infinity so an edit to an existing achievement is picked up
+  // the next time this remounts or the window regains focus; it does not schedule a fetch of its
+  // own, which is what the unresolved-row effect below is for. isLoading rather than isPending: a
+  // disabled query never stops being pending, which would leave the tab on its spinner forever.
   const {
     data: catalogue,
     isLoading,
+    isFetching,
     refetch,
   } = api.quests.getAchievementCatalogue.useQuery(undefined, {
     enabled: !!userData,
     staleTime: 5 * 60 * 1000,
   });
 
-  // Progress rows arrived but their definitions did not, so the list below is missing entries.
-  // Keyed on the data itself rather than on `isError`, which flickers false between the query's
-  // retries: what matters to the player is whether anything is missing, not how it went wrong.
-  const missingDefinitions =
-    !isLoading && !catalogue && (achievementProgress?.length ?? 0) > 0;
+  const definitions = useMemo(
+    () => new Map(catalogue?.map((q) => [q.id, q])),
+    [catalogue],
+  );
+
+  // Progress rows the cached catalogue cannot resolve, which the join below would otherwise drop
+  // in silence. Checked per row rather than as "did the catalogue arrive at all": one fetched
+  // before staff published an achievement is present but short, and a failed refresh keeps
+  // serving the older rows. Keyed on the data for the same reason `isError` is not used here - it
+  // reads false across the query's retry backoff, so a guard on it renders nothing.
+  const unresolvedIds = useMemo(
+    () =>
+      (achievementProgress ?? [])
+        .filter((progress) => !definitions.has(progress.questId))
+        .map((progress) => progress.questId)
+        .join(","),
+    [achievementProgress, definitions],
+  );
+
+  // Staleness alone never refetches a query that stays mounted, so a definition published after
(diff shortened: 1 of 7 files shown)


MathiasGruber:
@coderabbitai review

app/src/layout/Logbook.tsx
```diff
@@ -75,27 +75,52 @@ const Logbook: React.FC = () => {
...
-    if (quests && quests.length > 0 && !activeElement) {
+    if (quests.length > 0 && !activeElement) {
       const firstAchievement = quests[0];
       if (firstAchievement) {
         setActiveElement(firstAchievement.quest.name);
       }
     }
   }, [quests, activeElement]);
 
+  if (isLoading) return <Loader explanation="Loading achievements..." />;
```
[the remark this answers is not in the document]
MathiasGruber: Valid — fixed in 1c53bc79a. The data review reached the same conclusion independently, so this is two for two.

One correction to the suggested fix, though: `isError` is not the right signal, and I only found that by verifying rather than by reading. Against a forced 500 on the catalogue endpoint, `isError` is false through the query's retry backoff (`isLoading` is `isPending && isFetching`, and both flicker between attempts), so an `isError` guard rendered nothing at all and the tab still silently showed a short list.

The banner now keys on the data instead — definitions absent while progress rows exist — which is the condition the player actually cares about and is independent of how the query failed. Verified in the browser: banner + Retry appear on a forced failure, the entries that did arrive stay on screen rather than being replaced, Retry re-issues the query, and a healthy server renders all 25 entries with no banner.

MathiasGruber (commented):
Validated against the current PR head and the branch's installed TanStack Query 5.101.4. Two cache-consistency findings remain.

app/src/layout/Logbook.tsx
```diff
@@ -75,27 +75,74 @@ const Logbook: React.FC = () => {
...
+    refetch,
+  } = api.quests.getAchievementCatalogue.useQuery(undefined, {
+    enabled: !!userData,
+    staleTime: 5 * 60 * 1000,
+  });
+
+  // Progress rows arrived but their definitions did not, so the list below is missing entries.
+  // Keyed on the data itself rather than on `isError`, which flickers false between the query's
+  // retries: what matters to the player is whether anything is missing, not how it went wrong.
+  const missingDefinitions =
```
MathiasGruber: [P2] Detect missing definitions per progress row

This only catches a completely absent catalogue. A defined catalogue can still lack one or more `achievementProgress` IDs—for example when publish/hide/delete lands between the two requests, or when a failed refresh retains older cached data. The `flatMap` below then silently drops those rows while this condition keeps the Retry banner hidden. I reproduced the state with two progress rows and one definition: this condition was false and the join emitted only one row. Build the definition map first and treat `achievementProgress.some((p) => !definitions.has(p.questId))` as missing data.

app/src/layout/Logbook.tsx
```diff
@@ -75,27 +75,74 @@ const Logbook: React.FC = () => {
...
+  // refetch, and a definition that stayed cached past that point would leave a freshly published
+  // achievement rendering nothing. isLoading rather than isPending: a disabled query never stops
+  // being pending, which would leave the tab on its spinner forever.
+  const {
+    data: catalogue,
+    isLoading,
+    refetch,
+  } = api.quests.getAchievementCatalogue.useQuery(undefined, {
+    enabled: !!userData,
+    staleTime: 5 * 60 * 1000,
```
MathiasGruber: [P2] Schedule the catalogue refresh rather than only marking it stale

`staleTime` does not trigger a fetch for a mounted query; it only changes the freshness state. With the exact installed QueryObserver, after waiting past `staleTime` the result became `isStale: true` but the query-function call count remained 1. Because `profile.getUser` refreshes progress every five minutes and the quest-update invalidation only reaches the editor's local QueryClient, a tab that stays mounted and focused can retain old definitions/rewards indefinitely. Add a matching `refetchInterval`, or explicitly refetch/version-check when progress and catalogue diverge.

app/src/layout/Logbook.tsx
MathiasGruber: Agreed, fixed in b4987e625. You are right that `!catalogue` only catches a wholly absent catalogue — a short one passes it, and the `flatMap` then drops those rows with no banner.

Changed to exactly what you suggested: the definition map is built first and the condition is now `achievementProgress.some((p) => !definitions.has(p.questId))` (kept as a joined id string so it can also key the refetch effect). That subsumes the absent-catalogue case, since an undefined catalogue makes every row unresolved.

Verified in the browser rather than by reasoning: with the endpoint returning `.slice(0, -2)` so the catalogue loads but is two definitions short, the banner appears and the list renders 24 instead of 25 — the old condition was false in exactly that state. Restoring the endpoint and hitting Retry brings it back to 25 with the banner gone. The total-failure case still behaves as before (banner plus the entries that did arrive).

app/src/layout/Logbook.tsx
MathiasGruber: Correct, and my PR body was wrong because of it — I claimed "up to five minutes", which `staleTime` does not buy. Fixed in b4987e625 and the PR body corrected.

I went with the second of your two options rather than `refetchInterval`: polling a 76 KB static catalogue every five minutes for anyone sitting on the Achievements tab would give back a real slice of what this PR saves, and it would keep polling in the overwhelmingly common case where nothing has changed. Instead an effect keyed on the unresolved progress ids refetches the moment a row cannot be resolved — demand-driven, so it costs nothing until it is needed, and it is the same signal that drives the banner from the sibling comment.

Loop safety was the thing I wanted to be sure of, so I checked it rather than argued it: against a deliberately short catalogue, exactly two requests went out (initial + one refetch). The refetch returns the same rows, the id key is unchanged, the effect does not re-fire, and the banner takes over.

`staleTime` stays, but for the case divergence structurally cannot see: a staff edit to an existing achievement's text or rewards changes no ids, so it is picked up on the next mount or window focus.

MathiasGruber:
@coderabbitai review

app/src/layout/Logbook.tsx
```diff
@@ -75,27 +75,96 @@ const Logbook: React.FC = () => {
...
+    [achievementProgress, definitions],
   );
 
+  // Staleness alone never refetches a query that stays mounted, so a definition published after
+  // this was cached would otherwise never arrive. Pull again as soon as a row cannot be resolved,
+  // keyed on the ids themselves: a refetch that comes back with the same rows leaves the key
+  // unchanged and does not fire a second time.
+  useEffect(() => {
+    if (unresolvedIds) void refetch();
+  }, [unresolvedIds, refetch]);
```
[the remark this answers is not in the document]
MathiasGruber: Good catch, fixed in edb18c1b5. I had flagged the first-mount call to myself and checked the request count, but I assumed it deduped into the in-flight fetch — you are right that `refetch()` defaults to `cancelRefetch: true`, so it aborts and reissues instead. It now passes `cancelRefetch: false` and joins the request already going out.

The banner also waits for `isFetched` now, which closes the frame you describe: the window after progress arrives but before the catalogue fetch has started, where `isFetching` is still false.

Re-verified all three paths in the browser rather than reasoning about them:

- healthy mount — exactly **one** batched catalogue request, no banner, 25 entries
- catalogue short by two definitions — exactly **two** requests (initial + one refetch, the id key unchanged so it does not re-fire), banner shown, 24 entries
- endpoint returning 500 — banner shown with the entries that did arrive retained

Note the effect deliberately keys on the unresolved id string and not on `isFetching`: putting a fetch flag in the deps makes it re-fire every time the flag settles, which loops on a persistently short catalogue.

app/src/layout/Logbook.tsx
MathiasGruber: You are right that this was only partly solved, and it is now closed in 6515f8920 — I added the periodic refresh, which was the second of your two suggested options.

I had documented the edit case as a known residual (code comment, my earlier reply, PR body) rather than claiming it solved, but "documented" is not "handled", and your reproduction is accurate: an edit changes no ids, so the unresolved-id effect never fires.

**On the other option — a version/hash in getUser — one correction worth recording.** If the version were keyed on `Quest.updatedAt` it would not work at all: the column is `DEFAULT_GENERATED` with no `ON UPDATE`, and the edit path (`.set(input.data)`) never assigns it, so it is an insert timestamp. In production **1 of 388** quest rows has `updatedAt <> createdAt`. A version built on it would look like coverage while never changing on the very edit it is meant to catch. Only a content hash would do, and that puts hashing of the full catalogue back on `getUser` — the hot path this PR exists to shorten.

**Why polling is not the cost I earlier argued it was.** I pushed back on `refetchInterval` before; that argument does not survive the numbers. The query is mounted only while the Achievements tab is open, and it polls on the same 5-minute cadence `profile.getUser` already uses (`UserContext.tsx:102`) — a poll that, before this PR, carried these very definitions to every player on every page. So it is strictly less traffic than the code it replaces, for the only players it touches.

Verified rather than assumed, with the interval shortened and the endpoint renaming one achievement per call:

- document visible — the rendered name advanced (`POLLMARK-6575` → `6616`) with no remount and no refocus
- document hidden — unchanged across four intervals, so a backgrounded tab polls nothing (`refetchIntervalInBackground` stays false)

`staleTime` stays alongside it to cover remount and refocus, which an interval cannot reach while unmounted.

PR #1484, opened 2026-08-30
Fix Issue #1278: Allow users to pick a name and title color

### LINKED ISSUE #1278: Allow users to pick a name and title color

We would like for users to be able to pick a color to use for their name and title in the tavern.  Below are details on how we would like for the system to work.

- Users go into their profile settings to change the color of their name and title
- A color picker should be available for them to change the color.
- The colors should be pre-set so they cannot select a color to imitate staff colors
- For colors that would make it hard to see the text in light or dark mode, force them to use either the light or dark mode text color for visibility.
- The change for their username is separate from the change for their title.
- The cost to set/change their color is 10 reputation each.

Phrosfire:
# Pull Request

## License

## Summary by CodeRabbit

* **New Features**
  * Added customizable username and title colors for tavern profiles, with live previews and multiple presets.
  * Color changes cost reputation points; resetting to the default is free.
  * Custom colors now appear in tavern conversations and posts.
* **Bug Fixes**
  * Custom titles now reject reserved staff, faction, and village names.
  * Added validation to prevent invalid color selections, race-condition overspending, and unauthorized changes.
* **Tests**
  * Added coverage for color styling, purchases, validation, reserved titles, and accessibility contrast.


app/src/validators/user.ts
```diff
@@ -4,16 +4,28 @@ import {
...
 
 export const updateUserSchema = z.object({
   username: usernameSchema,
-  customTitle: z.string().min(0).max(199).optional(),
+  customTitle: z
+    .string()
+    .min(0)
+    .max(199)
+    .optional()
+    .refine((title) => !title || !isReservedCustomTitle(title), {
```
MathiasGruber: [P1] Preserve edits for legacy reserved titles

This unconditional schema refinement rejects any existing user whose unchanged `customTitle` now matches a reserved role or village. The staff edit form submits the complete user object, so editing an unrelated field such as rank or village fails before the handler can apply its `customTitleChanged` check. Keep `updateUserSchema` permissive for legacy values and enforce the reservation only inside the existing changed-value guard.

app/src/libs/tavernColors.ts
```diff
@@ -0,0 +1,142 @@
...
+    "Midnight",
+    "bg-blue-900",
+    "text-blue-600 dark:text-blue-400 font-bold",
+    "bg-blue-900",
+    "#1e3a8a",
+  ),
+  NAVY: titleStyle(
+    "Navy",
+    "bg-blue-600",
+    "text-blue-500 dark:text-sky-400 font-bold",
```
MathiasGruber: [P1] Use accessible light-mode username colors

Most light-mode username classes have insufficient contrast against the tavern popover background. Navy, Cobalt, Yellow, Slate, Gold, Crimson, Fuchsia, Mint, and Lime measure roughly 1.7–4.3:1, below 4.5:1 for this normal-sized username text. The linked issue explicitly requires readable light/dark colors, but the username test only checks for a `text-*` class. Please use darker light-mode shades and test contrast against both actual tavern backgrounds.

app/drizzle/constants.ts
```diff
@@ -985,6 +986,28 @@ export const BATTLE_LOG_FULL_LIMIT = 1000;
...
+  "CHARCOAL",
+  "GOLD",
+  "CRIMSON",
+  "FUCHSIA",
+  "MINT",
+  "LIME",
+] as const;
+export type TavernColorPreset = (typeof TavernColorPresets)[number];
+export const getTavernColorChangeCost = (color: TavernColorPreset) =>
+  color === "DEFAULT" ? 0 : COST_TAVERN_COLOR_CHANGE;
```
MathiasGruber: [P2] Charge the specified cost for Default

Returning to `DEFAULT` is still a color change, but this helper makes it free. Issue #1278 says setting/changing each color costs 10 reputation, without a free-reset exception. This also lets users with zero reputation change their persisted selection. Charge `COST_TAVERN_COLOR_CHANGE` here unless the product requirement is deliberately changed and the issue is updated.

app/bun.lock
```diff
@@ -1,5 +1,5 @@
 {
-  "lockfileVersion": 3,
+  "lockfileVersion": 1,
```
MathiasGruber: [P2] Revert unrelated lockfile regeneration

No package manifest changed, but this lockfile was regenerated with `lockfileVersion` 1 instead of 3, drops existing nested overrides, and updates many unrelated runtime/tooling dependencies. A frozen install consequently resolves versions such as `ai` 7.0.85, `openai` 7.8.0, `cytoscape` 3.34.2, and Biome 2.5.11 instead of the versions on main. Restore `app/bun.lock` from the base branch so this feature does not silently bundle a dependency upgrade.

app/src/layout/Conversation.tsx
```diff
@@ -443,6 +452,8 @@ const Conversation: React.FC<ConversationProps> = (props) => {
             role: userData.role,
             customTitle: userData.customTitle,
             federalStatus: userData.federalStatus,
+            tavernUsernameColor: posterTavernUsernameColor,
+            tavernTitleColor: posterTavernTitleColor,
```
MathiasGruber: [P2] Build the whole optimistic identity from the sender

When staff post as an AI, the new color fields come from `senderUser`, but the optimistic comment still takes `customTitle`, role, federal status, level, and village fields from the logged-in staff user. This can render the AI title color on the staff member title until the server response replaces the comment. Fetch the remaining display fields for `senderUser` and construct the complete optimistic identity from one effective profile.

MathiasGruber (approved):
Validated the reported issues, verified the fixes locally and through the signed-in dev flow, rebased onto current main, and confirmed all CI, CodeQL, CodeRabbit, and Vercel checks pass.

PR #1486, opened 2026-08-31
Farming improvements

Phrosfire:
# Pull Request

**Changes**
- Adds a toggle for displaying the night time overlay
- Displays experience (level) on the farming tab in users
- add cooking item type for gathered farm items.
- Adds cooking storage for farming items to go to.  Other items will be added here later as well.

## License

## Summary by CodeRabbit

* **New Features**
  * Added Cooking items with dedicated inventory and home-storage capacity.
  * Added Cooking tabs across storage, inventory, crafting, and shop interfaces.
  * Federal subscription tiers now include additional cooking storage.
  * Added a setting to show or hide day/night map overlays.
  * Farming listings now show experience alongside farming level.

* **Bug Fixes**
  * Updated storage, purchasing, rewards, crafting, and item transfers to enforce cooking capacity limits.
  * Cooking items no longer display battle-use effects or item-level badges.


app/src/server/api/routers/home.ts
```diff
@@ -273,30 +295,12 @@ export const homeRouter = createTRPCRouter({
...
-        if (
-          userItemResult.item.itemType === "MATERIAL" &&
-          nMaterials >= calcMaxMaterials(user)
-        ) {
-          return errorResponse("Materials inventory is full");
+        const inventoryBucket = getInventoryBucket(userItemResult.item);
+        const bucketCount =
+          nonStoredItems.filter((ui) => getInventoryBucket(ui.item) === inventoryBucket)
+            .length || 0;
+        if (bucketCount >= getInventoryBucketCapacity(inventoryBucket, user)) {
```
MathiasGruber, line 303: [P1] This is a check-then-write race: two requests retrieving different rows can read the same `bucketCount`, both pass, and both update their own `UserItem`, exceeding the inventory limit. The storage branch below has the same race for `storedInBucket`. Please serialize the capacity check and move with the repository user-snapshot CAS (or an equivalent transaction/lock), returning a retryable error when the snapshot changes.

app/bun.lock
```diff
@@ -1,5 +1,5 @@
 {
-  "lockfileVersion": 3,
+  "lockfileVersion": 1,
```
MathiasGruber: [P2] This PR downgrades the lockfile from version 3 to 1 and re-resolves hundreds of unrelated dependencies without a `package.json` change. With the repository-declared Bun 1.4.0, `bun install --frozen-lockfile` fails with `overrides in package.json changed since bun.lock was saved`. Please restore the base lockfile; this feature does not require dependency changes.

app/src/layout/CraftingCatalog.tsx
```diff
@@ -59,6 +59,11 @@ const CATEGORY_CONFIG: Record<
     icon: <FlaskConical className="h-8 w-8" />,
     itemTypes: ["CONSUMABLE"],
   },
+  COOKING: {
+    label: "Cooking",
+    icon: <ChefHat className="h-8 w-8" />,
+    itemTypes: ["COOKING"],
```
MathiasGruber, line 65: [P1] Adding `COOKING` here makes cooking recipes reachable, but `occupationRouter.craftItem` never classifies the output or checks `getInventoryBucketCapacity` before consuming materials and inserting one or more `UserItem` rows. A user whose cooking bucket is already full can therefore craft additional cooking stacks. Please compute the number of output rows and reject before the material CAS/consumption; add a regression test for a full cooking bucket.

app/src/server/api/routers/item.ts
```diff
@@ -2526,14 +2524,11 @@ export const itemRouter = createTRPCRouter({
...
-        return errorResponse("Inventory is full");
-      }
-      if (info.isEventItem && eventItemsCount >= calcMaxEventItems(user)) {
-        return errorResponse("Event item inventory is full");
-      }
-      if (info.itemType === "MATERIAL" && materialsCount >= calcMaxMaterials(user)) {
-        return errorResponse("Materials inventory is full");
+      const purchaseBucket = getInventoryBucket(info);
+      if (
+        bucketCounts[purchaseBucket] >= getInventoryBucketCapacity(purchaseBucket, user)
```
MathiasGruber, line 2532: [P1] This capacity check is not coupled to the later funds update/item insert. Two parallel purchases at `capacity - 1` can both pass and each insert a cooking row, bypassing the new limit. Please include the inventory snapshot in a CAS/transaction that owns the check through insertion, and ensure failed inserts or fund deductions cannot leave partial state.

app/src/hooks/day-night-overlay.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const SHOW_DAY_NIGHT_MAP_OVERLAYS_KEY = "showDayNightMapOverlays";
+
+/**
+ * Cached preference for day/night map shading overlays.
+ * Defaults to true so existing users keep current behavior.
+ */
+export const useDayNightMapOverlays = () => {
+  const [showDayNightMapOverlays, setShowDayNightMapOverlays] =
+    useLocalStorage<boolean>(SHOW_DAY_NIGHT_MAP_OVERLAYS_KEY, true);
```
MathiasGruber: [P2] `useLocalStorage` returns the default `true` during SSR but reads persisted `false` in the first browser render. Reproduced markup is server `data-v=true` versus first client `data-v=false`, which triggers React hydration failure and subtree regeneration whenever this setting is off. Please use an SSR-stable initial value and apply the persisted preference after mount.

MathiasGruber (approved):
Validated the farming, inventory-capacity, concurrency, hydration, lockfile, and display changes. All findings are addressed, review conversations are resolved, and the final CI cycle is green.

MathiasGruber (approved):

PR #1504, opened 2026-09-08
Gain half med exp on self heal

Phrosfire:
# Pull Request

When players heal themselves in the hospital, they gain half the medical experience instead of none.

## License

## Summary by CodeRabbit

- **New Features**
  - Self-healing is now supported.
  - Medical experience is awarded when healing yourself or another player.
  - Self-healing awards experience at half the standard rate.
  - Experience gains are limited to the remaining medical experience capacity.

- **Bug Fixes**
  - Prevents medical experience from exceeding the maximum limit or being awarded for invalid healing amounts.
  - Refreshes player information after healing to reflect updated resources and experience.

- **Tests**
  - Added coverage for healing, self-healing, experience caps, and invalid amounts.


app/src/server/api/routers/hospital.ts
```diff
@@ -127,10 +126,13 @@ export const hospitalRouter = createTRPCRouter({
       // Derived
       const { toHeal, pools } = calcHowMuchToHeal(u, t, input.healPercentage);
       const chakraCost = calcHealthToChakra(u, toHeal);
-      // Calculate experience gain, capped at 4 million
-      const rawExpGain = t.userId !== u.userId ? MEDNIN_HEAL_TO_EXP * toHeal : 0;
-      const expGain =
-        rawExpGain > 0 ? Math.min(rawExpGain, MEDNIN_EXP_CAP - u.medicalExperience) : 0;
+      // Calculate experience gain, capped at 4 million. Self-heals award half.
+      const expGain = calcHospitalHealExperience({
```
MathiasGruber, line 130: Now that a self-heal earns medical experience, the LEGENDARY self-heal path becomes a free-chakra + free-exp loop. The defect site is the target UPDATE further down (lines ~184-195, outside the diff, anchored here because that is the change that makes it exploitable).

For a LEGENDARY healer `calcMedninHealablePool` returns `["Health","Chakra","Stamina"]`, so the target UPDATE sets `curChakra = LEAST(t.curChakra + toHeal, t.maxChakra)` using `t`, which was fetched *before* the healer UPDATE subtracted `chakraCost` with `sql`${userData.curChakra} - ${chakraCost}``. When healer and target are the same row, the second statement runs after the first and overwrites the deduction with a stale snapshot.

Concrete scenario: LEGENDARY mednin (heal factor 0.15), maxChakra 8000, curChakra 7100 after spending 900 healing someone else, full HP and stamina. `calcHowMuchToHeal` at 100% gives toHeal = chakra deficit = 900, chakraCost = 135. UPDATE 1: curChakra 7100 -> 6965, medicalExperience +45. UPDATE 2: curChakra = LEAST(7100 + 900, 8000) = 8000. Net result: chakra restored to full, the 135 cost refunded, and +45 med exp. Repeat heal-others -> self-heal-to-refill indefinitely up to MEDNIN_EXP_CAP; chakra never runs out. Before this PR the same overwrite happened but paid 0 exp, so it was not an exp faucet. The client optimistic update in `HealingPopover`/`hospital/page.tsx` also subtracts `chakraCost` locally, so the displayed chakra diverges from the DB.

Fix: when `u.userId === t.userId`, apply the cost and the restore in a single UPDATE with relative SQL, e.g. `curChakra: sql`LEAST(${userData.curChakra} - ${chakraCost} + ${toHeal}, ${t.maxChakra})`` (and drop the separate target statement for that case), rather than two statements against the same row built from a pre-update snapshot.

MathiasGruber, line 130: This is about lines 107-120, which are outside the diff — anchoring here because this is the change that makes them matter.

Self-heal was previously worth 0 exp, so `input.userId === ctx.userId` was a path nobody had a reason to take. Now it pays, so it becomes a repeated, farmable action — and on that path the opening `Promise.all` runs `fetchUpdatedUser` twice against the same row.

Most of what's inside dedupes: the seven auxiliary reads all go through `scopedRead` and are keyed on either a constant or `userId`, so the second call reuses the first's in-flight promise. What does *not* dedupe is the expensive part:

- the nested `client.query.userData.findFirst` — bloodline, sageMode, activeReskin, clan, village + structures + relationshipA/B + sectors, anbuSquad, loadout, promotions, equipped items + item, open userQuests + quest, completedQuests, votes — the heaviest read in the endpoint, issued twice;
- `persistPassiveRegenToDb`, which `forceRegen: true` guarantees for an AWAKE user: with `includeVillageState` true that's a village-state `findFirst` plus an `UPDATE userData`, so the self-heal issues two concurrent regen writes to the same row before the procedure has written anything of its own.

So a self-heal pays two extra reads and an extra write on a row it is about to update twice more, and the two regen writes race each other with snapshots taken independently. The two snapshots also each run the tier/exam bootstrap loop on their own `user.userQuests`, so both can decide the user has no tier quest and call `insertNextQuest`.

One branch fixes it:

```ts
const isSelfHeal = input.userId === ctx.userId;
const [updatedUser, updatedTargetOrNull, relationships] = await Promise.all([
  fetchUpdatedUser({ client: ctx.drizzle, userId: ctx.userId, userIp: ctx.userIp, forceRegen: true }),
  isSelfHeal
    ? null
    : fetchUpdatedUser({ client: ctx.drizzle, userId: input.userId, forceRegen: true }),
  fetchAlliances(ctx.drizzle),
]);
const updatedTarget = updatedTargetOrNull ?? updatedUser;
```

Reusing the healer's fetch is also strictly more correct for the self case — `u` and `t` stop being two independent snapshots of one row.

app/src/server/api/routers/hospital.ts
```diff
@@ -127,10 +126,13 @@ export const hospitalRouter = createTRPCRouter({
...
+        healerId: u.userId,
+        targetId: t.userId,
```
MathiasGruber, line 132: This is about lines 171-200, outside the diff — anchoring on the line that makes the self-heal path hot.

On a self-heal `u.userId === t.userId`, so the guarded healer update (line 171) and the target update (line 184) are two sequential round-trips to the *same* row. The serialisation is unavoidable when healer and target differ — `uResult.rowsAffected` is what gates the heal — but when they are the same row the whole thing is one guarded statement:

```sql
UPDATE UserData
   SET medicalExperience = medicalExperience + :expGain,
       curChakra = LEAST(curChakra - :chakraCost + :toHeal, maxChakra),  -- only if pools includes Chakra
       curHealth = LEAST(curHealth + :toHeal, maxHealth),
       questData = :questData,
       regenAt = NOW()
 WHERE userId = :userId AND curChakra >= :chakraCost
```

One `rowsAffected === 1` check covers both, and the sensei share still runs in the existing `Promise.all`. That removes a full PlanetScale round-trip from a mutation this PR is about to make far more frequent.

Worth collapsing for a second reason: the target update writes *absolute* values derived from the pre-deduction `t` snapshot (`LEAST(${t.curChakra + toHeal}, ${t.maxChakra})`), not SQL increments. For a LEGENDARY mednin — the only rank whose `calcMedninHealablePool` returns `Chakra` — a self-heal therefore overwrites `curChakra` with the snapshot value plus the heal, discarding the `curChakra - chakraCost` the first statement just applied. The chakra cost is refunded and self-heal med exp becomes free to farm in a loop. Merging the two statements removes the stale-snapshot write along with the round-trip.

MathiasGruber (approved):

PR #1556, opened 2026-09-14
fix(native): dismiss overlays before Android Back navigation

MathiasGruber:
Android Back currently navigates or exits while a settings popover or dialog is open. Send a cancelable Escape event first so the existing Radix layer can dismiss itself or block dismissal, and navigate only when no layer consumes the event. This reuses the repository’s overlay behavior without a separate overlay registry.

Validation: reproduced Back backgrounding the app with audio settings open on the Android 16 emulator. The repository make test command passes: 1,909 passed, 159 database-dependent tests skipped, zero failures. Three focused NativeBridge tests cover consumed Back, history navigation, and exit without history using the existing Bun-compatible spy conventions. Typecheck, targeted Biome and diff checks pass. The updated frontend still needs emulator validation; the production Android bridge fix is tracked separately in #1555.

Preview validation: [TNR reviewer run](https://github.com/studie-tech/TheNinjaRPG/actions/runs/34844795325) passed settings-popover dismissal, ordinary Escape dismissal, history navigation, and exit without history against the deployed preview using an injected Android bridge stub. This validates browser behavior but does not replace real Android shell testing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated Android back-button behavior to respect open dialogs, menus, and other overlays.
  - When an overlay handles the back action, the app now keeps the current screen instead of navigating backward or exiting unexpectedly.
  - When no overlay is active, the back button continues to follow the appropriate browser history or app-exit behavior.


MathiasGruber:
@coderabbitai review

